PR #3, opened 2021-12-05
Add Update Dependencies button to Package Dependencies title bar

adam-fowler:
This is probably the operation I do the most often in relation to swift package manager, making it easy of access would be a good thing

I've included the package-lock.json as it has the wrong package name

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,11 +1,11 @@
 {
-  "name": "vscode-swift",
+  "name": "swift",
   "version": "0.0.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
-      "name": "vscode-swift",
+      "name": "swift",
       "version": "0.0.1",
       "devDependencies": {
         "@types/glob": "^7.1.4",
--- a/package.json
+++ b/package.json
@@ -21,12 +21,6 @@
   "contributes": {
     "commands": [
       {
-        "command": "swift.resolveDependencies",
-        "title": "Resolve Package Dependencies",
-        "icon": "$(refresh)",
-        "category": "Swift"
-      },
-      {
         "command": "swift.updateDependencies",
         "title": "Update Package Dependencies",
         "icon": "$(cloud-download)",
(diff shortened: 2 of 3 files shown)


svanimpe:
Now that resolves happen automatically anyway, maybe we can just change that "Refresh" button to do an update instead of a resolve?

adam-fowler:
I was going to suggest having a config option to enable/disable the automatic resolve. If you are working with large SPM projects it can be a pain.

adam-fowler:
Another thing I think the package resolve is not working completely. In that it does unnecessary resolves. The resolve should only be triggered by a change in `Package.swift`. It seems to trigger on a change to `Package.resolved`.  I'll add an issue related to this.

svanimpe:
The latter was intentional. @0xTim mentioned that users may pull in a newer version of Package.resolved, in which case there should be an additional `resolve`.

But it shouldn't be too hard to make both features optional.

adam-fowler:
> The latter was intentional. @0xTim mentioned that users may pull in a newer version of Package.resolved, in which case there should be an additional resolve.
> 
> But it shouldn't be too hard to make both features optional.

That doesn't make sense `swift package resolve` updates the `Package.resolved` file. Why would you run it again if it has been updated by an external process. You can see this happen when you press the `swift package update` button, it runs an unnecessary resolve immediately after

adam-fowler:
You will want to rebuild the package view based on changes to `Package.resolved`

0xTim (approved):

svanimpe:
Is it OK if I remove the Resolve button in favor of the Update button? Resolves should mostly happen automatically, so there shouldn't be a need for a button for it.

There's always the Resolve task should a user want to run it without opening a command line.

0xTim:
+1 from me

adam-fowler:
> Is it OK if I remove the Resolve button in favor of the Update button? Resolves should mostly happen automatically, so there shouldn't be a need for a button for it.

done

svanimpe (approved):

PR #5, opened 2021-12-06
Separate triggering of resolve, for different file updates

adam-fowler:
Updating `Package.swift` and `Package.resolved` triggers `swift package resolve`.


adam-fowler:
@0xTim This is an updated version of the PR, which will run `swift package resolve` on both `Package.swift` and `Package.resolved` files being changed. It catches the situation where it would run twice when we update the `Package.resolved` file after changing `Package.swift`

adam-fowler:
Gonna close this until we can come up with a better solution to the resolve tasks

PR #6, opened 2021-12-06
SwiftPackage and SwiftContext object

adam-fowler:
The idea behind this PR is to create a `Ctx` context object that is passed into all the classes. It should provide an easy way to access global data for the workspace.

Currently `Ctx` only contains the `ExtensionContext` and an `SPMPackage` object which manages the loading of the SPM data from `swift package describe --type json`. This data is needed (or will be) in multiple places across the project. Also since I centralised the loading of the package data I also moved the setting of the `contextKeys` to be in the same place.


src/SwiftTaskProvider.ts
```diff
@@ -132,11 +132,10 @@ export class SwiftTaskProvider implements vscode.TaskProvider {
     }
 
     /**
-     * Uses `swift package describe` to find all targets in this package.
+     * Uses spm package description cache to find all targets in this package.
```
0xTim, line 135: Nit spelling
```suggestion
     * Uses SwiftPM package description cache to find all targets in this package.
```

0xTim (changes requested):
One minor nit - otherwise LGTM

0xTim (approved):

src/ctx.ts
```diff
@@ -0,0 +1,20 @@
+import * as vscode from 'vscode';
+import { SPMPackage } from './package';
+
+export class Ctx {
+	private constructor(
+        public workspaceRoot: string,
```
svanimpe, line 6: `public` is the default for TypeScript. Is there a reason you've added it here?

src/ctx.ts
```diff
@@ -0,0 +1,20 @@
...
+        public extContext: vscode.ExtensionContext,
+        public spmPackage: SPMPackage
+    ) {}
+
+    static async create(
```
svanimpe, line 11: What's the motivation for using a factory method here?

I understand the general benefits of factory methods, but if there's no concrete need for one here, I'd prefer not to include them. Every layer of abstraction we add just makes the codebase harder to get into for newcomers.

src/ctx.ts
```diff
@@ -0,0 +1,20 @@
+import * as vscode from 'vscode';
+import { SPMPackage } from './package';
+
+export class Ctx {
```
svanimpe, line 4: Can we just name this type `Context`?

I know abbreviations are one of my pet peeves, but I very much prefer the clarity of a full name. Abbreviations like this noticeably increase my cognitive load when reading code.

src/PackageWatcher.ts
```diff
@@ -14,7 +15,13 @@ export class PackageWatcher {
     private packageFileWatcher?: vscode.FileSystemWatcher;
     private resolvedFileWatcher?: vscode.FileSystemWatcher;
 
-    constructor(private workspaceRoot: string) { }
+    private ctx: Ctx
+    private workspaceRoot: string
+
+    constructor(workspaceRoot: string, ctx: Ctx) {
```
svanimpe, line 21: This can just be:

```
constructor(
  private workspaceRoot: string
  private ctx: Ctx
) { }
```

This seems to be the common pattern in TypeScript.

src/PackageWatcher.ts
```diff
@@ -53,15 +57,7 @@ export class PackageWatcher {
...
-        watcher.onDidDelete(async () => {
-            if (await pathExists(this.workspaceRoot, 'Package.swift')) {
-                // Recreate Package.resolved.
-                this.handlePackageChange();
-            } else {
-                contextKeys.hasPackage = false;
-                contextKeys.packageHasDependencies = false;
-            }
-        });
+        watcher.onDidDelete(async () => await this.handlePackageChange());
```
svanimpe, line 60: With these changes, both methods are nearly identical, safe for the file name, so we can merge them and add the file name as a parameter.

src/package.ts
```diff
@@ -0,0 +1,57 @@
+import { privateEncrypt } from 'crypto';
+import contextKeys from './contextKeys';
+import { exec, pathExists } from './utilities';
+
+export class SPMPackage {
```
svanimpe, line 5: Can we just name this `Package`?

I assume there's no confusion here as to what kind of package we're talking about? If there is, I'd prefer a simpler `SwiftPackage` over `SPMPackage` (SPM is not the official abbreviation) or `SwiftPMPackage` (which has repeated information, as the P also stands for Package).

src/package.ts
```diff
@@ -0,0 +1,57 @@
...
+    ) {
+        this.setContextKeys()
+    }
+
+    public static async create(folder: string): Promise<SPMPackage> {
+        try {
+            let contents = await SPMPackage.loadPackage(folder)
+            return new SPMPackage(folder, contents)
+        } catch(error) {
+            // TODO: output errors
```
svanimpe, line 18: Which errors are being caught here? It's not immediately clear (or documented) that `loadPackage` can throw.

I assume `exec` throws an error when `swift package describe --type json` fails to find `Package.swift`. Is that correct? If so, can we make this explicit to improve the clarity of the code?

src/package.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export class SPMPackage {
+	private constructor(
+        readonly folder: string,
+        public contents: any
+    ) {
+        this.setContextKeys()
+    }
+
+    public static async create(folder: string): Promise<SPMPackage> {
```
svanimpe, line 13: Same question here: Do we really need the factory method?

src/package.ts
```diff
@@ -0,0 +1,57 @@
+import { privateEncrypt } from 'crypto';
+import contextKeys from './contextKeys';
+import { exec, pathExists } from './utilities';
+
+export class SPMPackage {
+	private constructor(
+        readonly folder: string,
+        public contents: any
```
svanimpe, line 8: Can we try to avoid using `any`? 

We could, for example, define interfaces declaring the fields we expect to find in the JSON. These don't have to be complete, but they would remove a lot of mystery from the code.

src/package.ts
```diff
@@ -0,0 +1,57 @@
...
+        try {
+            let contents = await SPMPackage.loadPackage(folder)
+            return new SPMPackage(folder, contents)
+        } catch(error) {
+            // TODO: output errors
+            return new SPMPackage(folder, null)
+        }
+    }
+
+    public static async loadPackage(folder: string): Promise<any> {
```
svanimpe, line 23: Does this have to be static? Can it be just `load`?

src/package.ts
```diff
@@ -0,0 +1,57 @@
...
+        const { stdout } = await exec('swift package describe --type json', { cwd: folder });
+        return JSON.parse(stdout)
+    }
+
+    public async reload() {
+        try {
+            this.contents = await SPMPackage.loadPackage(this.folder)
+        } catch {
+
```
svanimpe, line 32: Shouldn't we explicitly set `contents` to undefined here?

src/package.ts
```diff
@@ -0,0 +1,57 @@
...
+    public async reload() {
+        try {
+            this.contents = await SPMPackage.loadPackage(this.folder)
+        } catch {
+            
+        }
+        this.setContextKeys()
+    }
+
+    get products(): Array<any> {
```
svanimpe, line 37: This is, for example, where we could benefit from using interfaces to declare the fields we expect to see in the JSON output.

src/package.ts
```diff
@@ -0,0 +1,57 @@
...
+    get dependencies(): Array<any> {
+        return this.contents.dependencies
+    }
+
+    get targets(): Array<any> {
+        return this.contents.targets
+    }
+
+    private setContextKeys() {
+        if (this.contents == null) {
```
svanimpe, line 50: This can probably just be `if (this.contents)`. That should check for both `null` and `undefined`. I assume `== null` does the same thing, but I've learned that using `==` in JavaScript is almost always a bad idea 😅

svanimpe (changes requested):
Sorry for the many comments 😅
Let me know what you think. I can just handle them all and add documentation when I merge this.

src/ctx.ts
adam-fowler: changing to `public` gives the following error
```
'public' modifier cannot appear on a module or namespace element.
````

File: src/ctx.ts (hunk 3 above)
Comment (adam-fowler) [reply src/ctx.ts:11]: It is an async function. You cannot have async constructors

File: src/ctx.ts (hunk 4 above)
Comment (adam-fowler) [reply src/ctx.ts:4]: `Context` as with `Ctx` are both very vague. I copied this pattern from the rust vscode extension (they were using Ctx). Maybe rename to `SwiftContext` or `SPMContext`.

File: src/PackageWatcher.ts (hunk 6 above)
Comment (adam-fowler) [reply src/PackageWatcher.ts:60]: I am looking to have them call different functions, as updating `Package.swift` requires a reload of `swift package describe` while updating `Package.resolved` doesn't. Also the work LLDB work I've got in the pipeline requires functions to be called only on `Package.swift` update.

File: src/package.ts (hunk 7 above)
Comment (adam-fowler) [reply src/package.ts:5]: I chose `SPMPackage` because `Package` is too close to `package` which is a keyword in Typescript. Can still change if you feel it is necessary.

File: src/package.ts (hunk 8 above)
Comment (adam-fowler) [reply src/package.ts:18]: If you call `loadPackage` here and the `Package.swift` is invalid we should probably report the errors that are causing the issue. I'm guessing in the `Problems` section.

File: src/package.ts (hunk 9 above)
Comment (adam-fowler) [reply src/package.ts:13]: Again it is an async function and you cannot have an async constructor

File: src/package.ts (hunk 10 above)
Comment (adam-fowler) [reply src/package.ts:8]: Fair point

File: src/package.ts (hunk 11 above)
Comment (adam-fowler) [reply src/package.ts:23]: It is being called from the factory function, so needs to be static

File: src/ctx.ts (hunk 2 above)
Comment (svanimpe) [reply src/ctx.ts:6]: Ah yes. I think this shorthand syntax (declare properties via the constructor) requires every property to have an explicit access control modifier.

File: src/ctx.ts (hunk 3 above)
Comment (svanimpe) [reply src/ctx.ts:11]: Thanks, I didn't know that yet!

File: src/PackageWatcher.ts (hunk 6 above)
Comment (svanimpe) [reply src/PackageWatcher.ts:60]: OK, so I'll just leave them as-is?

File: src/package.ts (hunk 13 above)
Comment (adam-fowler) [reply src/package.ts:37]: ok will set them up

File: src/ctx.ts (hunk 4 above)
Comment (svanimpe) [reply src/ctx.ts:4]: OK. Perhaps we can go for `SwiftContext` and `SwiftPackage`?

File: src/package.ts (hunk 14 above)
Comment (adam-fowler) [reply src/package.ts:50]: Now that I'm adding interfaces it can only be `undefined` or a valid value.

File: src/package.ts (hunk 8 above)
Comment (svanimpe) [reply src/package.ts:18]: I was just thinking that we should also handle cases where users are just editing Swift files, outside of a Swift package.  We should just disable the features that require a package then, but not throw any errors.

File: src/package.ts (hunk 9 above)
Comment (svanimpe) [reply src/package.ts:13]: Got it!

File: src/ctx.ts (hunk 4 above)
Comment (adam-fowler) [reply src/ctx.ts]: Done

File: src/PackageWatcher.ts (hunk 5 above)
Comment (adam-fowler) [reply src/PackageWatcher.ts]: Done

File: src/package.ts (hunk 8 above)
Comment (adam-fowler) [reply src/package.ts:49]: But we should report errors which are errors in a Package.swift. Ok I should catch the no root manifest error and not report that

File: src/package.ts (hunk 10 above)
Comment (adam-fowler) [reply src/package.ts]: Added basic interfaces, they can be expanded in a later PR if required. As I added an interface for `Target` I replaced the one in `SwiftTaskProvider.ts`

File: src/package.ts (hunk 12 above)
Comment (adam-fowler) [reply src/package.ts]: done

File: src/package.ts (hunk 13 above)
Comment (adam-fowler) [reply src/package.ts]: done

File: src/package.ts (hunk 14 above)
Comment (adam-fowler) [reply src/package.ts]: This is now a test for `undefined` and also if `===` now ESLint is telling me about all of these

**Comment by adam-fowler** [comment]
> Sorry for the many comments 😅
> Let me know what you think. I can just handle them all and add documentation when I merge this.

Comments are good, always good to have someone else go through your code

File: src/context.ts [hunk 15]
```diff
@@ -0,0 +1,20 @@
+import * as vscode from 'vscode';
+import { SwiftPackage } from './package';
+
+export class SwiftContext {
+	private constructor(
+        public workspaceRoot: string,
+        public extContext: vscode.ExtensionContext,
+        public spmPackage: SwiftPackage
```
svanimpe, line 8: This should probably also be renamed? In this case, it can just be `package` I think?

adam-fowler: Can't rename it to `package` as it is a reserved word. Changed to `swiftPackage`

src/SwiftTaskProvider.ts
```diff
@@ -132,12 +124,9 @@ export class SwiftTaskProvider implements vscode.TaskProvider {
...
     /**
-     * Uses `swift package describe` to find all targets in this package.
+     * Uses SwiftPM package description cache to find all targets in this package.
      */
-    private async findTargets(): Promise<Target[]> {
-        const { stdout } = await exec('swift package describe --type json', { cwd: this.workspaceRoot });
-        return JSON.parse(stdout).targets.map((target: any) => {
-            return { name: target.name, type: target.type };
-        });
+    private findTargets(): Target[] {
```
svanimpe, line 129: This can be removed now. We can just write `const targets = this.ctx.swiftPackage.targets;` where this method was used.

src/context.ts
```diff
@@ -0,0 +1,20 @@
+import * as vscode from 'vscode';
+import { SwiftPackage } from './package';
+
+export class SwiftContext {
+	private constructor(
+        public workspaceRoot: string,
+        public extContext: vscode.ExtensionContext,
```
svanimpe, line 7: Last abbreviation nitpick 😅 Currently this property is a `SwiftContext.extContext: ExtensionContext`. Perhaps we can just name it `extensionContext`?

src/SwiftTaskProvider.ts
adam-fowler: @svanimpe I have a slightly different change which is in my LLDB branch, which adds a function to `SwiftPackage` that returns targets of a certain type (I needed it for the launch.json generation). Can this change wait until then?

src/context.ts
adam-fowler: ok

adam-fowler: done

src/package.ts
```diff
@@ -0,0 +1,92 @@
...
+
+// Swift Package Manager dependency
+export interface Dependency {
+    identity: string
+    url?: string
+    path?: string
+}
+
+// Class holding Swift Package Manager Package
+export class SwiftPackage implements PackageContents {
```
svanimpe, line 36: Just a general comment concerning filenames:

TypeScript doesn't seem to have any strict rules, so the convention I've adopted was: If a file's primary export is a type or object, name it after that type or object (in UpperCamelCase and lowerCamelCase respectively). Otherwise, pick an appropriate lowerCamelCase name.

By that convention, we could name the new files `SwiftPackage` and `SwiftContext`. Any thoughts on this convention?

CC @0xTim

adam-fowler: I'm not fussed either way. Given the symbol look up is pretty solid I don't think it is a huge issue. But I'm not against renaming either

src/package.ts
svanimpe: Did you find a way to report errors in the `Problems` view? I cannot find any API to do that, only via problem matchers on tasks. The alternative would be to show an error notification, but then we'd have to make sure not to show it over and over again.

adam-fowler: Not yet was going to leave that for another PR

src/SwiftTaskProvider.ts
svanimpe: Sure!

src/package.ts
adam-fowler: You are right? And it is fixed

svanimpe (approved):

PR #8, opened 2021-12-07
Add Apache License

### LINKED ISSUE #7: vscode-swift requires a LICENSE file
This will have the same issues as the other swift-server repositories https://github.com/swift-server/async-http-client/issues/523

0xTim:
Resolves #7


0xTim:
@adam-fowler @svanimpe you'll need to explicitly approve this as existing contributors as we're effectively relicensing it.

@swift-server/sswg thoughts on copyright notice etc?

LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
adam-fowler, line 189: @0xTim did you not just point out this issue in the AHC repo

0xTim: Yes, added this in case https://github.com/swift-server/async-http-client/issues/523#issuecomment-987971916 is right. Waiting for confirmation, but I won't merge until it's been OKed

scripts/generate_contributors_list.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
svanimpe, line 1: Is this a safe assumption? I've had some issues with this in the past, so I now use `/usr/bin/env bash`

adam-fowler (approved):
This is good to go when copyright holder is updated

LICENSE
ktoso: This is fine as is. More details: https://github.com/swift-server/async-http-client/issues/523#issuecomment-987971916

scripts/generate_contributors_list.sh
adam-fowler: I think this is fine as is. I'm guessing this is a direct copy of the swift-nio contributors code. It'll only ever be run by the maintainers of the project.

svanimpe (approved):

0xTim:
Cool think we're good, just waiting for final input from the SSWG

adam-fowler:
LGTM

PR #13, opened 2021-12-08
Add unit tests for getRepositoryName and fix #12

### LINKED ISSUE #12: If a package has a "." in it then it cannot be viewed in PackageDependency view
Load project https://github.com/soto-project/soto-core and try to open dependency `jmespath.swift`

svanimpe:
Relatively small fix:

Match the last path component first and only then strip out the `.git` extension. This allows dots in package names (to fix #12).

I've also took this opportunity to add some unit tests to the project 🎉


src/test/suite/utilities.test.ts
```diff
@@ -0,0 +1,33 @@
+import * as assert from 'assert';
+import { getRepositoryName } from '../../utilities';
+
+suite('Utilities Test Suite', () => {
```
adam-fowler, line 4: Guess we should be adding tests as a default, I've no idea how the test framework works in vscode though. Also would be good to add GitHub CI to run tests automatically on PRs (for another PR)

src/utilities.ts
```diff
@@ -19,6 +19,27 @@ export async function exec(command: string, options: cp.ExecOptions): Promise<{
...
+ * 
+ * The base name is the last path component of the URL, without the extension `.git`,
+ * and without an optional trailing slash.
+ */
+export function getRepositoryName(url: string): string {
+    // This regular expression consists of:
+    // - any number of characters that aren't a slash: ([^/]*)
+    // - optionally followed by a trailing slash: \/? 
+    // - at the end of the URL: $
+    const pattern = /([^/]*)\/?$/;
```
adam-fowler, line 33: Not a great fan of regular expressions. They are very opaque to users reading code. The comment above is very helpful

adam-fowler (approved):
Looks good

src/test/suite/utilities.test.ts
0xTim: I'll add this in a separate PR

0xTim (approved):

src/test/suite/utilities.test.ts
adam-fowler: @0xTim I meant for test the vscode-swift typescript not adding support for testing swift

0xTim: Yeah CI for the extension itself

src/utilities.ts
svanimpe: Yeah same here. I like what I can do what them, but I don't like reading them, hence the comment :)

PR #14, opened 2021-12-08
CoreLLDB integration

adam-fowler:
- add CoreLLDB to packageDependencies in `package.json`
- Added command to build `launch.json` based off contents of Package.swift
- Updating Package.swift will run `swift package describe` and update `launch.json` if necessary

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,8 +34,8 @@
         "category": "Swift"
       },
       {
-        "command": "swift.makeDebugConfig",
-        "title": "Generate launch configuration",
+        "command": "swift.generateLaunchConfig",
+        "title": "Generate Launch Configuration",
         "category": "Swift"
       }
     ],
@@ -97,7 +97,7 @@
           "when": "false"
         },
         {
-          "command": "swift.makeDebugConfig",
+          "command": "swift.generateLaunchConfig",
           "when": "swift.hasPackage"
         }
       ],
--- a/package.json
+++ b/package.json
@@ -149,6 +149,9 @@
       }
     ]
   },
+  "extensionDependencies": [
+    "vadimcn.vscode-lldb"
+  ],
   "scripts": {
     "vscode:prepublish": "npm run compile",
     "compile": "tsc -p ./",
--- a/package.json
+++ b/package.json
@@ -32,6 +32,11 @@
         "title": "Update Package Dependencies",
         "icon": "$(cloud-download)",
         "category": "Swift"
+      },
+      {
+        "command": "swift.makeDebugConfig",
+        "title": "Generate launch configuration",
+        "category": "Swift"
       }
     ],
     "configuration": [
@@ -90,6 +95,10 @@
         {
           "command": "swift.updateDependencies",
           "when": "false"
+        },
+        {
+          "command": "swift.makeDebugConfig",
+          "when": "swift.hasPackage"
         }
       ],
       "view/title": [
--- a/src/PackageWatcher.ts
+++ b/src/PackageWatcher.ts
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import * as vscode from 'vscode';
+import * as debug from './debug';
 import commands from './commands';
 import { SwiftContext } from './SwiftContext';
 
@@ -54,9 +55,9 @@ export class PackageWatcher {
         const watcher = vscode.workspace.createFileSystemWatcher(
             new vscode.RelativePattern(this.workspaceRoot, 'Package.swift')
         );
-        watcher.onDidCreate(async () => await this.handlePackageChange());
-        watcher.onDidChange(async () => await this.handlePackageChange());
-        watcher.onDidDelete(async () => await this.handlePackageChange());
+        watcher.onDidCreate(async () => await this.handlePackageSwiftChange());
+        watcher.onDidChange(async () => await this.handlePackageSwiftChange());
+        watcher.onDidDelete(async () => await this.handlePackageSwiftChange());
         return watcher;
     }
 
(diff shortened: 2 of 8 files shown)


0xTim (approved):
LGTM

src/commands.ts
```diff
@@ -27,6 +29,12 @@ const commands = {
         vscode.tasks.executeTask(task);
     },
 
+    makeDebugConfig(ctx: SwiftContext): Command {
```
svanimpe, line 32: Perhaps we don't need this factory if we can just use argument binding instead.

Can you try defining the command as:

```typescript
async makeDebugConfig(ctx: SwiftContext) {
    return await debug.makeDebugConfigurations(ctx);
},
```

and then registering it as:

```typescript
vscode.commands.registerCommand('swift.makeDebugConfig',
  this.makeDebugConfig.bind(this, ctx)
)
```

This works for me, but I'm not sure it's 100% what you want.

adam-fowler: Didn't know about the argument binding, will change.

adam-fowler:
@svanimpe anything holding this back?

package.json
```diff
@@ -32,6 +32,11 @@
         "title": "Update Package Dependencies",
         "icon": "$(cloud-download)",
         "category": "Swift"
+      },
+      {
+        "command": "swift.makeDebugConfig",
+        "title": "Generate launch configuration",
```
svanimpe, line 38: We should use Title Case here for consistency (Generate Launch Configuration). Also, it's a bit weird that it's called Launch Configuration externally, but `debugConfig` internally (also in all the methods) 😅

package.json
```diff
@@ -140,6 +149,9 @@
       }
     ]
   },
+  "extensionDependencies": [
+    "vadimcn.vscode-lldb"
```
svanimpe, line 153: This only prompted me to install CodeLLDB. The documentation I found on this was a bit unclear, but it seems like there's no way to silently install it as a dependency, without prompting the user first?

src/PackageWatcher.ts
```diff
@@ -76,8 +77,25 @@ export class PackageWatcher {
...
+        // with package resolution
+        debug.makeDebugConfigurations(this.ctx);
+        // if package has dependencies resolve them
+        if (this.ctx.swiftPackage.dependencies.length > 0) {
+            await commands.resolveDependencies();
+        }
+    }
+
+    /**
+     * Handles a create or change event for **Package.swift** and **Package.resolved**.
```
svanimpe, line 93: This documentation (and the one for `handlePackageSwiftChange`) is now a bit out-of-date because we have separate methods for both files.

src/SwiftPackage.ts
```diff
@@ -94,6 +94,24 @@ export class SwiftPackage implements PackageContents {
         return this.contents?.targets ?? [];
     }
 
+    get executableProducts(): Product[] {
+        return this.products.filter((product, index, array) => {
```
svanimpe, line 98: We can simplify this to:

```typescript
return this.products.filter(product => product.type.executable !== undefined);
```

Same for the other properties.

src/SwiftTaskProvider.ts
```diff
@@ -45,8 +45,8 @@ function createCleanTask(): vscode.Task {
 /**
  * Creates a {@link vscode.Task Task} to run an executable target.
  */
- function createExecutableTask(target: Target): vscode.Task {
-    return createSwiftTask('swift', ['run', target.name], `Run ${target.name}`, vscode.TaskGroup.Build);
+ function createExecutableTask(product: Product): vscode.Task {
```
svanimpe, line 48: The function name and documentation are out-of-sync with its implementation. Maybe it should be `createBuildTask` now?

src/debug.ts
```diff
@@ -0,0 +1,62 @@
...
+import { SwiftContext } from './SwiftContext';
+
+// Edit launch.json based on contents of Swift Package
+// Adds launch configurations based on the executables in Package.swift
+export async function makeDebugConfigurations(ctx: SwiftContext) {
+    const wsLaunchSection = vscode.workspace.getConfiguration("launch", vscode.Uri.file(ctx.workspaceRoot));
+    const launchConfigs = wsLaunchSection.get<any[]>("configurations") || [];
+
+    let configs = createDebugConfigurations(ctx);
+    var edited = false;
```
svanimpe, line 25: Most likely, this should be a `let` variable?

src/debug.ts
```diff
@@ -0,0 +1,62 @@
...
+// Return array of DebugConfigurations based on what is in Package.swift
+function createDebugConfigurations(ctx: SwiftContext): vscode.DebugConfiguration[] {
+    const executableProducts = ctx.swiftPackage.executableProducts;
+
+    return executableProducts.map((product) => {
+        return {
+            type: "lldb",
+            request: "launch",
+            name: `Run ${product.name}`,
+            program: "${workspaceFolder}/.build/debug/" + product.name,
```
svanimpe, line 55: I was surprised to see this also works when a product contains a space in its name. Even on Windows :)

Tasks handle this by strong quoting the command arguments, but I'm not sure who/what is handling it now.

package.json
adam-fowler: That's a little frustrating. At least telling the user is worthwhile

svanimpe:
When we merge this, I also need to update the README to add the new features and change the documentation on tasks. The section on adding command-line arguments will also move to **launch.json**.

We should also discussing setting up **lldb.library**. My plan was just to point users to the **Use Alternate Backend** command that comes with CodeLLDB, and if that fails, set it manually. Any thoughts?

Finally, we can also add the "Debugger" category to **package.json** now I think?

src/PackageWatcher.ts
adam-fowler: ok will fix

package.json
adam-fowler: ok will fix

src/SwiftPackage.ts
adam-fowler: good point

src/SwiftTaskProvider.ts
adam-fowler: ok

svanimpe (approved):

PR #16, opened 2021-12-08
Updates to package.json

adam-fowler:
Adds icon, description, publisher, keywords and flags it as a preview


0xTim (approved):

package.json
```diff
@@ -13,6 +15,11 @@
   "categories": [
     "Programming Languages"
   ],
+  "keywords": [
+    "swift",
+    "sswg"
```
svanimpe, line 20: Just wondering if this does anything? I don't assume anyone will ever explicitly search for `sswg`, and if they do, most likely it's unrelated to Swift?

svanimpe (approved):

package.json
adam-fowler: SSWG at the moment is the publisher name, we can remove at a later date if deemed unnecessary

PR #19, opened 2021-12-08
Add Basic CI

### LINKED ISSUE #15: Use CI to run tests on PRs

0xTim:
Taken from https://github.com/microsoft/vscode-platform-specific-sample/blob/main/.github/workflows/ci.yml as suggested in the docs.

Currently just builds the extension on an array of platforms and architectures and has dormant code to enable publishing when it comes to it.

Also includes dependabot to keep things up to date

Resolves #15


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,75 @@
...
+            platform: darwin
+            arch: arm64
+            npm_config_arch: arm64
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14.x
+      - run: npm install
```
adam-fowler, line 55: Need to call `npm run package` to actually compile the code. But we need the license file first or it'll get stuck waiting to say are you ok with there being no license file

0xTim: Yeah I'm planning on doing it with `vsce package` but as you said we need a license first

svanimpe (approved):

.github/workflows/ci.yaml
adam-fowler: Now we have a LICENSE you can change that to `vsce package`. Although you'll need to merge main in first

adam-fowler:
I know this is probably not going to get merged and CI will be elsewhere. But I thought I'd some comments. Tests should run tests. To run tests you can use
```
npm run test
```
We should also run tests across multiple versions of Swift. Perhaps 5.4 onwards. If we could have a regular (nightly/weekly) build that runs with the latest development build, next release development build that'd be great.

0xTim:
Closing in favour of #177

PR #20, opened 2021-12-08
SourceKit LSP integration

adam-fowler:
Copied code straight from apple [repo](https://github.com/apple/sourcekit-lsp/tree/main/Editors/vscode). Keeping identical at the moment just in case they edit their version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,8 +41,41 @@
               ".git",
               ".github"
             ]
-          }
+          },
+          "sourcekit-lsp.serverPath": {
+            "type": "string",
+            "default": "sourcekit-lsp",
+            "description": "The path of the sourcekit-lsp executable"
+          },
+          "sourcekit-lsp.serverArguments": {
+            "type": "array",
+            "default": [],
+            "items": {
+                "type": "string"
+            },
+            "description": "Arguments to pass to sourcekit-lsp. Argument keys and values should be provided as separate entries in the array e.g. ['--log-level', 'debug']"
+          },
+          "sourcekit-lsp.toolchainPath": {
+            "type": "string",
+            "default": "",
+            "description": "(optional) The path of the swift toolchain. By default, sourcekit-lsp uses the toolchain it is installed in."
+          },
+          "sourcekit-lsp.inlayHints.enabled": {
+            "type": "boolean",
+            "default": false,
+            "description": "(experimental) Render inlay type annotations in the editor."
+          },
+          "sourcekit-lsp.trace.server": {
+            "type": "string",
+            "default": "off",
+            "enum": [
+                "off",
+                "messages",
+                "verbose"
+            ],
+            "description": "Traces the communication between VS Code and the SourceKit-LSP language server."
         }
+}
       }
     ],
     "menus": {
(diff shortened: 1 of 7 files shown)


0xTim:
This will probably need attribution as it's copied directly with a `NOTICE.txt`, e.g. https://github.com/apple/swift-crypto/blob/main/NOTICE.txt

0xTim (approved):

package.json
```diff
@@ -41,8 +41,41 @@
               ".git",
               ".github"
             ]
-          }
+          },
+          "sourcekit-lsp.serverPath": {
```
svanimpe, line 45: We can probably make his a lot more user friendly by grouping the settings into a submenu and referring to the "Language Server" instead of "SourceKit-LSP". I can take a stab at that if you want.

src/sourcekit-lsp/extension.ts
```diff
@@ -0,0 +1,56 @@
...
+        // eslint-disable-next-line @typescript-eslint/naming-convention
+        sourcekit.options = { env: { ...process.env, SOURCEKIT_TOOLCHAIN_PATH: toolchain } };
+    }
+
+    const serverOptions: langclient.ServerOptions = sourcekit;
+
+    let clientOptions: langclient.LanguageClientOptions = {
+        documentSelector: [
+            'swift',
+            'cpp',
```
svanimpe, line 39: We probably only need `swift` here, as we don't activate the extension for the other file types.

svanimpe:
As a general question: do we want to leave the imported source code as is for the time being, or can we start integrating it with the rest of the extension (the configuration keys for example)?

0xTim:
@svanimpe I think we should merge as close to the copied code as possible to make following the attribution easy and then integrate it properly in a future PR

src/sourcekit-lsp/extension.ts
0xTim: We should activate the extension for all languages that SwiftPM supports. Building something like JWTKit which has a ton of code in would be hard if we didn't have that enabled

package.json
adam-fowler: As with code I want to keep these as close to original.

src/sourcekit-lsp/extension.ts
svanimpe: The SourceKit-LSP extension also has activation events for these languages, but these haven't been imported yet. Would you also want those?

I'm just worried this may turn out to be annoying for developers also working on C/C++ projects. I did some work in C last year, and I always had to disable SourceKit-LSP because it kept popping up errors in a valid C project.

0xTim: This is partly related to #21 and others but we probably need to try and be smart about activating the extension. The current Swift extension throws up errors every time you open _any_ workspace and it can't detect a Package.swift. If we make it clever enough so that the extension (and SourceKit LSP and others) are only activated on detection of a top level manifest (plus whatever we decide for #21) that should stop it being a problem

svanimpe (approved):

PR #25, opened 2021-12-09
Add Problem matchers to report errors in Package.swift

adam-fowler:
Add two problem matchers to report errors in Package.swift. Had to add two matchers as the package matchers outputs two formats of error.

Also `SwiftPackage.LoadPackage()` now catches the situation where there is a package or not. Returns `null` to indicate no Package, `NullPackage` indicates we have a package but failed to load it and then a package. Previous when a package failed to load it was as if it didn't exist. So the resolve would not run.

One issue left I have no idea how to resolve: I can't extract a filename from the errors of the form
```
<Package.swift folder>: error: <there is a problem>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,6 +109,32 @@
         }
       ]
     },
+    "problemMatchers": [
+      {
+        "name": "package-swift",
+        "owner": "swift",
+        "fileLocation": ["absolute"],
+        "pattern": {
+          "regexp": "^([a-zA-Z0-9.\/]*):\\s+(warning|error):\\s+(?!manifest parse)(.*)$",
+          "file": 1,
+          "severity": 2,
+          "message": 3
+        }
+      },
+      {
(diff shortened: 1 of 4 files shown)


svanimpe:
I just had a similar problem with problem matchers. They require a filename to be matched, otherwise, the file name defaults to the first capture group (I think), which will be incorrect. I tried hacking something together with an empty capture group but that didn't work.

I fear problem matchers aren't suitable for this kind of problem reporting.

adam-fowler:
> I fear problem matchers aren't suitable for this kind of problem reporting.

If SPM had consistent error messaging we could get this to work, all the time. I might see if I can add a PR to the Swift Package Manager to do this

adam-fowler:
Closing this for the moment until we come up with a better idea. I've extracted code I want to use elsewhere into separate PRs #34 and #35

PR #26, opened 2021-12-09
Stop multiple resolves running at same time

adam-fowler:
- Add function `executeTaskAndWait` which runs a task and waits for it to finish.
- Use this function in resolve/UpdateDependencies.
- Add global flags to only allow entry into these functions if they are not already running

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -15,6 +15,7 @@
 import * as vscode from 'vscode';
 import * as debug from './debug';
 import { SwiftContext } from './SwiftContext';
+import { executeTaskAndWait } from './SwiftTaskProvider';
 
 /**
  * References:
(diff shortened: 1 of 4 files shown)


src/commands.ts
```diff
@@ -23,35 +24,50 @@ import * as vscode from 'vscode';
  *   https://code.visualstudio.com/api/extension-guides/command
  */
 
-/**
+// flags to indicating whether a resolve or update is in progress
+var resolveRunning = false;
```
svanimpe, line 28: Do we need these to be globals, or can we move them into `commands` as properties?

adam-fowler: It wasn't happy having them inside `commands`. Commands is not a class

svanimpe: We can upgrade it to a class and just export a singleton object, like what I did with the `StatusItem`? That would allow us to make these flags `private`.

src/commands.ts
```diff
@@ -23,35 +24,50 @@ import * as vscode from 'vscode';
...
+        await executeTaskAndWait(task);
+
+        resolveRunning = false;
     },
 
     /**
      * Executes a {@link vscode.Task task} to update this package's dependencies.
      */
     async updateDependencies() {
+        if (updateRunning) { return; }
```
0xTim, line 59: Do we want to allow an update if a resolve is already running? That might cause issues but I think it's probably ok

adam-fowler: We probably need better management of all the `swift` calls. Maybe a queue to avoid calls happening concurrently. But also ways to skip unnecessary calls. But we should leave that until a later PR.

src/commands.ts
adam-fowler: I upgraded it to a class and made all the members static, so we don't need to drag a singleton around all over the place

svanimpe (approved):

PR #28, opened 2021-12-09
Add a StatusBarItem to show that a resolve task is in progress

svanimpe:
The goal of this PR was to silence the resolve tasks, so they don't always pop up the terminal. To that end, I've added a `StatusBarItem` that shows that a task is in progress. Only when the task returns an error code will it pop up the terminal.

This is only for resolve tasks for now, but we could extend it to other tasks as well. Although I'm not yet sure how to handle multiple tasks running in parallel (if VS Code even supports that).


0xTim (approved):
Description looks like it got cut off but otherwise LGTM

adam-fowler:
You might want to review #26. In it I have added a function that executes a task and then waits for it to complete. I then use this function in the resolve command. Instead of having yet another listener waiting for a resolve task to complete you could put your statusItem show / hide either side of the `executeTaskAndWait` function.

src/extension.ts
```diff
@@ -68,7 +69,9 @@ export async function activate(context: vscode.ExtensionContext) {
  * Deactivate the extension.
  * 
  * Any disposables registered in `context.subscriptions` will be automatically
- * disposed of, so there's nothing left to do here.
+ * disposed of.
  */
-export function deactivate() {}
+export function deactivate() {
+	statusItem.dispose();
```
adam-fowler, line 75: Any reason why you aren't just adding this to context.subscriptions. You'd need to conform `StatusItem` to `Disposable`.

src/commands.ts
```diff
@@ -38,6 +41,10 @@ const commands = {
             task.definition.args[0] === 'package' &&
             task.definition.args[1] === 'resolve'
         )!;
+        task.presentationOptions = {
+            reveal: vscode.TaskRevealKind.Silent
```
adam-fowler, line 45: If we get the problem matcher working we can default the resolve task to silent i.e. set the presentation option at creation

adam-fowler (changes requested):

src/extension.ts
svanimpe: No particular reason. I wasn't sure where to place the global `statusItem` object, and since it isn't created in the activation function, I just disposed of it here.

src/commands.ts
svanimpe: I was thinking if perhaps we should treat user-initiated tasks separately, and never silence them.

I also saw #31, but I don't know of a way to apply that to user-initiated tasks, so they may be treated differently there as well?

src/extension.ts
adam-fowler: You can't have it as a global. If you load a new workspace the item will have dispose called on it when deactivate is called when shutting down the first workspace. Then for the new workspace you will have a disposed StatusItem. 

You need to create a new StatusItem every time activate is called. Maybe you can store the StatusItem in the context, so it can be accessed. Also maybe wait for #32 as that changes the structure of the context.

svanimpe: Hmm, when I add logging to the constructor of the `statusItem`, I can see that VS Code recreates the extension (and the `statusItem`):
- When I go from a single folder to an unnamed workspace
- When I remove the first (top most) folder from a workspace

This is also mentioned in the API docs for `onDidChangeWorkspaceFolders`:

> onDidChangeWorkspaceFolders: Event<WorkspaceFoldersChangeEvent>
> 
> An event that is emitted when a workspace folder is added or removed.
> 
> Note: this event will not fire if the first workspace folder is added, removed or changed, because in that case the currently executing extensions (including the one that listens to this event) will be terminated and restarted so that the (deprecated) rootPath property is updated to point to the first workspace folder.

adam-fowler: Yeah moving between single folder workspace to multi-folder workspace creates a new session. I got slightly thrown by that when working on #32. 

Still the StatusItem lifecycle should be linked to the lifecycle of the session.

src/commands.ts
adam-fowler: Might as well set the `presentationOptions` when the resolve task is created in SwiftTaskProvider.ts

adam-fowler: Can we add this to the update dependencies task as well?

src/statusItem.ts
```diff
@@ -0,0 +1,42 @@
...
+export class StatusItem {
+
+    private item: vscode.StatusBarItem;
+
+    constructor() {
+        this.item = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left);
+        this.item.command = `terminal.focus`;
+    }
+
+    show(message: string, addProgressIndicator = true) {
```
adam-fowler, line 30: If you have multiple tasks running at the same time, the first task that finishes will hide the `statusItem`. 

Should we keep a record of tasks that have shown the status item in an array and then in `hide` remove items from array and only when the array is empty hide the status item.

src/commands.ts
svanimpe: One issue I see with adding `Silent` as the default presentation is that users who trigger the task from the **Run Task...** menu will not get much feedback when they start the task. It may be better to stick to the default presentation (pop up the terminal) when the user actively initiates the task?

For the update command: we could do this, but this command is currently only used for the Update Dependencies button in the Dependencies view, so it's also user-initiated.

src/statusItem.ts
svanimpe: I've added this, although it's not that useful for now. I've tested it by disabling the duplicate resolve/update checks and triggering some resolved while an update was running.

src/commands.ts
adam-fowler: There is no reason for the user to trigger resolve from the "run task..." menu, so I don't think this is an issue. Anyway there is only one copy of the resolve task and it runs straight off and you set it to silent. So it would be silent from then on regardless of how it was triggered.
 
The update command it is just a nicety. We have the status item we might as well use it.

svanimpe: OK, I'll add it for the update command.

I did notice that the task provider runs every time we call `vscode.tasks.fetchTasks()` to launch a command. So every command does run a different task object, with its own presentation settings.

adam-fowler: It doesn't always call run the task provider I've found.

adam-fowler: One last thing can we move the `start` to before the `try` and the `end` after the catch block. If the `executeTaskAndWait` should throw an error (not that it should) `statusItem.end` would never get called

svanimpe: Can anything other than `executeAndWait` throw an error? I'd need to move the `try` down a bit because I need the `task` object.

adam-fowler: @svanimpe just skip that. It won't work as expected. I think this is good to go once the update command is done.

adam-fowler (approved):

adam-fowler (approved):
Changes look good

PR #32, opened 2021-12-11
Basic Multi folder support

adam-fowler:
- Split SwiftContext in two: WorkspaceContext for whole workspace and FolderContext for each folder.
- Catch when folders are added/removed from project. Create/delete FolderContext on add/remove
- Add observers to be called on add/remove of project
- Add logObserver that logs adding and removing of folders
- Add observer that adds the dependency view for the root folder
- Add observer that calls resolve and make debug config for the root folder
- Edit TaskProvider to returns tasks based on contents of WorkspaceContext

In effect what we had before, but now we can extend this for multi root projects

I see two types of multi-folder project. One where you are editing dependencies of a project locally and one where the folders are separate projects. The first is fairly easy to setup. We can add an observer that calls `swift package edit/unedit` when folders are added and removed. The second is harder as currently Tasks run in the workspace root. We will a lot more work to get that working.


src/FolderContext.ts
```diff
@@ -0,0 +1,50 @@
...
+//===----------------------------------------------------------------------===//
+
+import * as vscode from 'vscode';
+import { PackageWatcher } from './PackageWatcher';
+import { SwiftPackage } from './SwiftPackage';
+
+export class FolderContext implements vscode.Disposable {
+    private packageWatcher?: PackageWatcher;
+
+	private constructor(
```
0xTim, line 22: Inconsistent formatting
```suggestion
    private constructor(
```

0xTim (changes requested):
Small nit, otherwise LGTM. 

For workspaces that contain multiple projects maybe we can add a top level `launch.json` file the extension can check for and know if it needs to be initialised

adam-fowler:
> For workspaces that contain multiple projects maybe we can add a top level `launch.json` file the extension can check for and know if it needs to be initialised

When you save a multi-folder project it places the launch.json configs in the saved project. I haven't looked too much into this yet but that seems to be where they go.

src/extension.ts
```diff
@@ -28,40 +28,54 @@ export async function activate(context: vscode.ExtensionContext) {
...
+		console.log(`${operation}: ${folder.rootFolder.uri.fsPath}`);
+	});
 
-	// Create the Package Dependencies view.
-	const dependenciesProvider = new PackageDependenciesProvider(ctx);
-	const dependenciesView = vscode.window.createTreeView('packageDependencies', {
-		treeDataProvider: dependenciesProvider,
-		showCollapseAll: true
+	// observer that will add dependency view based on whether a root workspace folder has been added
+	let addDependencyViewObserver = workspaceContext.observerFolders((folder, operation) => {
```
svanimpe, line 47:
```suggestion
	const addDependencyViewObserver = workspaceContext.observerFolders((folder, operation) => {
```

src/extension.ts
```diff
@@ -28,40 +28,54 @@ export async function activate(context: vscode.ExtensionContext) {
...
+			});
+			context.subscriptions.push(dependenciesView);
+		}
 	});
 
-	// Watch for changes to Package.swift and Package.resolved.
-	const packageWatcher = new PackageWatcher(workspaceRoot, ctx);
-	packageWatcher.install();
+	// observer that will resolve package for root folder
+	let resolvePackageObserver = workspaceContext.observerFolders(async (folder, operation) => {
```
svanimpe, line 59:
```suggestion
	const resolvePackageObserver = workspaceContext.observerFolders(async (folder, operation) => {
```

src/extension.ts
```diff
@@ -28,40 +28,54 @@ export async function activate(context: vscode.ExtensionContext) {
...
+		if (folder.isRootFolder && operation === 'add') {
+			// resolve root package
+			if (folder.isRootFolder) {
```
svanimpe, line 62: Why this double check here? The type checker will already narrow the type of `isRootFolder` from `boolean` to `true` based on the first check.

src/extension.ts
```diff
@@ -28,40 +28,54 @@ export async function activate(context: vscode.ExtensionContext) {
...
+				await debug.makeDebugConfigurations(folder);
+				await commands.resolveDependencies();
+			}
+		}
+	});
 
-	// Initialize the context keys and trigger a resolve task if needed.
-	packageWatcher.handlePackageSwiftChange();
+	// add workspace folders, already loaded
+	if (vscode.workspace.workspaceFolders !== undefined) {
```
svanimpe, line 71:
```suggestion
	if (vscode.workspace.workspaceFolders) {
```

That's quite common in JS/TS. It checks for  both `null` and `undefined`.

src/FolderContext.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    async reload() {
+        this.swiftPackage.reload();
+        if (this.isRootFolder) {
+            this.setContextKeys();
+        }
+    }
+
+    private setContextKeys() {
+        if (this.swiftPackage.contents === undefined) {
```
svanimpe, line 60: This should use `swiftPackage.isValid()` when #34 is merged.

src/FolderContext.ts
```diff
@@ -0,0 +1,68 @@
...
+        rootFolder: vscode.WorkspaceFolder,
+        isRootFolder: boolean
+    ): Promise<FolderContext> 
+    {
+        let swiftPackage = await SwiftPackage.create(rootFolder);
+        return new FolderContext(rootFolder, swiftPackage, isRootFolder);
+    }
+
+    async reload() {
+        this.swiftPackage.reload();
```
svanimpe, line 53:
```suggestion
        await this.swiftPackage.reload();
```

src/WorkspaceContext.ts
```diff
@@ -0,0 +1,73 @@
...
+        this.folders.push(folderContext);
+        this.observers.forEach(async fn => { await fn(folderContext, 'add'); });
+    }
+
+    // remove folder from workspace
+    removeFolder(folder: vscode.WorkspaceFolder) {
+        // find context with root folder
+        let context = this.folders.find((context: FolderContext, _) => { return context.rootFolder === folder; });
+        if (context === undefined) {
+            console.error(`Trying to delete folder ${folder} which has no record`);
```
svanimpe, line 55: Can this happen?

src/WorkspaceContext.ts
```diff
@@ -0,0 +1,73 @@
...
+            return;
+        }
+        // run observer functions in reverse order when removing
+        let observersReversed = [...this.observers];
```
svanimpe, line 59: Is there a reason we do this? If so, we could add it as a comment.

src/FolderContext.ts
```diff
@@ -0,0 +1,50 @@
...
+    get isRootFolder(): boolean {
+        return this._isRootFolder;
+    }
+
+    static async create(
+        rootFolder: vscode.WorkspaceFolder,
+        isRootFolder: boolean
+    ): Promise<FolderContext> 
+    {
+        let swiftPackage = await SwiftPackage.create(rootFolder);
```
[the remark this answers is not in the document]
adam-fowler: That is fixed, for the moment. `FolderContext` sets the contextKeys only when the folder is a root folder. Eventually this will have to change to be based on the active file but for the meantime it is working. I removed the contextKey setting from `SwiftPackage`

src/WorkspaceContext.ts
adam-fowler: It has and I'm not sure why, wanted to try catch it

src/WorkspaceContext.ts
adam-fowler: We run the observers in the reverse order because if observer B is dependent on resources from observer A which is setup before it, those resources are not removed before the observer B is.

src/WorkspaceContext.ts
```diff
@@ -0,0 +1,72 @@
...
+        // find context with root folder
+        let context = this.folders.find((context: FolderContext, _) => { return context.rootFolder === folder; });
+        if (context === undefined) {
+            console.error(`Trying to delete folder ${folder} which has no record`);
+            return;
+        }
+        this.observers.forEach(fn => fn(context!, 'remove'));
+        context.dispose();
+        // remove context with root folder
+        this.folders = this.folders.filter(context => context.rootFolder !== folder);
```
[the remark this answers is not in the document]
adam-fowler: You can't remove a single index from an array it seems. The only way to remove an element in the middle of an array is via filter

src/WorkspaceContext.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    observerFolders(fn: (folder: FolderContext, operation: 'add'|'remove') => unknown): vscode.Disposable {
```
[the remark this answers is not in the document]
adam-fowler: ok

src/WorkspaceContext.ts
adam-fowler: No it hasn't but it is worthwhile keeping in there just in case

src/WorkspaceContext.ts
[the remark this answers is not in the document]
svanimpe: `this.folders.splice(index, 1)` doesn't work?

src/extension.ts
```diff
@@ -28,40 +28,54 @@ export async function activate(context: vscode.ExtensionContext) {
...
-		workspaceRoot = vscode.workspace.workspaceFolders[0].uri.fsPath;
-	}
-
-	// Features past this point require a workspace folder.
-	if (!workspaceRoot) {
-		return;
-	}
+	let workspaceContext = new WorkspaceContext(context);
+	let onWorkspaceChange = vscode.workspace.onDidChangeWorkspaceFolders((event) => {
+		if (workspaceContext === undefined) { console.log("Trying to run onDidChangeWorkspaceFolders on deleted context"); return; }
```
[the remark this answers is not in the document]
adam-fowler: Yes it has and I'm not sure why. The function gets called on an old `WorkspaceContext` that has been deleted.

src/extension.ts
```diff
@@ -28,40 +28,54 @@ export async function activate(context: vscode.ExtensionContext) {
...
+		workspaceContext.onDidChangeWorkspaceFolders(event);
```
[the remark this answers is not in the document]
adam-fowler: `WorkspaceContext.onDidChangeWorkspaceFolders` calls functions internal to `WorkspaceContext` so not really. I did originally have the `workspace.onDidChangeWorkspaceFolders` inside `WorkspaceContext` but moved it out as I was concerned about cyclical references

src/extension.ts
adam-fowler: hmm not sure why

src/extension.ts
adam-fowler: ok

src/WorkspaceContext.ts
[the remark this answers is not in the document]
adam-fowler: I guess you could use splice. I'll replace

src/extension.ts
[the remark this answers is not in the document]
svanimpe: Do you know how to reproduce it? The only thing I can think of is that maybe the context gets disposed before the listener. Perhaps those subscriptions also gets disposed in reversed order, and we're supposed to add them in order we declare them?

src/extension.ts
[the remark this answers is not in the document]
svanimpe: Ah OK. I only saw it calls `addFolder` and `removeFolder`, both of which are public.

svanimpe (approved):

src/extension.ts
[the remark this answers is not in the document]
adam-fowler: removeFolder is now private :-)

src/extension.ts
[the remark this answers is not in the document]
adam-fowler: It could be the order things are deleted. Of course as soon as I put the message in I never saw it again

PR #33, opened 2021-12-11
Get local dependencies from SwiftPackage (fix #30)

### LINKED ISSUE #30: PackageDependenciesProvider.getLocalDependencies should use `SwiftPackage`
No need to run another swift process to get local packages

svanimpe:
This gets the list of local dependencies from `SwiftPackage`. To detect a local dependency, I've checked that it does not have a `requirement`, but does have a `url`. Can we depend on this always being the case?

I've also fixed a small bug in that directories shouldn't have an Open File command attached to them.

As a side note: When does a dependency have a `path` property instead of a `url`? I haven't seen any paths so far.

### DIFF
--- a/src/SwiftPackage.ts
+++ b/src/SwiftPackage.ts
@@ -43,7 +43,6 @@ export interface Dependency {
     identity: string
     requirement?: object
     url?: string
-    path?: string
 }
 
 // Class holding Swift Package Manager Package
--- a/src/SwiftPackage.ts
+++ b/src/SwiftPackage.ts
@@ -41,6 +41,7 @@ export interface Target {
 // Swift Package Manager dependency
 export interface Dependency {
     identity: string
+    requirement?: object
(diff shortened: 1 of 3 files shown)


src/SwiftPackage.ts
```diff
@@ -41,6 +41,7 @@ export interface Target {
 // Swift Package Manager dependency
 export interface Dependency {
     identity: string
+    requirement?: object
     url?: string
     path?: string
```
adam-fowler, line 45: Might as well remove `path` given it is never used. I had just assumed the package description would include a package path item given that is the label used in the Package.swift. Looks like I was wrong.

adam-fowler (approved):
LGTM

src/SwiftPackage.ts
svanimpe: Done

PR #34, opened 2021-12-12
Add InvalidPackage for package that exists but failed to load

adam-fowler:
Currently if a Package.swift fails to load it is as if it doesn't exist

Sometimes we want to know there is a Package.swift but that it failed to load. So I created a `InvalidPackage` class which is returned for a Package.swift that fails to load. `SwiftPackage.contents` can now be one of three thing
- Contents of a valid Package.swift
- `InvalidPackage` meaning a Package.swift was there but failed to load
- null meaning there was no Package.swift

I then added a test `SwiftPackage.foundPackage` which returns true for valid or invalid contents but not `null`

I use this in the PackageWatcher to decide when I should run the resolve task

### DIFF
--- a/src/FolderContext.ts
+++ b/src/FolderContext.ts
@@ -53,7 +53,7 @@ export class FolderContext implements vscode.Disposable {
     }
 
     private setContextKeys() {
-        if (this.swiftPackage.foundPackage) {
+        if (this.swiftPackage.isValid) {
             contextKeys.hasPackage = true;
             contextKeys.packageHasDependencies = this.swiftPackage.dependencies.length > 0;  
         }
--- a/src/FolderContext.ts
+++ b/src/FolderContext.ts
@@ -53,12 +53,12 @@ export class FolderContext implements vscode.Disposable {
     }
 
     private setContextKeys() {
-        if (this.swiftPackage.contents === undefined) {
-            contextKeys.hasPackage = false;
-            contextKeys.packageHasDependencies = false;
+        if (this.swiftPackage.foundPackage) {
+            contextKeys.hasPackage = true;
+            contextKeys.packageHasDependencies = this.swiftPackage.dependencies.length > 0;  
         }
-        contextKeys.hasPackage = true;
-        contextKeys.packageHasDependencies = this.swiftPackage.dependencies.length > 0;  
-    }
+        contextKeys.hasPackage = false;
+        contextKeys.packageHasDependencies = false;
+}
 }
 
--- a/src/PackageWatcher.ts
+++ b/src/PackageWatcher.ts
@@ -77,7 +77,7 @@ export class PackageWatcher {
      * launch configuration if required and then resolve the package
      * dependencies.
      */
-     async handlePackageSwiftChange() {
+    async handlePackageSwiftChange() {
         // Load SwiftPM Package.swift description 
         await this.ctx.reload();
         // Create launch.json files based on package description. Run this in parallel
@@ -94,7 +94,7 @@ export class PackageWatcher {
      * 
      * This will resolve any changes in the Package.resolved.
      */
-     async handlePackageResolvedChange() {
+    private async handlePackageResolvedChange() {
         if (this.ctx.isRootFolder && this.ctx.swiftPackage.foundPackage) {
             await commands.resolveDependencies();
         }
--- a/src/PackageWatcher.ts
+++ b/src/PackageWatcher.ts
@@ -84,7 +84,7 @@ export class PackageWatcher {
         // with package resolution
         debug.makeDebugConfigurations(this.ctx);
         // if package has dependencies resolve them
-        if (this.ctx.isRootFolder && this.ctx.swiftPackage.dependencies.length > 0) {
+        if (this.ctx.isRootFolder && this.ctx.swiftPackage.foundPackage) {
             await commands.resolveDependencies();
         }
     }
@@ -95,7 +95,7 @@ export class PackageWatcher {
      * This will resolve any changes in the Package.resolved.
      */
      async handlePackageResolvedChange() {
-        if (this.ctx.isRootFolder && this.ctx.swiftPackage.dependencies.length > 0) {
+        if (this.ctx.isRootFolder && this.ctx.swiftPackage.foundPackage) {
             await commands.resolveDependencies();
         }
     }
--- a/src/SwiftPackage.ts
+++ b/src/SwiftPackage.ts
@@ -45,36 +45,52 @@ export interface Dependency {
     url?: string
 }
 
+// package we attempted to load but failed
+class NullPackage implements PackageContents {
+    get name(): string { return ""; }
+    get products(): Product[] { return []; }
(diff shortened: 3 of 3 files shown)


src/SwiftPackage.ts
```diff
@@ -45,39 +45,55 @@ export interface Dependency {
     url?: string
 }
 
+// package we attempted to load but failed
+class InvalidPackage implements PackageContents {
```
svanimpe, line 49: Do you intend to use this class by name? If not, we don't really need a class or something that explicitly implements `PackageContents`. We could just return:

```javascript
{
  name: '',
  products: [],
  dependencies: [],
  targets: []
}
```

from `loadPackage`. That will also conform to `PackageContents`.

src/SwiftPackage.ts
```diff
@@ -45,39 +45,55 @@ export interface Dependency {
...
+    get dependencies(): Dependency[] { return []; }
+    get targets(): Target[] { return []; }
+}
+
 // Class holding Swift Package Manager Package
 export class SwiftPackage implements PackageContents {
 	private constructor(
         readonly folder: string,
-        public contents?: PackageContents
+        public contents?: PackageContents|null
```
svanimpe, line 60: Do you just want this to be an optional? If so, it should just be `contents?: PackageContents`. The type will then be `PackageContents | undefined`. Currently, it's `PackageContents | undefined | null`. Or do you want to treat `null` differently from `undefined`?

src/SwiftPackage.ts
```diff
@@ -45,39 +45,55 @@ export interface Dependency {
...
     ) {
         this.setContextKeys();
     }
 
     public static async create(folder: string): Promise<SwiftPackage> {
+        let contents = await SwiftPackage.loadPackage(folder);
+        return new SwiftPackage(folder, undefined);
```
svanimpe, line 67: `undefined` should be `contents` here I think?

adam-fowler: Good catch, cut and paste coding :-(

src/SwiftPackage.ts
adam-fowler: I struggled to return a `Promise<undefined>` from `loadPackage` so just switched to `null`

src/SwiftPackage.ts
adam-fowler: Just added SwiftPackage.isValid to differentiate between packages that loaded and failed to load

src/SwiftPackage.ts
svanimpe: Perhaps we could just use `PackageContents | null | undefined` to differentiate between the three states (valid / invalid / no package.swift)?
The getters already return the default values, so it feels a bit redundant to add an `InvalidPackage` class just to indicate a state.

Something like this:

```typescript
export class SwiftPackage implements PackageContents {
    private constructor(
        readonly folder: string,
        public contents: PackageContents | null | undefined
    ) {
        this.setContextKeys();
    }

    public static async create(folder: string): Promise<SwiftPackage> {
        let contents = await SwiftPackage.loadPackage(folder);
        return new SwiftPackage(folder, contents);
    }

    public static async loadPackage(folder: string): Promise<PackageContents | null | undefined> {
        try {
            const { stdout } = await exec('swift package describe --type json', { cwd: folder });
            return JSON.parse(stdout);
        } catch(error) {
            const execError = error as {stderr: string};
            if (execError.stderr.startsWith("error: root manifest")) {
                return undefined;
            } else {
                return null;
            }
        }
    }

    // ...

    // Return if Package.swift is valid
    public get isValid(): boolean {
        return this.contents ? true : false;
    }

    // Did we find a Package.swift
    public get foundPackage(): boolean {
        return this.contents !== undefined;
    }

   // ...

    private setContextKeys() {
        if (!this.contents) {
            contextKeys.hasPackage = false;
            contextKeys.packageHasDependencies = false;
        }
        contextKeys.hasPackage = true;
        contextKeys.packageHasDependencies = this.dependencies.length > 0;  
    }
}
```

adam-fowler: I did think about using `null` for a package that failed to load but initially felt it wasn't explicit enough in indicating a package failed to load. That's why I came up with the new type but given this is all internal I guess it wouldn't hurt just using `null`

src/PackageWatcher.ts
```diff
@@ -78,14 +78,14 @@ export class PackageWatcher {
      * launch configuration if required and then resolve the package
      * dependencies.
      */
-     async handlePackageSwiftChange() {
```
0xTim, line 81: We should look at integrating a linter/formatter for these kind of issues, I'll create an issue

src/SwiftPackage.ts
0xTim: In Swift we'd probably use an enum so I'd make this as explicit as possible

0xTim (approved):
In general LGTM, comments added

src/PackageWatcher.ts
adam-fowler: I've been running with ESLint, which seems to be the standard Linter. But it doesn't catch these annoying indent issues

src/SwiftPackage.ts
svanimpe: Unfortunately, enums in TypeScript aren't as powerful as they are in Swift, so we'd have to add an enum field just for the state, and store the actual state separately in `contents`. But that means we could have an inconsistent state where, for example, `state === .valid` but `contents === undefined`.

src/FolderContext.ts
```diff
@@ -53,12 +53,12 @@ export class FolderContext implements vscode.Disposable {
     }
 
     private setContextKeys() {
-        if (this.swiftPackage.contents === undefined) {
-            contextKeys.hasPackage = false;
-            contextKeys.packageHasDependencies = false;
+        if (this.swiftPackage.foundPackage) {
```
svanimpe, line 56: Shouldn't we use `isValid` here? Setting `hasPackage` will trigger the task provider and the dependency view, but neither will work properly if the package is invalid.

src/SwiftPackage.ts
```diff
@@ -49,32 +49,45 @@ export interface Dependency {
...
-            this.contents = undefined;
-        }
+    // Return if Package.swift is valid
+    public get isValid(): boolean {
+        return this.contents !== null && this.contents !== undefined;
     }
 
+    // Did we find a Package.swift
+    public get foundPackage(): boolean {
+        return this.contents !== null;
```
svanimpe, line 88:
```suggestion
        return this.contents !== undefined;
```

as undefined = no package found

But do we even need that anywhere? Shouldn't all features that depend on packages require a valid package?

src/FolderContext.ts
adam-fowler: I'm not sure either way. There was a point when `contextKeys.hasPackage` was the gate for running a resolve. Currently we need a resolve to run to report back errors in the `Package.swift`. We need to improve the error reporting there but that is for another time. Anyway changed to `isValid`

svanimpe (approved):

PR #41, opened 2021-12-14
Add Swift Output channel

adam-fowler:
And log everything we do in it. This will give us a useful log for when things go wrong in the wild.

You can find it in the dropdown list of the Output tab. It is called "Swift" funnily enough

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,1 @@
+invalid
--- a/Package.swift
+++ b/Package.swift
@@ -1,1 +1,0 @@
-invalid
--- a/src/FolderContext.ts
+++ b/src/FolderContext.ts
@@ -15,6 +15,7 @@
 import * as vscode from 'vscode';
 import { PackageWatcher } from './PackageWatcher';
 import { SwiftPackage } from './SwiftPackage';
+import { WorkspaceContext } from './WorkspaceContext';
 import contextKeys from './contextKeys';
 
 export class FolderContext implements vscode.Disposable {
@@ -23,10 +24,11 @@ export class FolderContext implements vscode.Disposable {
 	private constructor(
         public folder: vscode.WorkspaceFolder,
         public swiftPackage: SwiftPackage,
-        readonly isRootFolder: boolean
+        readonly isRootFolder: boolean,
+        workspaceContext: WorkspaceContext
     ) {
         if (this.isRootFolder) {
-            this.packageWatcher = new PackageWatcher(this);
+            this.packageWatcher = new PackageWatcher(this, workspaceContext);
             this.packageWatcher.install();
             this.setContextKeys();
         }
@@ -38,11 +40,12 @@ export class FolderContext implements vscode.Disposable {
 
     static async create(
         rootFolder: vscode.WorkspaceFolder,
-        isRootFolder: boolean
+        isRootFolder: boolean,
+        workspaceContext: WorkspaceContext
     ): Promise<FolderContext> 
     {
         let swiftPackage = await SwiftPackage.create(rootFolder);
-        return new FolderContext(rootFolder, swiftPackage, isRootFolder);
+        return new FolderContext(rootFolder, swiftPackage, isRootFolder, workspaceContext);
     }
 
     async reload() {
--- a/src/SwiftOutputChannel.ts
+++ b/src/SwiftOutputChannel.ts
@@ -13,9 +13,8 @@
 //===----------------------------------------------------------------------===//
 
 import * as vscode from 'vscode';
-import { Disposable } from 'vscode-jsonrpc';
 
-export class SwiftOutputChannel implements Disposable {
+export class SwiftOutputChannel {
     private channel: vscode.OutputChannel;
 
     constructor() {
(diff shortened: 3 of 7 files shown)


Package.swift
```diff
@@ -0,0 +1 @@
+invalid
```
0xTim, line 1: Was this meant to be added?

0xTim (changes requested):
LGTM other than the query around the manifest

Package.swift
adam-fowler: Whoops that bled in from a test branch.

src/SwiftOutputChannel.ts
```diff
@@ -0,0 +1,50 @@
...
+// See CONTRIBUTORS.txt for the list of VSCode Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import * as vscode from 'vscode';
+import { Disposable } from 'vscode-jsonrpc';
+
+export class SwiftOutputChannel implements Disposable {
```
svanimpe, line 18: We could just name this `Logger`, since that's basically what it does?

src/SwiftOutputChannel.ts
```diff
@@ -0,0 +1,50 @@
...
+    constructor() {
+        this.channel = vscode.window.createOutputChannel("Swift");
+    }
+
+    dispose() {
+        this.channel.dispose();
+    }
+
+    log(message: string, folder?: string) {
+        if (folder !== undefined) {
```
svanimpe, line 30:
```suggestion
        if (folder) {
```

src/SwiftOutputChannel.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    constructor() {
+        this.channel = vscode.window.createOutputChannel("Swift");
+    }
+
+    dispose() {
+        this.channel.dispose();
+    }
+
+    log(message: string, folder?: string) {
```
svanimpe, line 29: Perhaps `folder` can just be of type `WorkspaceFolder`, so we don't always have to find its path when we call `log`?

src/SwiftOutputChannel.ts
```diff
@@ -0,0 +1,50 @@
...
+        if (folder !== undefined) {
+            folder += ": ";
+        }
+        this.channel.appendLine(`${this.nowFormatted}: ${folder ?? ""}${message}`);
+    }
+
+    logStart(message: string, folder?: string) {
```
svanimpe, line 36: Can we automate part of this start/end combination? Perhaps adding the triple dots automatically, and letting `logEnd` print either an optional error parameter, or a default "Done" message for consistency?

src/extension.ts
```diff
@@ -40,7 +40,7 @@ export async function activate(context: vscode.ExtensionContext) {
 
 	// observer for logging workspace folder addition/removal
 	const logObserver = workspaceContext.observerFolders((folder, operation) => {
```
svanimpe, line 42:
```suggestion
	const logObserver = workspaceContext.observerFolders((folderContext, operation) => {
```

Maybe that would clear up the `folder.folder` references.

src/FolderContext.ts
```diff
@@ -23,10 +24,11 @@ export class FolderContext implements vscode.Disposable {
 	private constructor(
         public folder: vscode.WorkspaceFolder,
         public swiftPackage: SwiftPackage,
-        readonly isRootFolder: boolean
+        readonly isRootFolder: boolean,
+        workspaceContext: WorkspaceContext
```
svanimpe, line 28: We could upgrade this to a property. That way, we don't have to pass both the folder context and the workspace context to PackageWatcher.

adam-fowler: That'd be a cyclical reference

src/SwiftOutputChannel.ts
adam-fowler: I think `folder` is probably the wrong name. It is a just a name to indicate where a log message is coming from. This might not be a folder. 

`label`?

src/SwiftOutputChannel.ts
adam-fowler: I wanted to leave some flexibility here, don't necessarily want to always post `done`

src/SwiftOutputChannel.ts
svanimpe: OK

src/SwiftOutputChannel.ts
svanimpe: OK

src/FolderContext.ts
svanimpe: We already have two cycles here I think:

`FolderContext` <-> `PackageWatcher`
`WorkspaceContext` -> `FolderContext -> `PackageWatcher`-> `WorkspaceContext`

We should double-check how we store our state to make sure the mark-and-sweep GC can clean everything up.

adam-fowler: I'll add another PR to clean the cycle up

svanimpe: So this is good to go then?

svanimpe (approved):

PR #44, opened 2021-12-14
Add building of release/test builds

adam-fowler:
Add launch.json configs for building release executables and running tests

Did a quick check on Linux and it isn't work just at the moment

### DIFF
--- a/src/SwiftTaskProvider.ts
+++ b/src/SwiftTaskProvider.ts
@@ -68,13 +68,13 @@ function createCleanTask(): vscode.Task {
         createSwiftTask(
             'swift', 
             ['build', '--product', product.name], 
-            `Build ${product.name}`, 
+            `Build Debug ${product.name}`, 
             { group: vscode.TaskGroup.Build }
         ),
         createSwiftTask(
             'swift', 
             ['build', '-c', 'release', '--product', product.name], 
-            `Build ${product.name}`, 
+            `Build Release ${product.name}`, 
             { group: vscode.TaskGroup.Build }
         )
     ];
--- a/src/SwiftTaskProvider.ts
+++ b/src/SwiftTaskProvider.ts
@@ -63,13 +63,21 @@ function createCleanTask(): vscode.Task {
 /**
  * Creates a {@link vscode.Task Task} to run an executable target.
  */
- function createBuildTask(product: Product): vscode.Task {
-    return createSwiftTask(
-        'swift', 
-        ['build', '--product', product.name], 
-        `Build ${product.name}`, 
-        { group: vscode.TaskGroup.Build }
-    );
+ function createBuildTasks(product: Product): vscode.Task[] {
+    return [
+        createSwiftTask(
+            'swift', 
+            ['build', '--product', product.name], 
+            `Build ${product.name}`, 
+            { group: vscode.TaskGroup.Build }
+        ),
+        createSwiftTask(
+            'swift', 
+            ['build', '-c', 'release', '--product', product.name], 
+            `Build ${product.name}`, 
+            { group: vscode.TaskGroup.Build }
+        )
+    ];
 }
 
 /**
@@ -175,7 +183,7 @@ export class SwiftTaskProvider implements vscode.TaskProvider {
             if (!folder.isRootFolder) { continue; }
             const executables = folder.swiftPackage.executableProducts;
             for (const executable of executables) {
-                tasks.push(createBuildTask(executable));
+                tasks.push(...createBuildTasks(executable));
             }
         }
         return tasks;
--- a/src/SwiftTaskProvider.ts
+++ b/src/SwiftTaskProvider.ts
@@ -42,7 +42,7 @@ interface TaskConfig {
 function createBuildAllTask(): vscode.Task {
     return createSwiftTask(
         'swift', 
-        ['build'], 
+        ['build', '--build-tests'], 
         'Build All', 
         { group: vscode.TaskGroup.Build }
     );
--- a/src/SwiftTaskProvider.ts
+++ b/src/SwiftTaskProvider.ts
@@ -40,9 +40,10 @@ interface TaskConfig {
  * This excludes test targets.
  */
 function createBuildAllTask(): vscode.Task {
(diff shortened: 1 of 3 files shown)


src/debug.ts
```diff
@@ -43,20 +47,65 @@ export async function makeDebugConfigurations(ctx: FolderContext) {
...
+            }
+        ];    
     });
 }
+
+// Return array of DebugConfigurations for tests based on what is in Package.swift
+async function createTestConfigurations(ctx: FolderContext): Promise<vscode.DebugConfiguration[]> {
+    if (ctx.swiftPackage.getTargets('test').length === 0) { return []; }
+    try {
+        // If running on darwin. Find xctest exe and run pointing at xctest resources
```
0xTim, line 82: You _should_ be able to build and run on Darwin without Xcode by just installing the toolchain. It doesn't work with tests at the moment but we probably shouldn't lock ourselves into requiring Xcode

adam-fowler: Once we have custom path for swift I can look into that. The default at the moment is use Xcode. Im not sure how xctest stuff works with additional tool chains.

adam-fowler: @0xTim I've have a further investigation and I'm not sure you can run tests without Xcode being installed. The `xctest` executable doesn't come with the swift toolchain. It is only available from a Xcode installation.

src/SwiftTaskProvider.ts
```diff
@@ -40,9 +40,10 @@ interface TaskConfig {
  * This excludes test targets.
  */
 function createBuildAllTask(): vscode.Task {
+    const additionalArgs = (process.platform !== 'darwin') ? ['--enable-test-discovery'] : [];
```
0xTim, line 43: What version of Swift do we want to support? I'm in favour of 5.5 and up which means test discovery is on by default

adam-fowler: There is an issue with the automatic test discovery where if you have an empty `LinuxMain.swift` it doesn't work. But you can't remove the `LinuxMain.swift` if your library supports earlier versions of swift.

Also I don't see a reason for not supporting earlier versions of swift.

svanimpe (commented):
Unfortunately, testing doesn't work on Windows. I can run tests with `swift test --enable-test-discovery` but debugging via the launch config results in an "Unknown error".

I've checked that the **.xctest** file is present after compilation, but it's not an executable.

svanimpe:
It's also curious that on Windows, if I debug in release mode, I do see the breakpoints getting hit. This doesn't work when compiling in debug mode. The variable names are mangled though.

src/debug.ts
```diff
@@ -14,20 +14,29 @@
...
+    ];
     let edited = false;
     for (const config of configs) {
         const index = launchConfigs.findIndex(c => c.name === config.name);
         if (index !== -1) {
-            if (launchConfigs[index].program !== config.program) {
-                const answer = await vscode.window.showErrorMessage(`Launch configuration '${config.name}' already exists. Do you want to update it?`, 'Cancel', 'Update');
+            if (launchConfigs[index].program !== config.program || 
+                launchConfigs[index].cwd !== config.cwd ||
+                launchConfigs[index].preLaunchTask !== config.preLaunchTask) {
```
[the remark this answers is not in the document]
adam-fowler: Yes

src/debug.ts
```diff
@@ -43,20 +52,67 @@ export async function makeDebugConfigurations(ctx: FolderContext) {
...
+            return [{
+                type: "lldb",
+                request: "launch",
+                name: `Test ${ctx.swiftPackage.name}`,
+                program: `./.build/debug/${ctx.swiftPackage.name}PackageTests.xctest`,
+                cwd: `\${workspaceFolder:${ctx.folder.name}}`,
+                preLaunchTask: `swift: Build All`
+            }];
+        }
+    } catch {
```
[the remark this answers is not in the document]
adam-fowler: > I've checked that the **.xctest** file is present after compilation, but it's not an executable.

Great so Windows do it differently to Linux. I guess we'll have to find out how we can trigger tests correctly on Windows. Temporarily I'll disable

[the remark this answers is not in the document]
adam-fowler: getXcodePath can throw I believe

adam-fowler:
> I've checked that the **.xctest** file is present after compilation, but it's not an executable.

Great so windows do it differently to linux. I guess we'll have to find out how we can trigger tests correctly on Windows. Temporarily I'll disable.

src/debug.ts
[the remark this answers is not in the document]
svanimpe: It shouldn't throw I think, because you catch any errors and return `undefined` instead? If I understand JavaScript async/await correctly, that `undefined` should be a resolved promise, no?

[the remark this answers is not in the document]
adam-fowler: I haven't got the code in front of me at the moment. I'll look at it tomorrow. You're probably right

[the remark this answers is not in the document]
adam-fowler: ok temporarily disabled for win32 until we work out what to do, also removed the unnecessary try catch

src/debug.ts
0xTim: Ah I was hoping that had been fixed. I've opened this as an SR bug https://bugs.swift.org/browse/SR-15020 a while back so lets stick with what we have for now until it's been fixed

src/SwiftTaskProvider.ts
0xTim: Personally to avoid complexity and warnings I'd be in favour of ignoring `LinuxMain.swift` was ever a thing but happy to defer

0xTim (approved):

src/SwiftTaskProvider.ts
adam-fowler: @0xTim libraries that support older versions of swift (eg Soto) need so tests can run on those versions of swift

0xTim: Soto doesn't support less than 5.1 right? So test discovery will always be available to anyone running it. (This is all hypothetical at this point anyway as I'm happy to merge as the PR is)

adam-fowler: no it doesn't :-)

svanimpe (approved):

PR #51, opened 2021-12-15
Configure ESLint and Prettier

svanimpe:
This PR configures our ESLint rules and enable code formatting with Prettier.

Some details:
- Linting rules are reset to the defaults, with formatting rules disabled in favor of formatting with Prettier.
- Prettier is very opinionated (so we don't have to argue about formatting styles), so there's few configuration options.
- I've only enabled formatting for TypeScript and JSON files.
- The plan was to enable format-on-save. I will do this once you approve of these configuration settings.
- I've added a `format` script (`npm run format`) that can be used in CI to check formatting (without touching any code).
- I've added https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode as a recommended extension. Contributers will need to install this and the ESLint extension to enable linting and formatting.

@adam-fowler @0xTim Can you review the formatting rules in `prettierrc`? Once those get an OK, I will add additional commits to this PR to fix any existing linting issues, and to show the results of Prettier's formatting.


0xTim (approved):
Fine by me

svanimpe:
There's two ESLint rules in particular that we may have to consider disabling:

1. No explicit `any`: https://github.com/typescript-eslint/typescript-eslint/blob/v5.4.0/packages/eslint-plugin/docs/rules/no-explicit-any.md
2. No non-null assertions (force unwraps): https://github.com/typescript-eslint/typescript-eslint/blob/v5.4.0/packages/eslint-plugin/docs/rules/no-non-null-assertion.md

The first pops up when working with JSON. We could work around this by declaring an interface for the JSON format we expect to parse. This involves a bit more work, but is acceptable for me as it also documents the format we intend to parse.

The second I'm inclined to disable. We only use a force unwrap in cases where we just checked that a value is non-null, but where the type narrower is unable to strip `| undefined` or `| null` from the type based on the control flow. In particular, the type narrower doesn't understand that in the following example, `url` is provably non-null:

```typescript
return this.ctx.swiftPackage.dependencies.filter(
    dependency => !dependency.requirement && dependency.url 
).map(dependency => new PackageNode(
    dependency.identity,
    dependency.url!,
    'local',
    'local'
));
```

adam-fowler:
So I installed prettier in my vscode. I don't see it doing anything. 
Also I don't get warnings about things like `===` instead of `==` anymore.
When I run `npm run format` I get the error message `sh: prettier: command not found`

adam-fowler:
How do I actually see the results of running prettier.
Also can we increase the line limit. It seems to be 80 chars. Given the screens most people work on these days that seems really low. 120 as a starting point?

svanimpe:
The easiest way to test would be to flip formatOnSave to `true` in **.vscode/settings.json**:

```
    // Configure Prettier
    "editor.defaultFormatter": "esbenp.prettier-vscode",
    "[json]": {
        "editor.formatOnSave": false
    },
    "[typescript]": {
        "editor.formatOnSave": false
    }
```

The formatting configuration is in **.prettierrc.json**.

Prettier recommends a line width of 80, because it doesn't use it as a hard limit. I've already set it to 100, but feel free to up that to 120 if it's still not enough.

adam-fowler (approved):
Up the line length to 120 and I'm happy with this

PR #52, opened 2021-12-15
Weak reference

adam-fowler:
I've added this to avoid cyclical references.

This is an implementation of [`WeakRef`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef) which doesn't seem to be available yet.
It stores references in a dictionary, to avoid cyclical references. 
I have used it in one place so far `PackageWatcher`
I also fixed the Workspace cyclical reference, by removing `ExtensionContext` from it. We don't use it anywhere at the moment.

### DIFF
--- a/src/FolderContext.ts
+++ b/src/FolderContext.ts
@@ -17,6 +17,7 @@ import { PackageWatcher } from './PackageWatcher';
 import { SwiftPackage } from './SwiftPackage';
 import { WorkspaceContext } from './WorkspaceContext';
 import contextKeys from './contextKeys';
+import { WeakReference } from './utilities/WeakReference';
 
 export class FolderContext implements vscode.Disposable {
     private packageWatcher?: PackageWatcher;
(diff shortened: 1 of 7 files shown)


src/FolderContext.ts
```diff
@@ -29,7 +30,6 @@ export class FolderContext implements vscode.Disposable {
     ) {
         if (this.isRootFolder) {
             this.packageWatcher = new PackageWatcher(this, workspaceContext);
-            this.packageWatcher.install();
```
0xTim, line 32: What does this have to do with weak references?

adam-fowler: The install and constructor have been combined in this PR. I guess it isn't completely related

0xTim (approved):

adam-fowler:
Most modern javascript engines garbage collection can deal with cyclical references so this shouldn't be an issue.

PR #57, opened 2021-12-16
Setup CoreLLDB Configuration automatically

adam-fowler:
Attempts to find the correct version of LLDB for debugging Swift and then suggests to the user they allow it to update their CoreLLDB settings. The code is the same as is used in CoreLLDB's use alternate backend command.

Testing to do:

- [x] test with standard macOS swift install
- [x] test with swift path set on macOS
- [x] test on Linux
- [x] test on Windows

### DIFF
--- a/src/WorkspaceContext.ts
+++ b/src/WorkspaceContext.ts
@@ -121,7 +121,7 @@ export class WorkspaceContext implements vscode.Disposable {
 
         // show dialog for setting up LLDB
         vscode.window.showInformationMessage(
-            "CoreLLDB requires the correct Swift version of LLDB for debugging. Do you want to set this up in your global settings or the workspace Settings?", 
+            "CoreLLDB requires the correct Swift version of LLDB for debugging. Do you want to set this up in your global settings or the workspace settings?", 
             'Cancel', 'Global', 'Workspace').then(result => {
                 switch (result) {
                 case 'Global':
--- a/src/WorkspaceContext.ts
+++ b/src/WorkspaceContext.ts
@@ -16,7 +16,8 @@ import * as vscode from 'vscode';
 import { FolderContext } from './FolderContext';
 import { StatusItem } from './StatusItem';
 import { SwiftOutputChannel } from './SwiftOutputChannel';
-import { execSwift, getXCTestPath } from './utilities';
+import { execSwift, getSwiftExecutable, getXCTestPath } from './utilities';
+import { getLLDBLibPath } from './lldb';
 
 // Context for whole workspace. Holds array of contexts for each workspace folder
 // and the ExtensionContext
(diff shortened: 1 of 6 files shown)


svanimpe:
I've tried this out on Windows. It seems to detect LLDB correctly, but setting it triggers this bug: https://github.com/vadimcn/vscode-lldb/issues/566
Therefore, we may need to skip this on Windows. If we don't set the LLDB library, we can at least run executables.

On macOS, I noticed that if the LLDB library is set to **/Applications/Xcode.app/Contents/SharedFrameworks/LLDB.framework/Versions/A/LLDB** (which is what CoreLLDB autodetects), then I still get the prompt asking me to set (and thus overwrite) it. The new value is then **/Applications/Xcode.app/Contents/Developer/usr/bin/lldb**, which also works fine.

adam-fowler:
> On macOS, I noticed that if the LLDB library is set to **/Applications/Xcode.app/Contents/SharedFrameworks/LLDB.framework/Versions/A/LLDB** (which is what CoreLLDB autodetects)

When does CoreLLDB autodetect this? The version of lldb it used when I installed was the one packaged with the extension.

adam-fowler:
> I've tried this out on Windows. It seems to detect LLDB correctly, but setting it triggers this bug: [vadimcn/vscode-lldb#566](https://github.com/vadimcn/vscode-lldb/issues/566) Therefore, we may need to skip this on Windows. If we don't set the LLDB library, we can at least run executables.

I'll temporarily disable the windows version

svanimpe:
> When does CoreLLDB autodetect this? 

If I run the "Use Alternate Backend" command, as suggested here:

https://github.com/vadimcn/vscode-lldb/wiki/Swift

adam-fowler:
> > When does CoreLLDB autodetect this?
> 
> If I run the "Use Alternate Backend" command, as suggested here:
> 
> https://github.com/vadimcn/vscode-lldb/wiki/Swift

I've replicated the LLDB code here, so now both should return same values.

src/WorkspaceContext.ts
```diff
@@ -95,6 +85,45 @@ export class WorkspaceContext implements vscode.Disposable {
...
+        const libPath = await getLLDBLibPath(getSwiftExecutable('lldb'));
+        if (!libPath) { return; }
+
+        const lldbConfig = vscode.workspace.getConfiguration('lldb');
+        const configLLDBPath = lldbConfig.get<string>('library');
+        if (configLLDBPath === libPath) { return; }
+
+        // show dialog for setting up LLDB
+        vscode.window.showInformationMessage(
+            "CoreLLDB requires the correct Swift version of LLDB for debugging. Do you want to set this up in your global settings or the workspace Settings?",
```
0xTim, line 112: Nit casing
```suggestion
            "CoreLLDB requires the correct Swift version of LLDB for debugging. Do you want to set this up in your global settings or the workspace settings?", 
```

0xTim (changes requested):
One nit

0xTim (approved):

src/WorkspaceContext.ts
```diff
@@ -95,6 +85,45 @@ export class WorkspaceContext implements vscode.Disposable {
...
+    }
+
+    // find LLDB version and setup path in CoreLLDB
+    async setLLDBVersion() {
+        // don't set LLDB on windows as swift version is not working at the moment
+        if (process.platform === 'win32') { return; }
+        const libPath = await getLLDBLibPath(getSwiftExecutable('lldb'));
+        if (!libPath) { return; }
+
+        const lldbConfig = vscode.workspace.getConfiguration('lldb');
```
svanimpe, line 106: We could also use our `configuration` object here.
Although the property model is a bit limited, so we'd need to add a `setLLDBLibrary` method instead of a property setter, so that you can also pass in the configuration target.
Which do you prefer?

svanimpe (approved):

PR #55, opened 2021-12-15
Add new entries to .vscodeignore

adam-fowler:

PR #61, opened 2021-12-18
Bundle using esbuild

### LINKED ISSUE #50: Clean up folder and bundle extension
This is the output when we run `rpm run package`.

`
This extension consists of 227 files, out of which 169 are JavaScript files. For performance reasons, you should bundle your extension: https://aka.ms/vscode-bundle-extension . You should also exclude unnecessary files by adding them to your .vscodeignore: https://aka.ms/vscode-vscodeignore
`

We should be able to clean up some of those files and we should look into https://aka.ms/vscode-bundle-extension

svanimpe:
This uses `esbuild` to bundle the extension into a single file. Some notable changes:

* The output of `esbuild` goes into the **dist** folder. Only that file needs to be included in the extension. We can ignore the regular compilation output that goes into **out**.
* Make sure to install the recommended extensions (listed in **.vscode/extensions.json**) as the new one includes a problem matcher for `esbuild`.
* **package.json** still includes the old compilation scripts. Some of them are still used for testing, which relies on regular compilation, not bundling.

With this, the extension now includes only 11 files. The output of `npm run package` says 13 files, but I think that includes directories. In any case, you can check which files are included with `npx vsce ls`.

This fixes #50 and supersedes #55

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "onLanguage:swift",
     "workspaceContains:Package.swift"
   ],
-  "main": "./out/extension.js",
+  "main": "./dist/extension.js",
   "contributes": {
     "commands": [
       {
@@ -144,12 +144,15 @@
     "vadimcn.vscode-lldb"
   ],
   "scripts": {
-    "vscode:prepublish": "npm run compile",
-    "compile": "tsc -p ./",
-    "watch": "tsc -watch -p ./",
-    "pretest": "npm run compile && npm run lint",
-    "lint": "eslint src --ext ts",
+    "vscode:prepublish": "npm run esbuild-base -- --minify",
+    "esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node --target=node14",
+    "esbuild": "npm run esbuild-base -- --sourcemap",
+    "esbuild-watch": "npm run esbuild-base -- --sourcemap --watch",
+    "compile": "tsc",
+    "watch": "tsc --watch",
+    "lint": "eslint src --ext ts && tsc --noEmit",
     "format": "prettier --check src **/*.json",
+    "pretest": "npm run compile && npm run lint",
     "test": "node ./out/test/runTest.js",
     "package": "vsce package"
   },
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -144,12 +144,15 @@
...
-    "vscode:prepublish": "npm run compile",
-    "compile": "tsc -p ./",
-    "watch": "tsc -watch -p ./",
-    "pretest": "npm run compile && npm run lint",
-    "lint": "eslint src --ext ts",
+    "vscode:prepublish": "npm run esbuild-base -- --minify",
+    "esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node --target=node14",
+    "esbuild": "npm run esbuild-base -- --sourcemap",
+    "esbuild-watch": "npm run esbuild-base -- --sourcemap --watch",
+    "compile": "tsc",
```
adam-fowler, line 151: how come the `-p ./` has been removed?

package.json
```diff
@@ -144,12 +144,15 @@
...
+    "watch": "tsc --watch",
+    "lint": "eslint src --ext ts && tsc --noEmit",
```
adam-fowler, line 153: What does the `tsc --noEmit` do here?

adam-fowler (commented):
This looks fine. Not totally sure how this works so will have to trust you on it. Just a couple of questions

adam-fowler:
Nice to see the package half in size though

package.json
svanimpe: I took a closer look at the commands as I was updating them, and it turns out `-p ./` is just the default (meaning **tsconfig.json** is in the current/root directory. I also noticed the watch option was missing the extra dash.

package.json
svanimpe: `esbuild` doesn't do type checking, so the recommendation was to run `tsc --noEmit` (meaning: don't emit compiler output, just type check) at the linting stage.

During development, VS Code should still flag any errors.

adam-fowler (approved):

0xTim (approved):
LGTM

Issue #65, opened 2021-12-21
Package dependency view is not updating when Package.swift updates


adam-fowler:
The resolve is run but when the task finishes `PackageDependenciesProvider` does not recognise the task. It seems as if vscode has messed with the definition when you run it. The type gets changed and the definition arguments are undefined

adam-fowler:
#66 would fix this

svanimpe:
That's really weird. I did some digging around and found out why this happens:

In **commands.ts**, if we modify the fetched task:

```typescript
task.presentationOptions = {
  reveal: vscode.TaskRevealKind.Silent
};
```

it seems like VS Code creates a new task, with a new definition. If I instead set the `presentationOptions` when creating a resolve task in the provider, the definition is unmodified. This is probably also why in `SwiftTaskProvider.resolveTask` I had to create a new `Task` object, and was unable to just "fill in" the tasked passed in as a parameter.

So yeah, seems like we really need to use a task's name to identify it, not its ID or definition.

PR #66, opened 2021-12-21
Use Task name to identify tasks

adam-fowler:
Using task definition `command` and `args` seems to produce inconsistent results. See #65. Instead if you use `Task.name` you get the correct result every time.

I have added static variables for all the standard task names to ensure consistency across the project.


svanimpe (approved):
Looks good.

Seeing as task names are now used as identifiers, we should also use them in launch configurations (`preLaunchTask`), to avoid typo's.

0xTim (approved):

PR #68, opened 2021-12-21
Add build tasks for all folders in a workspace

adam-fowler:
When we add a folder to a multi-folder workspace it automatically brings in the launch.json configs for that folder. This PR generates the related tasks so those configurations still work


0xTim (approved):

adam-fowler:
Realised today this doesn't include build all tasks for each folder which the tests will need. Going to leave this one until post v0.1.0

PR #72, opened 2021-12-22
Add debug configuration for tests on Windows.

### LINKED ISSUE #63: Work out what executable to run to debug tests on Windows
When creating a launch configuration to debug tests we need the actual executable that is run when the tests are running (we don't want to attach to the swift executable).

- On macOS we run `xctest mytests.xctest`.
- On linux we run `my tests.xctest`.

I assumed that Windows would be the same as Linux, but @svanimpe has pointed out this is not the case.

@compnerd can you help us? What is the executable we should attach to on Windows?

svanimpe:
This fixes #63 

It doesn't support cross-compilation, but it does use **DEVELOPER_DIR** and **Info.plist** to find the path to **XCTest.dll**.

### DIFF
--- a/src/debug.ts
+++ b/src/debug.ts
@@ -14,7 +14,7 @@
 
 import * as vscode from 'vscode';
 import { FolderContext } from './FolderContext';
-import { getXCTestDLLPath, getXcodePath } from './utilities';
+import { getXcodePath } from './utilities';
 
 /** 
  * Edit launch.json based on contents of Swift Package.
@@ -89,7 +89,7 @@ async function createTestConfigurations(ctx: FolderContext): Promise<vscode.Debu
     if (ctx.swiftPackage.getTargets('test').length === 0) { return []; }
 
     if (process.platform === 'darwin') {
-        // On Apple platforms, find the path to xctest
+        // On macOS, find the path to xctest
         // and point it at the .xctest bundle from the .build directory.
         const xcodePath = await getXcodePath();
         if (xcodePath === undefined) {
@@ -107,14 +107,7 @@ async function createTestConfigurations(ctx: FolderContext): Promise<vscode.Debu
     } else if (process.platform === 'win32') {
         // On Windows, add XCTest.dll to the PATH,
         // and then run the .xctest bundle from the .build directory.
-        let xctestPath: string;
-        try {
-            xctestPath = await getXCTestDLLPath();
-        } catch (error) {
-            vscode.window.showErrorMessage(
-                `I was unable to create a debug configuration for testing ` + 
-                `because I cannot find XCTest.dll. ${error}` 
-            );
+        if (!ctx.workspaceContext.xcTestPath) {
             return [];
         }
         return [{
@@ -124,7 +117,7 @@ async function createTestConfigurations(ctx: FolderContext): Promise<vscode.Debu
             program: `./.build/debug/${ctx.swiftPackage.name}PackageTests.xctest`,
             cwd: `\${workspaceFolder:${ctx.folder.name}}`,
             env: {
-                'path': `${xctestPath};\${env:PATH}`
+                'path': `${ctx.workspaceContext.xcTestPath};\${env:PATH}`
             },
             preLaunchTask: `swift: Build All`
         }]; 
@@ -139,6 +132,4 @@ async function createTestConfigurations(ctx: FolderContext): Promise<vscode.Debu
             preLaunchTask: `swift: Build All`
         }]; 
     }
-
-    return [];
 }
--- a/src/debug.ts
+++ b/src/debug.ts
@@ -14,7 +14,7 @@
 
 import * as vscode from 'vscode';
 import { FolderContext } from './FolderContext';
-import { getXcodePath } from './utilities';
+import { getXCTestDLLPath, getXcodePath } from './utilities';
 
 /** 
  * Edit launch.json based on contents of Swift Package.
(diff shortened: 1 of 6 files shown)


adam-fowler (commented):
Generally looks good. Couple of issues/questions

src/debug.ts
```diff
@@ -106,16 +104,40 @@ async function createTestConfigurations(ctx: FolderContext): Promise<vscode.Debu
...
             preLaunchTask: `swift: Build All`
         }];
+    } else if (process.platform === 'win32') {
+        // On Windows, add XCTest.dll to the PATH,
+        // and then run the .xctest bundle from the .build directory.
+        let xctestPath: string;
+        try {
+            xctestPath = await getXCTestDLLPath();
+        } catch (error) {
+            vscode.window.showErrorMessage(
```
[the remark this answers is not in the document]
svanimpe: I could store it in the workspace context, since it's a global setting?
But the `createTestConfigurations` function only has access to the folder context, not the workspace context. I'd really like to have a back-reference to the workspace context in each folder context.  I don't even think we need to worry about reference cycles here, due to how the garbage collector works, but I'd need to double-check that.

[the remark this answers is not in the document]
adam-fowler: Looks like cyclical references aren't a concern anymore with modern JS implementations so maybe it isn't an issue in the end. Add a `WorkspaceContext` reference into the `FolderContext`. It will probably make our life easier.

src/utilities.ts
```diff
@@ -122,4 +123,34 @@ export async function getXcodePath(): Promise<string|undefined> {
...
     }
-}
\ No newline at end of file
+}
+
+/**
+ * Contents of **Info.plist** on Windows.
+ */
+interface InfoPlist {
+
```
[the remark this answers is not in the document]
svanimpe: I guess that's a matter of style? I've always preferred having a blank line (paragraph break) after a type declaration header, so it stands out more. In this case, it doesn't make much of a difference.

I've also noticed that Prettier removes them, which is quite controversial (https://github.com/prettier/prettier/issues/4870) 😕 So if we keep Prettier, they will all get removed.

src/debug.ts
```diff
@@ -106,16 +104,40 @@ async function createTestConfigurations(ctx: FolderContext): Promise<vscode.Debu
...
+                `I was unable to create a debug configuration for testing ` + 
+                `because I cannot find XCTest.dll. ${error}`
```
0xTim, line 116: This sounds a bit sentient 😅 
```suggestion
                `Unable to create a debug configuration for testing ` + 
                `because XCTest.dll cannot be found. ${error}` 
```

0xTim (approved):
LGTM other than @adam-fowler 's comments. One suggestion around error message wording

src/debug.ts
```diff
@@ -87,12 +87,10 @@ function createExecutableConfigurations(ctx: FolderContext): vscode.DebugConfigu
 // Return array of DebugConfigurations for tests based on what is in Package.swift
 async function createTestConfigurations(ctx: FolderContext): Promise<vscode.DebugConfiguration[]> {
     if (ctx.swiftPackage.getTargets('test').length === 0) { return []; }
-    // if platform is windows temporarily disable test launch config generation until we work
-    // out how to run the test executable on windows
-    if (process.platform === "win32") { return []; }
 
-    // If running on darwin. Find xctest exe and run pointing at xctest resources
     if (process.platform === 'darwin') {
+        // On Apple platforms, find the path to xctest
```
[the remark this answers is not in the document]
svanimpe: Done

src/debug.ts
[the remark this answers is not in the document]
svanimpe: Done

src/debug.ts
svanimpe: Done.

I figured it may sound friendlier in the first person 😄 Passive voice can be harder to read for non-native speakers.

src/utilities.ts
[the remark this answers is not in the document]
svanimpe: Done

svanimpe (commented):
I've moved the **XCTest.dll** loading to the first time we add a workspace folder. This means we'll still repeat the error when we add a second folder (because it starts a new session). However, we'll never show it twice at the same time.

I've also considered moving this code to **extension.ts** as an observer, but since we store the `xcTestPath` in the `WorkspaceContext`, I found it best to manage it there.

src/debug.ts
```diff
@@ -106,16 +104,40 @@ async function createTestConfigurations(ctx: FolderContext): Promise<vscode.Debu
...
+            xctestPath = await getXCTestDLLPath();
+        } catch (error) {
+            vscode.window.showErrorMessage(
+                `I was unable to create a debug configuration for testing ` + 
+                `because I cannot find XCTest.dll. ${error}` 
+            );
+            return [];
+        }
+        return [{
+            type: "lldb",
```
compnerd, line 114: Are we always generating dwarf debug information?  Are you also passing `-Xlinker -debug:dwarf` when building?  Without that, this is not particularly useful.

compnerd:
Does it make sense to split out the changes for adding the new dependencies into a separate commit so that the change itself is easier to digest?

src/debug.ts
adam-fowler: @svanimpe could this be why your windows debug sessions aren't working

svanimpe: Yes, that fixes part of it. If I add those build flags, I can set breakpoints. However, I still can't resolve any variables (https://bugs.swift.org/browse/SR-14243)

adam-fowler: Is that with CoreLLDB setup for swift?

adam-fowler (approved):
Apart from changing the error message to indicate the swift installation may be corrupt this is good to go. 

Maybe in a second PR you can add the dwarf debug linker flag to the build tasks

src/debug.ts
svanimpe: Yes, I get that error when I set `liblldb.dll`. I also get the same error when running the `LLDB: Use Alternate Backend` command.

With an empty `lldb.library` setting, and these build flags, I can set breakpoints, but not inspect any variables.

src/debug.ts
svanimpe: Done

PR #73, opened 2021-12-22
Add support for loading test fixtures

svanimpe:
This is just some code I wrote to load fixtures from the root **test** directory. I ended up not needing it after all, but it should be useful for future testing.

### DIFF
--- a/src/test/fixtures.ts
+++ b/src/test/fixtures.ts
@@ -1,0 +1,33 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VSCode Swift open source project
+//
+// Copyright (c) 2021 the VSCode Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of VSCode Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import * as vscode from 'vscode';
+import * as fs from 'fs/promises';
+import * as path from 'path';
+
+/**
+ * Returns the {@link vscode.Uri URI} of a resource in the **test** directory.
+ */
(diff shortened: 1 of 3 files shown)


adam-fowler:
I was thinking of having an `assets/test/` folder outside of the `src/` folder.
I have half a PR doing package loading tests using assets/test/ 
I guess we could use a function to get fullpaths for folders, but it would be better to be something shorter `testAssetURI`

svanimpe:
I think the default is a root **test** directory (not to be confused with the **test** directory inside **src**). VS Code actually generated that directory, but it got deleted as part of https://github.com/swift-server/vscode-swift/commit/6376cfa67e5ab8ebf795ca3470ce4eddf3c37382

adam-fowler:
> I think the default is a root **test** directory (not to be confused with the **test** directory inside **src**). VS Code actually generated that directory, but it got deleted as part of [6376cfa](https://github.com/swift-server/vscode-swift/commit/6376cfa67e5ab8ebf795ca3470ce4eddf3c37382)

As long they aren't mixed in the src code, I'm happy

adam-fowler:
This would be useful as well
```swift
class TestWorkspaceFolder implements vscode.WorkspaceFolder {
    constructor(
        readonly uri: vscode.Uri
    ) {}
    get name(): string { return path.basename(this.uri.fsPath); }
    get index(): integer { return 0; }
}
/**
  * @returns the {@link vscode.WorkspaceFolder workspace folder} of a resource folder in the **test** directory.
  */
 export function getTestResourceWorkspaceFolder(name: string): vscode.WorkspaceFolder {
    return new TestWorkspaceFolder(getTestResourceUri(name));
}
```
Although I be keener on shorter function names still.

0xTim (approved):

adam-fowler:
I've added versions of these in the package tests PR. We can probably close this without merging.

PR #75, opened 2021-12-23
Add build arguments for debugging on Windows

svanimpe:
This just adds some build arguments for debugging on Windows as @compnerd suggested.

### DIFF
--- a/src/SwiftTaskProvider.ts
+++ b/src/SwiftTaskProvider.ts
@@ -39,10 +39,15 @@ interface TaskConfig {
 
 /**
  * Creates a {@link vscode.Task Task} to build all targets in this package.
- * This excludes test targets.
  */
 function createBuildAllTask(): vscode.Task {
-    const additionalArgs = (process.platform !== 'darwin') ? ['--enable-test-discovery'] : [];
+    const additionalArgs: string[] = [];
+    if (process.platform !== 'darwin') {
+        additionalArgs.push('--enable-test-discovery');
+    }
+    if (process.platform === 'win32') {
+        additionalArgs.push('-Xlinker', '-debug:dwarf');
+    }
     return createSwiftTask(
         ['build', '--build-tests', ...additionalArgs, ...configuration.buildArguments], 
         SwiftTaskProvider.buildAllName, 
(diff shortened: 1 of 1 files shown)


adam-fowler (approved):
LGTM

src/SwiftTaskProvider.ts
```diff
@@ -39,10 +39,15 @@ interface TaskConfig {
...
  */
 function createBuildAllTask(): vscode.Task {
-    const additionalArgs = (process.platform !== 'darwin') ? ['--enable-test-discovery'] : [];
+    const additionalArgs: string[] = [];
+    if (process.platform !== 'darwin') {
+        additionalArgs.push('--enable-test-discovery');
+    }
+    if (process.platform === 'win32') {
+        additionalArgs.push('-Xlinker', '-debug:dwarf');
+    }
```
compnerd, line 50: Please expose this as an option.  This guarantees a ~90% chance of completely disabling debugging abilities (if you have a mixed codebase).  Codeview doesn't let you always debug Swift code, but for mixed codebases, it is preferred, and when it works, is more stable than DWARF.  It does however preclude the use of lldb.

src/SwiftTaskProvider.ts
```diff
@@ -65,9 +70,10 @@ function createCleanTask(): vscode.Task {
  * Creates a {@link vscode.Task Task} to run an executable target.
  */
  function createBuildTasks(product: Product): vscode.Task[] {
+    const debugArguments = process.platform === 'win32' ? ['-Xlinker', '-debug:dwarf'] : [];
```
compnerd, line 73: If you are using codeview, this should be `-debug` rather than `-debug:dwarf`.  Also, you need to ensure that `-use-ld=link` is not passed in when using this argument as this is only valid with lld.

compnerd: (Oh, alternatively, and even better - ensure that `-use-ld=lld` is present for `-debug:dwarf`)

svanimpe: I had to look up a few things here, so just to make sure I understand this correctly:

- Is CodeView the default on Windows (when just using `swift build` for example)?
- Should we always add `-debug` when creating a debug build using CodeView?
- What would be an appropriate name for this configuration setting? "Debug using DWARF"?

What should be the default, if we want to make sure things "just work" for new developers?

compnerd: > I had to look up a few things here, so just to make sure I understand this correctly:
> 
> * Is CodeView the default on Windows (when just using `swift build` for example)?

No, the default is no debug info :). `-g -debug-info-format=codeview` enables codeview, `-g` enables DWARF.

> * Should we always add `-debug` when creating a debug build using CodeView?

Yes, `-debug` should be specified always when building with CodeView debug info.

> * What would be an appropriate name for this configuration setting? "Debug using DWARF"?

That is what I would call it.  "Debug (DWARF)" and "Debug (CodeView)" is what I use.  But the using rather than the parenthetical is better for display purposes perhaps.

> What should be the default, if we want to make sure things "just work" for new developers?

My preference would be CodeView.  There are some bugs that we do need to sort out, but the same can be said for DWARF (and arguably more).  However, if you do use CodeView, you can deal with mixed-language projects (C/C++/Swift) and debug across system code and Swift code.  Additionally, I've found that lldb is still not particularly stable and causes trouble, and using DWARF mandates the use of lldb (at this point).

PR #84, opened 2021-12-24
Add `focus`, `unfocus` events for workspace folders

adam-fowler:
`focus` and `unfocus` events are sent to observers when a file is opened based on which workspace folder the file is in.
If when opening a file it is from a different workspace folder than the previous active file then an `unfocus` event is sent for the previous active workspace folder and a `focus` event is sent for the new workspace folder.

This PR assumes the prettier PR #79 has been merged

### DIFF
--- a/src/WorkspaceContext.ts
+++ b/src/WorkspaceContext.ts
@@ -81,7 +81,7 @@ export class WorkspaceContext implements vscode.Disposable {
      * set the focus folder
      * @param folder folder that has gained focus
      */
-    async focusFolder(folder: vscode.WorkspaceFolder) {
+    async focusFolder(folder?: vscode.WorkspaceFolder) {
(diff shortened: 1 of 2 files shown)


src/WorkspaceContext.ts
```diff
@@ -65,12 +133,21 @@ export class WorkspaceContext implements vscode.Disposable {
                 );
             }
         }
-        for (const observer of this.observers) {
-            await observer(folderContext, "add");
+        await this.fireEvent(folderContext, "add");
+
+        // is this the first folder then set a focus event
```
0xTim, line 138: Nit 
```suggestion
        // if this is the first folder then set a focus event
```

0xTim (approved):
One minor typo but LGTM

PR #92, opened 2021-12-27
Fix LLDB config setup

adam-fowler:
I missed a section of the LLDB config setup. This is causing it to fail setting up correctly when using remote containers.


0xTim (approved):
LGTM

PR #111, opened 2022-01-13
Add reset/clean commands

adam-fowler:
Also add package reset and resolve to dependency view

question: should reset run a resolve immediately after as it completely clears out the `.build` folder


0xTim (approved):
Re rerunning resolve after a reset I'd vote for yes. Pretty sure this is how Xcode works

adam-fowler:
> Re rerunning resolve after a reset I'd vote for yes. Pretty sure this is how Xcode works

done

PR #119, opened 2022-01-19
Edit package menu

adam-fowler:
Added commands
- useLocalDependency
- editDependency
- uneditDependency
- openInWorkspace

Added right click menu to dependency view which use above commands

- Added menu items for remote packages
  - Edit Locally: This will create local version of module that can be edited and add it to the workspace.
  - Use local version: This will bring up a file dialog so you can choose a local version of the dependency
- Added menu item for local or edited dependency
  - Open in Workspace: This will add the dependency folder to the workspace
- Added menu item for edited dependency
  - Revert to Original Version: This will revert to the version of the dependency detailed in Package.swift

An edited dependency will have the pencil icon next to it, instead of the archive icon

The extension works out what are edited packages are by looking at the contents of the Packages folder.

### DIFF
--- a/src/ui/PackageDependencyProvider.ts
+++ b/src/ui/PackageDependencyProvider.ts
@@ -143,13 +143,24 @@ export class PackageDependenciesProvider implements vscode.TreeDataProvider<Tree
             return [];
         }
         if (!element) {
-            // Build PackageNodes for all dependencies.
+            // Build PackageNodes for all dependencies. Because Package.resolved might not
+            // be up to date with edited dependency list, we need to remove the edited
+            // dependencies from the list before adding in the edit version
             const children = [
                 ...this.getLocalDependencies(folderContext),
                 ...this.getRemoteDependencies(folderContext),
-                ...(await this.getEditedDependencies(folderContext)),
-            ].sort((first, second) => first.name.localeCompare(second.name));
-            return children;
+            ];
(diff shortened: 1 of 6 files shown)


adam-fowler:
At the moment it's a bit flakey sometimes. Package resolves fail then all the package stuff disappears and you need to run a resolve externally from the editor.

src/ui/PackageDependencyProvider.ts
```diff
@@ -193,6 +210,35 @@ export class PackageDependenciesProvider implements vscode.TreeDataProvider<Tree
...
+                        if (item.isSymbolicLink()) {
+                            folder = await fs.readlink(folder, "utf8");
+                        }
+                        return new PackageNode(item.name, folder, "local", "editing");
+                    })
+            );
+        } catch {
+            // ignore errors. They basically mean there was no Packages folder
+        }
+        return [];*/
```
0xTim, line 239: We should probably remove this if we're not using it

0xTim (approved):
LGTM, just one tidy up to do

src/ui/PackageDependencyProvider.ts
adam-fowler: done

adam-fowler:
@0xTim I removed the `edit locally` option as the `swift package edit` implementation is really flakey. If it fails it doesn't clean up after itself and then future attempts always fail.

PR #122, opened 2022-01-22
packaging: bump to current node.js

compnerd:
Node 14 is in maintenance LTS releases, the active releases are on the Node 16 release path.  Node seems to indicate that Node 16 should be preferred.  This updates the target to use node 16 rather than node 14.  This behavioural chance also now matches the behaviour of the VSCode extension when it resided in SourceKit-LSP.

### DIFF
--- a/package.json
+++ b/package.json
@@ -199,7 +199,7 @@
   ],
   "scripts": {
     "vscode:prepublish": "npm run esbuild-base -- --minify",
-    "esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node --target=node14",
+    "esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node",
     "esbuild": "npm run esbuild-base -- --sourcemap",
     "esbuild-watch": "npm run esbuild-base -- --sourcemap --watch",
     "compile": "tsc",


svanimpe:
I think I added that flag because VS Code generated the extension to target 14 by default. If we bump this, we should probably also bump the version in **package.json** and verify everything still works.

package.json
```diff
@@ -199,7 +199,7 @@
   ],
   "scripts": {
     "vscode:prepublish": "npm run esbuild-base -- --minify",
-    "esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node --target=node14",
+    "esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node",
```
adam-fowler, line 202: Should this be changed to `--target=node16` instead of just removing it?

compnerd: We could keep it, but would need to continually monitor it.  I believe that the default is the current active branch, which seems like the better approach to me.

adam-fowler: Without the `--target` are we not building with whatever is setup on the computer. We could see changes based on who builds the package. 

I'm not overly experienced here so don't know if this is a real issue.

compnerd: Ah, interesting.  Okay, then I suppose it makes sense to use `--target=node16`.  However, it does mean that we would need to be a bit more vigilant to keep it up-to-date.

PR #125, opened 2022-01-26
Add Version Object for comparing Swift versions

adam-fowler:
As features appear in swift we will want to take advantage of them but to do that we need to know what version of swift we are running. This PR adds a version object and extracts the version from the output of `swift --version`.

I've also added tests to ensure the version comparisons are working.

The first use of this will be checking whether we need to restart the LSP server when we add a new file. Currently we need to do this, but when swift 5.6  is released it should no longer be necessary.

### DIFF
--- a/src/WorkspaceContext.ts
+++ b/src/WorkspaceContext.ts
@@ -201,7 +201,7 @@ export class WorkspaceContext implements vscode.Disposable {
             const version = stdout.trimEnd();
             this.outputChannel.log(version);
             // extract version
-            const match = version.match(/Apple Swift version ([\S]+)/);
+            const match = version.match(/Swift version ([\S]+)/);
             if (match) {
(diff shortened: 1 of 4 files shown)


src/WorkspaceContext.ts
```diff
@@ -45,6 +50,13 @@ export class WorkspaceContext implements vscode.Disposable {
         this.statusItem.dispose();
     }
 
+    /** Get swift version and create WorkspaceContext */
+    static async create(extensionContext: vscode.ExtensionContext): Promise<WorkspaceContext> {
+        // get swift version and then create
```
0xTim, line 55: This seems a bit redundant

0xTim (approved):
LGTM

src/WorkspaceContext.ts
adam-fowler: Unfortunately in Typescript you can't have async constructors, so I need to create an async static function

PR #130, opened 2022-01-30
Restart LSP Server on new file is running Swift 5.5 or lower

adam-fowler:
Swift 5.6 adds support for the LSP server `workspace/didChangeWatchedFiles` message. Until we have this though we need to restart the LSP every time a new file is created/deleted if we want symbol completion to work. 

Had to add a bunch of promises to ensure we aren't restarting the LSP server in the middle of restarting it already.


0xTim (approved):
LGTM

adam-fowler:
Verified with Alex Hoppen from sourcekit-lsp team they are happy with this

PR #133, opened 2022-02-03
Allow events to be triggered with a null folder

### LINKED ISSUE #127: not work on my mac
This plugin is not work on my mac, e.g: `Code completion`, but the `Swift Development Environment` can work normally. I'm confused, why is this happening...

adam-fowler:
This is mainly being implemented so we can start the LSP server without a workspace folder. LSP server requires a focus event to startup, if you startup with no folders then a focus event is sent with a null folder.

This should fix #127


0xTim (approved):
LGTM

Issue #131, opened 2022-02-01
Allow specifying subdirectory for Swift package


aslilac:
My use case involves having a Swift package in a repo that contains a lot of code in other languages as well, so instead of my Package.swift file being located at the root where the extension expects, it's located somewhere more like packages/swift-package/Package.swift.

It'd be nice for the extension to have an option for this. Something like `"swift.workingDirectories": ["packages/swift-package"]` would work well, and allow for specifying multiple Swift packages to instantiate language servers for. Not sure if the extension is set up to handle multiple packages like that very easily, but this is what the ESLint extension does, for example.

0xTim:
We support multiple directories containing different packages as of 0.2.0, are you experiencing this issue with that version? @adam-fowler this should work right?

adam-fowler:
We don't support the `Package.swift` in a sub-directory at the moment. What we support is workspaces with multiple folders. If you add your swift package folder to the workspace using the menu item `File -> Add Folder to Workspace` then it'll work.

Out of interest are there other language extensions that allow for a manifest file in a sub-folder

adam-fowler:
@aslilac This could get complex when multiple Package.swift files are found.

aslilac:
Not sure how you'd find multiple? You only step up until you see the closest one, and then stop (which is exactly what TS IntelliSense and rust-analyzer do)

adam-fowler:
That'll require some changes but should be doable

adam-fowler:
@aslilac I have a PR #137 for supporting loading of swift projects in sub-folders. It would be great if you could test it out. If you could build a version of the extension from the associated branch and run it to check if it works for you. There are details on building the extension to be found [here](https://github.com/swift-server/vscode-swift/blob/main/CONTRIBUTING.md)

aslilac:
Awesome! I'll check it out tonight!

adam-fowler:
> Awesome! I'll check it out tonight!

Did you manage to check it out? The changes are now on main with some additional fixes that were required.

PR #134, opened 2022-02-04
Replace WorkspaceFolder in FolderContext with a Uri

adam-fowler:
This is part of the process of getting #131 working. Previously FolderContext used WorkspaceFolder to define where the root of the swift package. To support #131 we need to have FolderContext whose root folder is defined by an arbitrary file path (or vscode.Uri).

This change requires 
- adding a `cwd` parameter to tasks to define where the task should be run as we cannot use the workspace folder scope.
- adding `FolderContext.name` to return a user friendly name for the folder
- launch configurations are edited to include the correct path based on the folder uri.
- to get the LSP server working I have to create a temporary `WorkspaceFolder` from the package uri.

I also deleted `executeSwiftTaskAndWait` and `executeShellTaskAndWait` as neither are used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -206,6 +206,10 @@
             "items": {
               "type": "string"
             }
+          },
+          "cwd": {
+            "description": "The folder to run the task in.",
+            "type": "string"
           }
         },
         "required": [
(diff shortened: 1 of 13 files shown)


0xTim (approved):
LGTM

PR #137, opened 2022-02-15
Package.swift in sub-folder

adam-fowler:
Sometimes projects consists of multiple components and the Swift elements might be in sub-folders of the project folder.

This PR allows for discovery of Package.swift files in a sub-folder of a workspace folder. The Package.swift can be found in two different ways
- When a folder is added to the workspace, it will check to see if there is a Package.swift in the root of that folder. If so then the folder is considered a project folder and added to the WorkspaceContext and thus LSP starts up, build tasks are generated etc.
- When a user opens a file it will check to see if that file is inside one of the currently active folders, if not it will then search down the directory structure until it reaches the workspace folder of that file. The last folder is finds that contains a Package.swift is considered the project folder for that file and it is added to the WorkspaceContext and LSP starts up etc

Issues:
- This does not support project folders that are inside project folders. I could have added support for this but then you would have the issue where a project that contains a Package.swift in its source would not load as it would consider the Package.swift file to be a project file. There is no way to tell if it a valid Package.swift without loading it which seems a little too much work every time you open a file. 
- Also by disallowing projects inside projects I have a shortcut to verifying a file is inside an already loaded project by just checking if the file is inside the folder. This means I don't have to keep searching down the tree for Package.swift files.
- I don't load all projects at startup. This is to avoid having to run resolve on loads of projects. If you have a workspace for a large number of swift projects eg the examples folder for hummingbird or maybe a project with loads of swift lambdas you possibly don't want to be running resolve on all these projects until you are actual working on them.

### DIFF
--- a/src/WorkspaceContext.ts
+++ b/src/WorkspaceContext.ts
@@ -13,10 +13,11 @@
 //===----------------------------------------------------------------------===//
 
 import * as vscode from "vscode";
+import * as path from "path";
 import { FolderContext } from "./FolderContext";
 import { StatusItem } from "./ui/StatusItem";
 import { SwiftOutputChannel } from "./ui/SwiftOutputChannel";
-import { execSwift, getSwiftExecutable, getXCTestPath } from "./utilities/utilities";
+import { execSwift, getSwiftExecutable, getXCTestPath, pathExists } from "./utilities/utilities";
 import { getLLDBLibPath } from "./debugger/lldb";
 import { LanguageClientManager } from "./sourcekit-lsp/LanguageClientManager";
 import { Version } from "./utilities/version";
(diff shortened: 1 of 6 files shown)


src/WorkspaceContext.ts
```diff
@@ -87,6 +87,20 @@ export class WorkspaceContext implements vscode.Disposable {
...
 
+    /** Add workspace folders at initialisation */
+    async addWorkspaceFolders() {
+        // add workspace folders, already loaded
+        if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) {
+            for (const folder of vscode.workspace.workspaceFolders) {
+                await this.addWorkspaceFolder(folder);
+            }
+        }
+        // fire focus event on null folder to startup language server if we don't have a currently focus folder
```
0xTim, line 98: Nit
```suggestion
        // fire focus event on null folder to startup language server if we don't have a currently focused folder
```

0xTim (approved):
One minor nit, otherwise LGTM

Issue #80, opened 2021-12-24
Windows Debug settings


adam-fowler:
>> I had to look up a few things here, so just to make sure I understand this correctly:
> 
> * Is CodeView the default on Windows (when just using `swift build` for example)?

> No, the default is no debug info :). `-g -debug-info-format=codeview` enables codeview, `-g` enables DWARF.

> > * Should we always add `-debug` when creating a debug build using CodeView?

> Yes, `-debug` should be specified always when building with CodeView debug info.

> > * What would be an appropriate name for this configuration setting? "Debug using DWARF"?

> That is what I would call it.  "Debug (DWARF)" and "Debug (CodeView)" is what I use.  But the using rather than the parenthetical is better for display purposes perhaps.

> What should be the default, if we want to make sure things "just work" for new developers?

> My preference would be CodeView.  There are some bugs that we do need to sort out, but the same can be said for DWARF (and arguably more).  However, if you do use CodeView, you can deal with mixed-language projects (C/C++/Swift) and debug across system code and Swift code.  Additionally, I've found that lldb is still not particularly stable and causes trouble, and using DWARF mandates the use of lldb (at this point).

_Originally posted by @compnerd in https://github.com/swift-server/vscode-swift/pull/75#discussion_r774765271_

Hi @compnerd,

I'm a little confused previously you said for DWARF we should use parameters `-use-ld=lld -debug:dwarf`. Now you are saying we just need `-g`.

As an aside the swift extension provides integration with the CoreLLDB extension so we will be targeting LLDB for the immediate future, so I assume we want DWARF.

compnerd:
> >> I had to look up a few things here, so just to make sure I understand this correctly:
> 
> > 
> 
> > * Is CodeView the default on Windows (when just using `swift build` for example)?
> 
> 
> 
> > No, the default is no debug info :). `-g -debug-info-format=codeview` enables codeview, `-g` enables DWARF.
> 
> 
> 
> > > * Should we always add `-debug` when creating a debug build using CodeView?
> 
> 
> 
> > Yes, `-debug` should be specified always when building with CodeView debug info.
> 
> 
> 
> > > * What would be an appropriate name for this configuration setting? "Debug using DWARF"?
> 
> 
> 
> > That is what I would call it.  "Debug (DWARF)" and "Debug (CodeView)" is what I use.  But the using rather than the parenthetical is better for display purposes perhaps.
> 
> 
> 
[...]

PR #148, opened 2022-02-21
Set windows DWARF debug options

adam-fowler:
Set windows debug options for DWARF debug generation as detailed in #80 

I'm unable to test these options, but they are what @compnerd suggested for DWARF debug output.
The option to do CodeView debug info will be in a different PR. I would rather someone else (with access to a Windows PC) did it though as I cannot test.

### DIFF
--- a/src/SwiftTaskProvider.ts
+++ b/src/SwiftTaskProvider.ts
@@ -40,6 +40,11 @@ interface TaskConfig {
     prefix?: string;
 }
 
+/** arguments for generating windows debug builds */
+function win32BuildOptions(): string[] {
+    return ["-Xswiftc", "-g", "-Xswiftc", "-use-ld=lld", "-Xlinker", "-debug:dwarf"];
+}
+
 /**
  * Creates a {@link vscode.Task Task} to build all targets in this package.
  */
@@ -49,7 +54,7 @@ function createBuildAllTask(folderContext: FolderContext): vscode.Task {
         additionalArgs.push("--enable-test-discovery");
     }
     if (process.platform === "win32") {
-        additionalArgs.push("-Xlinker", "-debug:dwarf");
+        additionalArgs.push(...win32BuildOptions());
     }
     let buildTaskName = SwiftTaskProvider.buildAllName;
     if (folderContext.relativePath.length > 0) {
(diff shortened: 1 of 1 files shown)


0xTim (approved):

compnerd (approved):
The argument changes look good to me.

compnerd:
@adam-fowler I think that the bigger question is how to hook that up into the build

adam-fowler:
> @adam-fowler I think that the bigger question is how to hook that up into the build. That is, how do we expose that to the user?

Sorry not sure what you mean. How do we expose what? The build options are added to the build task automatically

compnerd:
> Sorry not sure what you mean. How do we expose what? The build options are added to the build task automatically

I suspect that we will want to surface the debug information option to the user: do we use DWARF or do we use CodeView?  Or should we just unconditionally emit CodeView?

adam-fowler:
> I suspect that we will want to surface the debug information option to the user: do we use DWARF or do we use CodeView? Or should we just unconditionally emit CodeView?

We can add a config option to switch between the two. 

Debugging options seem a little light on Windows side if you don't have a full copy of Visual Studio. There is the extension you linked to that uses WinDbg, but it is 3 years old and doesn't have a great deal of downloads over that time, which makes me a little nervous on being reliant on it.

adam-fowler:
I've added #153 to discuss CodeView debug Info support

compnerd:
> Debugging options seem a little light on Windows side if you don't have a full copy of Visual Studio. There is the extension you linked to that uses WinDbg, but it is 3 years old and doesn't have a great deal of downloads over that time, which makes me a little nervous on being reliant on it.

You don't need a fully copy of Visual Studio for debugging on Windows.  The only piece of Visual Studio that we need installed for Swift on Windows is the build tools, and even that is strictly for a small subset of headers that are not available standalone (See `%ProgramFiles%\Microsoft Visual Studio\2022\*\Tools\MSVC\include`).  The Windows SDK is the one true hard dependency.  Both of these are for building code of course, not for debugging.

For debugging, I tend to find that [WinDBGX](https://www.microsoft.com/en-us/p/windbg-preview/9pgjgd53tn86?activetab=pivot:overviewtab) works quite well and is a standalone product that you can install from the Windows Store.  This actually is backed by DbgEngine.dll, which is the same debugging path as Visual Studio.  You can also use WinDBG, which is part of the debugging tools in the Windows SDK if you so desire.

PR #150, opened 2022-02-21
Add runSingleFile command

adam-fowler:
Add command that runs swift file through swift interpreter.
If file isn't saved it will save it to a temp file before running
Can you use this as a basic swift playground.


0xTim (approved):
This looks really useful and worthy of adding to the README

PR #152, opened 2022-02-21
Clear terminal window for every new build

adam-fowler:
Found this useful, when working on the Sendable issues. Kept trying to fix issues from previous builds


0xTim (approved):
Definite +1 on this, this has always tripped me up!

PR #154, opened 2022-02-28
Add TestExplorer with test discovery

adam-fowler:
This is the start of the TestExplorer work. I am going to break it into a number of PRs. This PR contains the basic test discovery. 

Function `discoverTestsInWorkspace` parses the output of `swift test --skip-build --list-test` and then adds/removes TestItems based on the results. This does mean you have the limitation that the project needs built.

Test discovery is run when a project gets focus, or when a build process has just finished.


0xTim (approved):
🎉

PR #156, opened 2022-03-01
Test runner

adam-fowler:
Add ability to run tests discovered in the previous PR #154 
- Builds test configuration
- Run tests
- Use CodeLLDB stdio tag to output result of tests to a file. Unfortunately it doesn't seem possible to output both stdout and stderr to the same file. macOS outputs to both, while non Darwin platforms output to just stdout
- Parse results of tests and flag which TestItems passed/failed
- Non Darwin platforms complicate things slightly by never outputting the test target name, so if two test targets have a test function with the same class and function name it is hard to differentiate them. The only way to do this is if one of them fails. At that point you can work out the test target from the file name which caused the error

### DIFF
--- a/src/TestExplorer/TestRunner.ts
+++ b/src/TestExplorer/TestRunner.ts
@@ -195,7 +195,8 @@ export class TestRunner {
         const lines = output.split("\n").map(item => item.trim());
 
         for (const line of lines) {
-            const passedMatch = /Test Case '-\[(\S+)\s(.*)\]' passed/.exec(line);
+            // Regex "Test Case '-[<test target> <class.function>]' passed"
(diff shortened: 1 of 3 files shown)


src/TestExplorer/TestRunner.ts
```diff
@@ -0,0 +1,306 @@
...
+                },
+                reason => {
+                    this.testRun.appendOutput(reason);
+                    this.testRun.end();
+                }
+            );
+    }
+
+    /**
+     *
```
0xTim, line 164: Looks like this might be missing a comment?

0xTim (approved):

PR #169, opened 2022-03-02
Conditionally add `--enable-test-discovery` flag

### LINKED ISSUE #166: Append `--enable-test-discovery` only when Swift < 5.4

stevapple:
Resolves #166

### DIFF
--- a/package.json
+++ b/package.json
@@ -147,7 +147,7 @@
           "swiftpm.testDiscovery.always": {
             "type": "boolean",
             "default": true,
-            "description": "Always enable test discovery for Swift packages on non-Darwin platforms."
+            "description": "Always enable test discovery for Swift packages on non-Darwin platforms.  Requires Swift 5.1 or later."
           }
         }
       }
--- a/package.json
+++ b/package.json
@@ -143,11 +143,6 @@
               "verbose"
             ],
             "description": "Traces the communication between VS Code and the SourceKit-LSP language server."
-          },
-          "swiftpm.testDiscovery.always": {
-            "type": "boolean",
-            "default": true,
-            "description": "Always enable test discovery for Swift packages on non-Darwin platforms.  Requires Swift 5.1 or later."
           }
         }
       }
--- a/package.json
+++ b/package.json
@@ -143,6 +143,11 @@
               "verbose"
             ],
             "description": "Traces the communication between VS Code and the SourceKit-LSP language server."
+          },
+          "swiftpm.testDiscovery.always": {
+            "type": "boolean",
+            "default": true,
+            "description": "Always enable test discovery for Swift packages on non-Darwin platforms."
           }
         }
       }
--- a/src/LinuxMain.ts
+++ b/src/LinuxMain.ts
@@ -1,0 +1,47 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VSCode Swift open source project
+//
+// Copyright (c) 2022 the VSCode Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of VSCode Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import * as vscode from "vscode";
+import { FolderContext } from "./FolderContext";
+import { pathExists } from "./utilities/utilities";
+
+/**
+ * Cache the existence of `Tests/LinuxMain.swift`.
+ */
+export class LinuxMain {
+    private fileWatcher: vscode.FileSystemWatcher;
+    public exists?: boolean;
+
+    constructor(private folderContext: FolderContext) {
+        pathExists(this.folderContext.folder.fsPath, "Tests", "LinuxMain.swift").then(
+            exists => (this.exists = exists)
+        );
+
+        this.fileWatcher = vscode.workspace.createFileSystemWatcher(
+            new vscode.RelativePattern(this.folderContext.folder, "Tests/LinuxMain.swift"),
+            false,
+            true
+        );
+        this.fileWatcher.onDidCreate(() => (this.exists = true));
+        this.fileWatcher.onDidDelete(() => (this.exists = false));
+    }
+
+    /**
+     * Disposes the {@link vscode.FileSystemWatcher file system watcher} when
+     * the extension deactivates.
+     */
+    dispose() {
+        this.fileWatcher?.dispose();
+    }
+}
--- a/src/LinuxMain.ts
+++ b/src/LinuxMain.ts
@@ -13,7 +13,6 @@
 //===----------------------------------------------------------------------===//
 
 import * as vscode from "vscode";
-import { FolderContext } from "./FolderContext";
 import { pathExists } from "./utilities/utilities";
 
 /**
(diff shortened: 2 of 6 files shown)


stevapple:
Question: What to do when there's neither `LinuxMain.swift` nor the test discovery feature?

src/utilities/utilities.ts
```diff
@@ -173,3 +176,27 @@ export async function getXCTestPath(): Promise<string> {
     }
     return path.join(platformPath, "Developer", "Library", `XCTest-${version}`, "usr", "bin");
 }
+
+/**
+ * @returns SwiftPM flag for enabling test discovery
+ */
+export async function testDiscoveryFlag(ctx: FolderContext): Promise<string[]> {
```
adam-fowler, line 183: With this we have a couple of functions (see also win32BuildOptions) that are used to provide build options. They are in two different files. Perhaps we should centre them in one file BuildOptions.ts

package.json
```diff
@@ -143,6 +143,11 @@
               "verbose"
             ],
             "description": "Traces the communication between VS Code and the SourceKit-LSP language server."
+          },
+          "swiftpm.testDiscovery.always": {
+            "type": "boolean",
```
adam-fowler, line 148: Do we really need the config option?

src/utilities/utilities.ts
```diff
@@ -173,3 +176,27 @@ export async function getXCTestPath(): Promise<string> {
...
+    // Test discovery is only available in SwiftPM 5.1 and later.
+    if (ctx.workspaceContext.swiftVersion.isLessThan(new Version(5, 1, 0))) {
+        return [];
+    }
+    // Test discovery is always enabled on Darwin.
+    if (process.platform !== "darwin" && ctx.swiftPackage.getTargets("test").length > 0) {
+        const alwaysDiscoverTests = vscode.workspace
+            .getConfiguration("swiftpm")
+            .get<boolean>("testDiscovery.always", true);
+        const hasLinuxMain = await pathExists(ctx.folder.fsPath, "Tests", "LinuxMain.swift");
```
adam-fowler, line 193: Instead of testing for the existence of `LinuxMain.swift` all the time, perhaps the existence of it should be stored in SwiftPackage and setup in `SwiftPackage.create`

src/TestExplorer/TestExplorer.ts
```diff
@@ -84,9 +84,13 @@ export class TestExplorer {
      */
     async discoverTestsInWorkspace() {
         try {
-            const { stdout } = await execSwift(["test", "--skip-build", "--list-tests"], {
-                cwd: this.folderContext.folder.fsPath,
-            });
+            const enableTestDiscovery = await testDiscoveryFlag(this.folderContext);
+            const { stdout } = await execSwift(
```
adam-fowler, line 88: You shouldn't need the test discovery flag here.

package.json
stevapple: There’s still someone who prefers using `LinuxMain` (e.g. for CI compatibility), and I think we’d better offer such choice.

src/utilities/utilities.ts
stevapple: I moved it here because it is shared across multiple functionalities. If `win32BuildOptions` is also shared (currently not) I believe that’s worth consideration.

src/TestExplorer/TestExplorer.ts
stevapple: 🤔Oh, I didn’t see that it is automatically triggered after build tasks.

Yet I’m still wondering whether it makes sense to append `--skip-build` here. Does it mean that a user needs to trigger a build task manually so as to see the tests?

package.json
adam-fowler: Fair point. Wondering how much work we put into supporting earlier versions of swift.

src/utilities/utilities.ts
adam-fowler: I think a BuildOptions file is still worthwhile. The Windows options are probably going to get more complex if we look to add options for CodeView

src/TestExplorer/TestExplorer.ts
adam-fowler: If you have a large project, which needs completely rebuilt to get a list of tests, all the user will see is an empty test view for ages. I was going to try and add a message to say "please build your project if you want to view tests".

stevapple: I see. Gonna resolve this after my sleep:)

src/utilities/utilities.ts
stevapple: Since the flag is not necessary for `TestExplorer`, I’m going to move it back to `SwiftTaskProvider.ts` for this pitch. Maybe some day we’ll need `BuildOptions.ts` but I think that’s not now🤔 We only have ~10 lines and 2 functions now.

src/LinuxMain.ts
```diff
@@ -0,0 +1,48 @@
...
+        pathExists(this.folderContext.folder.fsPath, "Tests", "LinuxMain.swift").then(
+            exists => (this.exists = exists)
+        );
+
+        this.fileWatcher = vscode.workspace.createFileSystemWatcher(
+            new vscode.RelativePattern(this.folderContext.folder, "Tests/LinuxMain.swift"),
+            true,
+            false,
+            true
+        );
```
adam-fowler, line 36: Are these flags correct `ignoreCreateEvents: true, ignoreChangeEvents: false, ignoreDeleteEvents: true`

src/PackageWatcher.ts
```diff
@@ -22,7 +22,7 @@ import { FolderEvent, WorkspaceContext } from "./WorkspaceContext";
  * Any changes to these files will update the context keys, trigger a `resolve` task,
  * and update the Package Dependencies view.
  */
-export class PackageWatcher {
+export class PackageWatcher implements vscode.Disposable {
```
adam-fowler, line 25: The `implements vscode.Disposable` isn't really necessary

src/SwiftTaskProvider.ts
```diff
@@ -40,6 +41,28 @@ interface TaskConfig {
     prefix?: string;
 }
 
+/** flag for enabling test discovery */
+async function testDiscoveryFlag(ctx: FolderContext): Promise<string[]> {
```
adam-fowler, line 45: This shouldn't need to be async anymore

src/SwiftTaskProvider.ts
```diff
@@ -48,10 +71,10 @@ function win32BuildOptions(): string[] {
 /**
  * Creates a {@link vscode.Task Task} to build all targets in this package.
  */
-function createBuildAllTask(folderContext: FolderContext): vscode.Task {
+async function createBuildAllTask(folderContext: FolderContext): Promise<vscode.Task> {
```
adam-fowler, line 74: This shouldn't need to be async anymore

src/SwiftTaskProvider.ts
```diff
@@ -192,7 +215,7 @@ export class SwiftTaskProvider implements vscode.TaskProvider {
             if (!folderContext.swiftPackage.foundPackage) {
                 continue;
             }
-            tasks.push(createBuildAllTask(folderContext));
+            tasks.push(await createBuildAllTask(folderContext));
```
adam-fowler, line 218: This shouldn't need to be async anymore

src/TestExplorer/TestExplorer.ts
```diff
@@ -19,7 +19,7 @@ import { FolderEvent, WorkspaceContext } from "../WorkspaceContext";
 import { TestRunner } from "./TestRunner";
 
 /** Build test explorer UI */
-export class TestExplorer {
+export class TestExplorer implements vscode.Disposable {
```
adam-fowler, line 22: Again is this necessary? Just wondering why you've added them. I guess they give more information about the class

src/LinuxMain.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Cache the existence of `Tests/LinuxMain.swift`.
+ */
+export class LinuxMain implements vscode.Disposable {
+    private fileWatcher: vscode.FileSystemWatcher;
+    public exists?: boolean;
+
+    constructor(private folderContext: FolderContext) {
+        pathExists(this.folderContext.folder.fsPath, "Tests", "LinuxMain.swift").then(
```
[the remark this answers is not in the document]
stevapple: This change would make all `hasLinuxMain` dependent `async` IMO.  I’m wondering if it’s worth it.
Another solution is to allow `hasLinuxMain` return `undefined` and let the caller decide what to do. This will add to complexity at call site.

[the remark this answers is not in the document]
adam-fowler: If you call `LinuxMain.create()` in `WorkspaceContext.create()` and pass the fully built `LinuxMain` into the `WorkspaceContext` constructor, anyone accessing `linuxMain` via `WorkspaceContext` won't have to deal with an undefined `hasLinuxMain`

0xTim (approved):
LGTM other that Adam's comments. I'm fine with ignoring Swift versions pre 5.1 for certain

package.json
adam-fowler: I understand what it is for. I just don't think it is something we need to support. I don't want to clutter the configuration with settings that get used by virtually no one.

adam-fowler:
@stevapple Do you think you are going to have time to resolve the remaining issues for this to get merged.

adam-fowler (approved):
LGTM

PR #171, opened 2022-03-03
Always show package dependencies view

adam-fowler:
As long as there is a package, even if there aren't any package dependencies.

The view has useful shortcut buttons (resolve, reset, update package) that you lose if it is hidden

Thoughts?


0xTim (approved):
Yeah I agree, if it's a package we should display the package view

PR #177, opened 2022-03-07
Setup docker for ci.swiftserver.group

adam-fowler:


0xTim (approved):

adam-fowler:
@0xTim given we aren't running tests at the moment there is no need to check against multiple version of Swift. Shall we reduce the checks down to one (swift:5.5) for the moment.

0xTim:
Yeah I think that's fair

PR #182, opened 2022-03-08
Add feature request issue template

adam-fowler:
I'm not necessarily a fan of this template but I think it is required if we have a bug report one


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: ''
```
0xTim, line 5: We should auto add enhancements as well
```suggestion
labels: 'enhancement'
```

PR #186, opened 2022-03-09
Add SwiftToolchain class

adam-fowler:
Stores details about the current swift toolchain. Collects various functions into one class

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -1,0 +1,123 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VSCode Swift open source project
+//
+// Copyright (c) 2021 the VSCode Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of VSCode Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import * as fs from "fs/promises";
+import * as path from "path";
+import * as plist from "plist";
+import configuration from "../configuration";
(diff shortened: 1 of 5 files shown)


src/toolchain/toolchain.ts
```diff
@@ -0,0 +1,123 @@
...
+    private static async getDeveloperDir(): Promise<string | undefined> {
+        try {
+            switch (process.platform) {
+                case "darwin": {
+                    const { stdout } = await execFile("xcode-select", ["-p"]);
+                    return stdout.trimEnd();
+                }
+
+                case "win32": {
+                    const developerPath = process.env.DEVELOPER_DIR;
```
stevapple, line 80: This should be exactly what `DeveloperDir` is on Windows, so I think the naming is improper. Change it back into `getXCTestDir` instead?

stevapple (changes requested):

src/toolchain/toolchain.ts
adam-fowler: Sorry yeah I got that one wrong.

adam-fowler: I've split it into two functions `getDeveloperDir` and `getXCTestPath`

adam-fowler: @stevapple would you be able to verify I haven't broken Windows

stevapple:
It works:)

stevapple (approved):

src/toolchain/toolchain.ts
```diff
@@ -0,0 +1,150 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VSCode Swift open source project
+//
+// Copyright (c) 2021 the VSCode Swift project authors
```
0xTim, line 5: We should update the script to update the date to be 2021-2022

PR #187, opened 2022-03-09
DNM: test PR

tomerd:


docker/docker-compose.yaml
```diff
@@ -28,7 +28,8 @@ services:
 
   test:
     <<: *common
-    command: /bin/bash -xcl "npm ci && npm run format && npm run compile && npm run package"
+    command: /bin/bash -xcl "mkdir /tmp/code && cp -r ./ /tmp/code/ && cd /tmp/code && npm ci && npm run format && npm run compile && npm run package"
```
adam-fowler, line 31: Was wondering if you could run chmod on the node_modules folder after the `npm ci` call so subsequent runs can clean up the folder.
```suggestion
    command: /bin/bash -xcl "npm ci && chmod -R a+rw node_modules && npm run format && npm run compile && npm run package"
```

tomerd: not a problem that chmod can solve - the build process of the offending module writes files as a different user which does not exist out side the container so you must be sudo to delete them

PR #197, opened 2022-03-16
Fix CI: Make new folder and copy everything into new folder

adam-fowler:


tomerd:

tomerd:

0xTim (approved):
We're no longer running `npm compile` - is that deliberate?

adam-fowler:
> We're no longer running `npm compile` - is that deliberate?

`npm run package` will do a compile so there is no need for the separate call.

PR #198, opened 2022-03-18
Migrate to upstream LSP inlay hints on the client side

fwcd:
Official support for inlay hints in LSP will soon make the custom implementation superfluous. For more information, see the accompanying server-side PR (which this PR should be considered blocked on):

- https://github.com/apple/sourcekit-lsp/pull/465

This PR removes the client-side implementation of inlay hints as custom text decorations, which will no longer be needed once VSCode's and `vscode-languageclient`'s support for native inlay hints ships in the next release.

Additionally, `vscode-languageclient` is upgraded to a beta release, so native inlay hints can already be tested in a recent VSCode Insiders build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -174,7 +174,8 @@
           "sourcekit-lsp.inlayHints.enabled": {
             "type": "boolean",
             "default": true,
-            "description": "Render inlay type annotations in the editor. Inlay hints require Swift 5.6 or later.",
+            "description": "Render inlay type annotations in the editor. Inlay hints require Swift 5.6.",
+            "deprecationMessage": "SourceKit-LSP's custom inlay hints are deprecated as Swift 5.7 and newer use LSP-native inlay hints. These can be enabled using the setting 'editor.inlayHints.enabled'.",
             "order": 3
           },
           "sourcekit-lsp.trace.server": {
--- a/package.json
+++ b/package.json
@@ -175,7 +175,7 @@
             "type": "boolean",
             "default": true,
             "description": "Render inlay type annotations in the editor. Inlay hints require Swift 5.6.",
-            "deprecationMessage": "SourceKit-LSP's custom inlay hints are deprecated as Swift 5.7 and newer use LSP-native inlay hints. These can be enabled using the setting 'editor.inlayHints.enabled'.",
+            "markdownDeprecationMessage": "**Deprecated**: Please use `#editor.inlayHints.enabled#` instead.",
             "order": 3
           },
           "sourcekit-lsp.trace.server": {
--- a/src/sourcekit-lsp/LanguageClientManager.ts
+++ b/src/sourcekit-lsp/LanguageClientManager.ts
@@ -26,7 +26,7 @@ import {
 } from "../utilities/utilities";
 import { Version } from "../utilities/version";
 import { FolderEvent, WorkspaceContext } from "../WorkspaceContext";
-import { activateInlayHints } from "./inlayHints";
+import { activateLegacyInlayHints } from "./inlayHints";
 import { FolderContext } from "../FolderContext";
 
 /** Manages the creation and destruction of Language clients as we move between
@@ -69,7 +69,7 @@ export class LanguageClientManager {
     private onDidCreateFileDisposable: vscode.Disposable;
     private onDidDeleteFileDisposable: vscode.Disposable;
     private onChangeConfig: vscode.Disposable;
-    private inlayHints?: vscode.Disposable;
+    private legacyInlayHints?: vscode.Disposable;
     private supportsDidChangedWatchedFiles: boolean;
     private restartedPromise?: Promise<void>;
     private currentWorkspaceFolder?: vscode.Uri;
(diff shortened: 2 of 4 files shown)


adam-fowler:
Hi @fwcd thanks for this.

I do actually have a separate branch which implements this in a slightly different manner. See PR #199 

I must say I don't understand Microsoft's versioning for the language client. It appears they don't have a proper release of 8.0.0 and the `next` releases are kinda beta releases. But nothing says this explicitly. Given this I would rather wait on a full 8.0.0 release, before moving up.

I see this also relies on changes to sourcekit-lsp so it won't work until there is a new release of Swift.

Eventually your solution is correct and we should just rely on the LanguageClient to provide these. In the meantime it might be better to go for mine.

fwcd:
Yeah, that sounds good. I've created this branch this early mostly to have a client for playing around with the new sourcekit-lsp that supports the proposed LSP request. Holding off until Swift and VSCode ship new releases sounds totally reasonable. Once ready, we should definitely also move to a stable vscode-languageclient again.

adam-fowler:
I see vscode-languageclient v8.0.0 is out and that your pull request for the `sourcekit-lsp` changes is proceeding. I would like to split this PR into two the upgrade to 8.0.0 and then using the new InlayHint requests. 

If it is ok with you I would like to do the v8 upgrade as I would like to see what new stuff it has. I'll leave the InlayHint code to you. Remember we need to still support the old InlayHints method as Swift 5.6 requires it.

fwcd:
That sounds good. The v8 update unfortunately includes a breaking change, the `LanguageClient.onReady` method is gone, so we have to keep track of this promise manually after starting the client.

Regarding the old inlay hints, we might want to deprecate (and, by default, disable) the `sourcekit-lsp.inlayHints.enabled` option, since requesting the non-standard hints on a newer `sourcekit-lsp` (that uses the standard inlay hints) would cause errors I think. Though that might also be better placed in another PR.

I will see if I can slim down this PR to only include the v8 update (+ necessary changes due to API breaks).

adam-fowler:
> That sounds good. The v8 update unfortunately includes a breaking change, the `LanguageClient.onReady` method is gone, so we have to keep track of this promise manually after starting the client.
> 
Yeah saw that
> Regarding the old inlay hints, we might want to deprecate (and, by default, disable) the `sourcekit-lsp.inlayHints.enabled` option, since requesting the non-standard hints on a newer `sourcekit-lsp` (that uses the standard inlay hints) would cause errors I think. Though that might also be better placed in another PR.
> 
We can deprecate it and point users to the standard hint display setting
> I will see if I can slim down this PR to only include the v8 update (+ necessary changes due to API breaks).
I'm looking at the v8 update just now. It has broken a couple of things related to restarting the server so I need to resolve these first

adam-fowler:
You still need to support the old inlay hints for people running swift 5.6. You can check if someone is running Swift 5.6 or earlier using
```
if(workspaceContext.swiftVersion.isLessThan(new Version(5, 7, 0)) {
    ...
}
```

package.json
```diff
@@ -402,6 +402,6 @@
   "dependencies": {
     "@types/plist": "^3.0.2",
     "plist": "^3.0.5",
-    "vscode-languageclient": "^8.0.0"
+    "vscode-languageclient": "^8.0.1"
```
adam-fowler, line 405: Any reason to force 8.0.1?

fwcd: No specific reason, this was in my previous patch, but there shouldn't be anything specific to 8.0.1 here.

fwcd:
Good catch, the branch should now only disable inlay hints for Swift >= 5.7. For clarity and to avoid ambiguity with the new LSP types, I have also prefixed the non-standard inlay hint structures with 'legacy'.

adam-fowler:
From what I can see this looks good. I am going to hold off merging it though until there is a 5.7 development build released that includes the new InlayHInt support. Does that sound ok?

fwcd:
Sure!

adam-fowler:
Can you deprecate the `sourcekit-lsp.inlayHints.enabled` setting and link to the `editor.inlayHints.enabled` in the deprecation message.

fwcd:
Done!

package.json
```diff
@@ -174,7 +174,8 @@
           "sourcekit-lsp.inlayHints.enabled": {
             "type": "boolean",
             "default": true,
-            "description": "Render inlay type annotations in the editor. Inlay hints require Swift 5.6 or later.",
+            "description": "Render inlay type annotations in the editor. Inlay hints require Swift 5.6.",
+            "deprecationMessage": "SourceKit-LSP's custom inlay hints are deprecated as Swift 5.7 and newer use LSP-native inlay hints. These can be enabled using the setting 'editor.inlayHints.enabled'.",
```
adam-fowler, line 178: If you use the markdown deprecation you can add a link to the setting. Also lets keep the message simple.
```suggestion
            "markdownDeprecationMessage": "**Deprecated**: Please use `#editor.inlayHints.enabled#` instead.",
```

adam-fowler (approved):
Is see the new InlayHint support is now in the swift 5.7 nightlies, so lets merge this

PR #203, opened 2022-03-21
Add support for loading Swift 5.6 Package resolved

adam-fowler:
Swift 5.6 incremented the version number of Package.resolved and changed format slightly.

This PR adds support for loading both v1 and v2 of Package.resolved
Also adds a test to make sure we can load both


0xTim (approved):
Semantically versioned Swift 🙃 

At least we have a resolved version file to check

PR #208, opened 2022-03-23
Test Explorer improvements

adam-fowler:
- Don't run non-debug tests through the debugger
  - This means when they crash you aren't stuck inside a debugger 
  - I have access to `stdout` and `stderr` from the test process and can stream them into Test Explorer output pane, thus giving the user better feedback. On macOS test output comes in a mixture of `stdout` and `stderr` and these buffer at different rates which makes life difficult. So temporarily I have disabled showing `stdout` because test results are output to `stderr`
- Added proper cancellation processing. If you cancel a test run the test process will be killed.
- Display Test Explorer output as soon as testing starts.

### DIFF
--- a/src/TestExplorer/TestRunner.ts
+++ b/src/TestExplorer/TestRunner.ts
@@ -17,6 +17,8 @@ import * as fs from "fs/promises";
 import * as path from "path";
 import { createTestConfiguration, createDarwinTestConfiguration } from "../debugger/launch";
 import { FolderContext } from "../FolderContext";
+import { execFile } from "../utilities/utilities";
+import { createBuildAllTask, executeTaskAndWait } from "../SwiftTaskProvider";
 
 /** Class used to run tests */
 export class TestRunner {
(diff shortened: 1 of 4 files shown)


src/TestExplorer/TestRunner.ts
```diff
@@ -106,100 +108,207 @@ export class TestRunner {
...
+     * @returns
+     */
+    private createLaunchConfigurationForTesting(
+        debugging: boolean,
+        outputFile: string | null = null
+    ): vscode.DebugConfiguration | null {
+        const testList = this.testItems.map(item => item.id).join(",");
+
+        if (process.platform === "darwin") {
+            // if debugging on macOS need to create a custom profile so we can set the system
```
0xTim, line 141: I'm assuming this should be
```suggestion
            // if debugging on macOS need to create a custom procfile so we can set the system
```

0xTim (approved):
LGTM - one comment spelling query

src/TestExplorer/TestRunner.ts
adam-fowler: It's a custom launch configuration, I'll change from profile to that

PR #209, opened 2022-03-23
Add regex to catch skipped tests

adam-fowler:
And then mark the TestItem as skipped


0xTim (approved):
I won't pretend to be able to tell the regex is correct 😂 but looks good to me

adam-fowler:
> I won't pretend to be able to tell the regex is correct 😂 but looks good to me

regex101.com is the best site for testing this kind of thing

PR #213, opened 2022-03-24
Extract Linux test search

adam-fowler:
Extract Linux test search into separate function and use with skipped tests as well.


src/TestExplorer/TestRunner.ts
```diff
@@ -445,4 +430,40 @@ export class TestRunner {
             }
         }
     }
+
+    /**
+     * Linux test output does not include the target name. So I have to work out which target
+     * the test is in via the test name and if it failed the filename from the error. In theory
+     * If a test fails the filename for where it failed should indicate which target it is in.
```
0xTim, line 437: Very minor nit
```suggestion
     * if a test fails the filename for where it failed should indicate which target it is in.
```

0xTim (approved):
LGTM

PR #216, opened 2022-03-28
CI: Run tests on swiftserver

adam-fowler:
Run `npm test`
Requires it to be run by `xvfb` in a separate shell file. If you include in docker-compose file it doesn't run
Had to install a bunch of additional stuff in Dockerfile to get it to work
Also added fix for test that only works in swift 5.6

### DIFF
--- a/docker/test.sh
+++ b/docker/test.sh
@@ -1,0 +1,1 @@
+xvfb-run -a npm test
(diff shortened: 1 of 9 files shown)


adam-fowler:
@tomerd if you are happy with this can you enable the swift 5.6 and 5.4 targets.

tomerd:

docker/test.sh
```diff
@@ -0,0 +1 @@
+xvfb-run -a npm test
```
tomerd, line 1: probably belongs in `script` rather than `docker` directory, but up to you

adam-fowler: I'm happy with it where it is. It's a single line script which is only ever called by the docker-compose file

0xTim (approved):

PR #219, opened 2022-03-30
Add Language status item for Swift

adam-fowler:
Provides a menu that displays the current version of Swift and the option to jump to the Package.swift for the currently edited file. The menu is accessible via the `{ }` icon next to the Swift label in the bottom right hand corner. 

This is a standard now. Check the same for typescript and json in the vscode-swift project.

This requires v1.65.0 of VSCode.

### DIFF
--- a/src/ui/LanguageStatusItems.ts
+++ b/src/ui/LanguageStatusItems.ts
@@ -39,27 +39,27 @@ export class LanguageStatusItems implements vscode.Disposable {
             "swiftlang-version",
             LanguageStatusItems.documentSelector
         );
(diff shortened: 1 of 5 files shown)


src/ui/LanguageStatusItems.ts
```diff
@@ -0,0 +1,75 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VSCode Swift open source project
+//
+// Copyright (c) 2021 the VSCode Swift project authors
```
0xTim, line 5: The copyright should be updated for new files to point to 2022

0xTim (approved):

src/ui/LanguageStatusItems.ts
adam-fowler: Damn keep missing that

PR #220, opened 2022-04-01
Task manager

adam-fowler:
It looks like there can only be one `vscode.tasks.onDidEndTaskProcess` handler running at any one time. So running two tasks at the same time might mean one of the tasks isn't registered as completed.

This PR centralise all task tracking in one place. Instead of calling `vscode.tasks.onDidEndTaskProcess` in multiple places. Have the TaskManager make this call and then other systems can subscribe to these message with TaskManagers own version of `onDidEndTaskProcess`


adam-fowler:
I would like to get this into a point release as it is the cause of all sorts of weirdness

0xTim (approved):

PR #231, opened 2022-04-06
Experimental: Background compilation

adam-fowler:
Adds settings config to run compilation in the background whenever a file is saved. It is currently defaulted to off


0xTim (approved):
LGTM

PR #235, opened 2022-04-11
Add Environment Variables setting for tests

adam-fowler:
Given there is no launch config for running tests there is no way to add environment variables when running tests. This PR resolves this by adding a setting for environment variables which are used when running tests.


0xTim (approved):
Great addition!

PR #241, opened 2022-04-13
debugger: simplify the logic for launching a test on Windows

compnerd:
Create a local variable with the `xctestPath` similar to the Darwin
case.  This reduces the line length subsequently and makes the test
match the use rather than use a proxy variable for the test.


adam-fowler (approved):
That makes much better sense that what was there before

adam-fowler:

PR #242, opened 2022-04-13
toolchain: remove dependency on `DEVELOPER_DIR` on Windows

compnerd:
This environment variable is not stable and should be avoided going
forward.  The toolchain no longer relies on it and will be removed in a
future toolchain release.  The paths which are being computed relative
to these variables can be computed in alternative manners.  This should
be safe with older releases and newer releases of the toolchain on
Windows.

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -34,30 +34,29 @@ export class SwiftToolchain {
         public swiftVersionString: string,
         public swiftVersion: Version,
         public toolchainPath?: string,
-        public developerDir?: string,
+        public sdkroot?: string,
         public xcTestPath?: string
     ) {}
 
     static async create(): Promise<SwiftToolchain> {
         const version = await this.getSwiftVersion();
         const toolchainPath = await this.getToolchainPath();
-        const developerDir = await this.getDeveloperDir();
-        let xcTestPath: string | undefined;
-        if (developerDir) {
-            xcTestPath = await this.getXCTestPath(developerDir);
-        }
+        const sdkroot = this.getSDKROOT();
+        const xcTestPath = await this.getXCTestPath(sdkroot);
         return new SwiftToolchain(
             version.name,
             version.version,
             toolchainPath,
-            developerDir,
+            sdkroot,
             xcTestPath
         );
     }
 
     logDiagnostics(channel: SwiftOutputChannel) {
         channel.logDiagnostic(`Toolchain Path: ${this.toolchainPath}`);
-        channel.logDiagnostic(`Developer Dir: ${this.developerDir}`);
+        if (this.sdkroot) {
+            channel.logDiagnostic(`SDKROOT: ${this.sdkroot}`);
+        }
         if (this.xcTestPath) {
             channel.logDiagnostic(`XCTestPath: ${this.xcTestPath}`);
         }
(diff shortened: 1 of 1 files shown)


adam-fowler:

src/toolchain/toolchain.ts
```diff
@@ -35,29 +35,37 @@ export class SwiftToolchain {
         public swiftVersion: Version,
         public toolchainPath?: string,
         public developerDir?: string,
```
adam-fowler, line 37: Is it necessary to store the developer directory anymore? It is only needed temporarily when calculating the xctest folder for Xcode. We could move the `xcode-select -p` call inside `getXCTestPath`.

adam-fowler:
Does Swift 5.5 and earlier still require the developer dir environment variable?

src/toolchain/toolchain.ts
```diff
@@ -35,29 +35,37 @@ export class SwiftToolchain {
...
+        public sdkroot?: string,
```
adam-fowler, line 38: Also is it necessary to store the `sdkroot` if it is only used in `getXCTestPath`?

stevapple: I guess we will definitely use it elsewhere, eg. in pre-5.7 compile flags using with script mode.

It’s a question if we should store `SDKROOT` for Darwin too because it also makes some sense there.

adam-fowler: Ok we should store `SDKROOT` if we are going to use it in the future. Not sure we need it for Darwin though

compnerd:
Yes, 5.6 and earlier will use it, but it is safe if you do not.  The results will be the same in those releases.

src/toolchain/toolchain.ts
compnerd: I don't think it is - but I don't know why it was being saved in the first place.

src/toolchain/toolchain.ts
compnerd: Could you please be explicit if you would like me to make that change?  I don't have a strong opinion on whether we collect and store the value or not.

src/toolchain/toolchain.ts
adam-fowler: Originally it was stored so we could calculate the xcTestPath in a different part of the codebase. With all these collected together these is no need to store it.

src/toolchain/toolchain.ts
adam-fowler: We should continue to store sdkroot. So no change needed here

adam-fowler:
@compnerd you need to install the `Prettier - Code Formatter` extension in vscode to format the code. With it installed every time you save it will format the code correctly. 

The merge is being blocked based on code formatting issues.

src/toolchain/toolchain.ts
```diff
@@ -85,42 +84,28 @@ export class SwiftToolchain {
...
-                return path.join(developerDir, "usr", "bin");
-
+            case "darwin": {
+                const { stdout } = await execFile("xcode-select", ["-p"]);
+                return path.join(stdout.trimEnd(), "usr", "bin");
+            }
             case "win32": {
-                const platformPath = path.join(developerDir, "Platforms", "Windows.platform");
+                if (!sdkroot) {
+                    return undefined;
```
adam-fowler, line 106: Should we allow the extension to startup if someone hasn't got an SDKROOT environment variable. Tests won't work. Should we throw an Error here instead. VSCode will catch it and display the error. Something like
```ts
throw Error("SDKROOT environment variable should be set");
```

compnerd: My personal opinion: we should.  The `SDKROOT` environment variable is just a convenience thing.  If you do not have that set (and I often do not), I expect that the user has intentionally not installed or uninstalled the build platform's SDK.  In such a case, they would need _some_ SDK and the assumption is that they are compiling to a foreign platform.  That is to say, this is setup in such a way that if I do not setup `SDKROOT`, I can simply pass `-sdk C:\Library\Developer\Platforms\Android.platform\Developer\SDKs\Android.sdk -target armv7-unknown-linux-androideabi` and use the toolchain to build for android :)

adam-fowler: ok

adam-fowler (approved):

PR #243, opened 2022-04-13
TestExplorer: repair the messaging on Windows

compnerd:
The Windows error message does not match Darwin nor Linux.  Add a
special case to ensure that the message is displayed properly.


tomerd:

adam-fowler:
> The Windows error message does not match Darwin nor Linux. Add a
> special case to ensure that the message is displayed properly.

Any reason there is an inconsistency here?

adam-fowler (approved):
LGTM

PR #247, opened 2022-04-14
TestExplorer: fold duplicate if/else (NFC)

compnerd:
Fold away the conditional as both sides are identical.  This simply
reduces the code but should have no impact.


adam-fowler (approved):
Yeah originally these were different. No point having the if now

PR #248, opened 2022-04-14
debugger: repair the Windows launching of tests

compnerd:
The tests would silently fail previously due to the `Path` being setup
improperly.  We assumed that we could create a string that could be
evaluated by Powershell which is not the case as the environment is
passed to the node.js function `execAsync`, but would also mean that we
would be reliant on a specific shell.  The mismatched case would also
result in `PATH` and `Path` being set, and it is unclear which would
actually get used.


adam-fowler (approved):
LGTM

PR #252, opened 2022-04-14
Introduce `getExecutableName`

stevapple:
Introduce `getExecutableName` to append `.exe` suffix for tools on Windows.

PR #253, opened 2022-04-14
Judge and cache if the new driver is used

stevapple:
Based on #252, this PR introduce a new cached variable to judge if the new driver is being used.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -187,6 +187,10 @@ async function runSingleFile(ctx: WorkspaceContext) {
         return;
     }
 
+    if (process.platform === "win32" && !ctx.toolchain.newSwiftDriver) {
+        return;
+    }
+
     let filename = document.fileName;
     let isTempFile = false;
     if (document.isUntitled) {
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -188,6 +188,9 @@ async function runSingleFile(ctx: WorkspaceContext) {
     }
 
     if (process.platform === "win32" && !ctx.toolchain.newSwiftDriver) {
+        await vscode.window.showErrorMessage(
+            "Run Swift Script is unavailable with the legacy driver on Windows."
+        );
         return;
     }
 
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -179,7 +179,7 @@ export class SwiftToolchain {
         if (await pathExists(toolDirectory, getExecutableName("swift-driver"))) {
             return true;
         }
-        if (await !pathExists(toolDirectory, getExecutableName("swift-frontend"))) {
+        if ((await pathExists(toolDirectory, getExecutableName("swift-frontend"))) !== true) {
             return false;
         }
         // check if swift is symlinked to swift-frontend
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -17,7 +17,7 @@ import * as path from "path";
 import * as plist from "plist";
 import configuration from "../configuration";
 import { SwiftOutputChannel } from "../ui/SwiftOutputChannel";
-import { execFile, execSwift } from "../utilities/utilities";
+import { execFile, execSwift, getExecutableName, pathExists } from "../utilities/utilities";
 import { Version } from "../utilities/version";
 
 /**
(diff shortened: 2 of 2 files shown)


src/toolchain/toolchain.ts
```diff
@@ -156,4 +162,33 @@ export class SwiftToolchain {
...
+            return true;
+        }
+        if (process.env.SWIFT_USE_OLD_DRIVER) {
+            return false;
+        }
+        // judge from tool existence
+        if (await pathExists(toolDirectory, getExecutableName("swift-driver"))) {
+            return true;
+        }
+        if (await !pathExists(toolDirectory, getExecutableName("swift-frontend"))) {
```
adam-fowler, line 182: Is there any situation where `swift-driver` doesn't exist and the new driver is used.

src/toolchain/toolchain.ts
```diff
@@ -156,4 +162,33 @@ export class SwiftToolchain {
...
+            return false;
+        }
+        // check if swift is replaced by the new driver
+        const swiftDriverBuffer = await fs.readFile(
+            path.join(toolDirectory, getExecutableName("swift"))
+        );
+        const swiftFrontendBuffer = await fs.readFile(
+            path.join(toolDirectory, getExecutableName("swift-frontend"))
+        );
+        return !swiftDriverBuffer.equals(swiftFrontendBuffer);
```
adam-fowler, line 192: Instead of comparing buffers, could you not check the symbolic links point to the same source.

src/utilities/utilities.ts
```diff
@@ -100,13 +100,22 @@ export async function execSwift(
     return await execFile(swift, args, options, folderContext);
 }
 
+/**
+ * Get the file name of executable
+ *
+ * @param exe name of executable to return
+ */
+export function getExecutableName(exe: string): string {
```
adam-fowler, line 108: Is worthwhile combining this with `getSwiftExecutable` and `getSwiftExecutable` always appends the `.exe` to the end?

src/toolchain/toolchain.ts
stevapple: Swift tool binaries are _copied_ instead of _linked_ on Windows most of the time, so comparing the file itself is the correct way.

src/toolchain/toolchain.ts
stevapple: Yes there is — when `swift` is replaced by the new driver.

src/utilities/utilities.ts
stevapple: I believe it already does that👀

This change belongs to #252 , if you have more questions let’s just move there

adam-fowler: Yeah just realised. Will move comment across

src/toolchain/toolchain.ts
stevapple: I added symlink comparison before buffer. Mark this as resolved

src/commands.ts
```diff
@@ -187,6 +187,10 @@ async function runSingleFile(ctx: WorkspaceContext) {
         return;
     }
 
+    if (process.platform === "win32" && !ctx.toolchain.newSwiftDriver) {
+        return;
```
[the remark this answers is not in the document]
stevapple: When should we display such message?

I’m not quite familiar with VS Code extension lifecycle. If we put it here, when will the user see the message box?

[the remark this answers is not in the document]
adam-fowler: If you put it there the user will see the message box.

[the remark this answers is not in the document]
stevapple: Yah I know but — _when_ will they see it?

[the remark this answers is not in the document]
adam-fowler: When they run the command

stevapple:
Verified on Windows:)

adam-fowler (approved):
LGTM. 
Not overly keen on the `checkNewDriver` function which tests absolutely everything but not sure we can get around it.
Have you test legacy Windows display message as expected?

stevapple:
Yes, test on both drivers.

adam-fowler:
When did swift-driver appear? I'm seeing it in swift 5.5 I thought it came in with 5.6. I'm just trying to work out why we are going through all this to verify that the new driver is being used, instead of just testing a version?

stevapple:
I think it comes with Swift 5.5 on Linux/macOS, and will come in 5.7 on Windows. But user can always disable it or remove it, and Swift will still work with legacy driver.

stevapple:
Another reason to not judge by version is that there may be toolchain developers using their own toolchain, and Swift Driver is optional.

adam-fowler:
ok

compnerd:
You should be able to check the version - Windows only supports swift-driver with >=5.7 and only supports the C++ driver on <5.7.

PR #255, opened 2022-04-17
Add custom toolchain to `Path` on Windows

stevapple:
RT.

Notice that this is not enough for supporting an alternative toolchain on Windows. The next step is to support custom `SDKROOT` and append it into `Path` too.

### DIFF
--- a/src/utilities/utilities.ts
+++ b/src/utilities/utilities.ts
@@ -97,6 +97,15 @@ export async function execSwift(
     folderContext?: FolderContext
 ): Promise<{ stdout: string; stderr: string }> {
     const swift = getSwiftExecutable();
+
+    // add custom toolchain to path on Windows
+    if (process.platform === "win32" && configuration.path) {
+        options.env = {
+            Path: `${configuration.path};${process.env.Path}`,
+            ...options.env,
+        };
+    }
+
     return await execFile(swift, args, options, folderContext);
 }
 


adam-fowler:
Is there not a `-sdk` parameter we could use and add to the `win32BuildOptions` in `SwiftTaskProvider.ts`?

stevapple:
> Is there not a `-sdk` parameter we could use and add to the `win32BuildOptions` in `SwiftTaskProvider.ts`?

No. Windows driver today will always require `%SDKROOT%`. Besides, we also need to append the SDK to `%Path%` or Swift Driver won’t work itself. That’s tricky because Windows doesn’t have `rpath` support and all dynamic libraries need to be in `%Path%`.

adam-fowler:
I was sure @compnerd mentioned a `-sdk` command line parameter.

stevapple:
I think you may be misunderstanding what `SDKROOT` is. `SDKROOT` is the default SDK to fall back to — which is, most usually, the SDK of the host system. `-sdk` should only be used to control the target SDK with cross compilation if possible.

stevapple:
There is some problem with the Windows Swift toolchain today. SwiftPM and Swift Driver depends on DLLs from the SDK, but they’re not copied over to the toolchain. Once all toolchain-required DLLs are available alongside, we can remove SDKROOT injections and rely on `-sdk` only.

src/utilities/utilities.ts
```diff
@@ -97,6 +97,15 @@ export async function execSwift(
     folderContext?: FolderContext
 ): Promise<{ stdout: string; stderr: string }> {
     const swift = getSwiftExecutable();
+
+    // add custom toolchain to path on Windows
+    if (process.platform === "win32" && configuration.path) {
+        options.env = {
+            Path: `${configuration.path};${process.env.Path}`,
+            ...options.env,
```
adam-fowler, line 105: If `options.env` already has a `Path` value in it, it will overwrite your new `Path` value.

adam-fowler:
> There is some problem with the Windows Swift toolchain today. SwiftPM and Swift Driver depends on DLLs from the SDK, but they’re not copied over to the toolchain. Once all toolchain-required DLLs are available alongside, we can remove SDKROOT injections and rely on `-sdk` only.

Ok fair enough

compnerd:
> There is some problem with the Windows Swift toolchain today. SwiftPM and Swift Driver depends on DLLs from the SDK, but they’re not copied over to the toolchain. Once all toolchain-required DLLs are available alongside, we can remove SDKROOT injections and rely on `-sdk` only.

This is not a bug.  This is a feature and by design.  The toolchain is dependent on the runtime components and they are not to be copied into the toolchain.  This is one of the problems with depending on yourself.  The toolchain doesn’t depend on the SDK.

> EDIT: rereading your comment it seems to imply that -sdk is useless if it requires the path to be set as well.

There are three pieces at play here.
1. Toolchain - the compilers, assembler, linker, and debugger.
2. SDK - the developer only content that is required for compiling for the intended host
3. Runtime - the pieces that are meant for redistribution to users of the software.

Each item is different and packaged separately.  As issues get resolved, everything will fall into place for parallel installations.

stevapple:
@compnerd Given that runtime is separated from SDK, I think this makes sense👀 Then what we should do is to provide a config for runtime path.

I’m going to close this PR in favor of a more precise way to deal with runtime injection. See later.

compnerd:
Something to consider - but I'm not familiar enough with VSCode to know if this is possible (even if buggy due to my mistakes/lack of knowledge): can we use WMIC or C++?  That would allow us to actually enumerate installed packages, find the runtime MSI (yes, the installer installs multiple products), get the install location and pre-populate the value.  In the future, multiple versions being permitted to be co-installed would mean that you can select the version you want (this is more for application launching than the toolchain).  Now, even with this, it would be nice to have the user specified location for a locally built, uninstalled version to be used.

adam-fowler:
> can we use WMIC or C++?

Don't know much about WMIC but the first MS webpage I found out about it said it was deprecated

You could write a small C++ executable and spawn it from typescript. For example the LLDB extension is mainly written in Rust with a Typescript wrapper. But if it was just to pre-populate a single setting that is platform specific it is probably not worth the effort.

PR #240, opened 2022-04-12
Extension Settings doc.md

adam-fowler:
Document giving more detail what each setting does

PR #256, opened 2022-04-17
Support custom SDK

stevapple:
- More precise output for SDKs;
- Support non-standardized SDK layout on Windows;
- Support custom SDK through configuration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -189,6 +189,17 @@
             "order": 4
           }
         }
+      },
+      {
+        "title": "Advanced",
+        "properties": {
+          "swift.SDK": {
+            "type": "string",
+            "default": "",
+            "description": "The path of the SDK to compile against. The default SDK is determined by the environment on macOS and Windows.",
+            "order": 1
+          }
+        }
       }
     ],
     "menus": {
--- a/src/SwiftPackage.ts
+++ b/src/SwiftPackage.ts
@@ -14,7 +14,7 @@
 
 import * as vscode from "vscode";
 import * as fs from "fs/promises";
-import { execSwift, swiftpmSDKFlags } from "./utilities/utilities";
+import { execSwift } from "./utilities/utilities";
 
 /** Swift Package Manager contents */
 export interface PackageContents {
@@ -151,10 +151,11 @@ export class SwiftPackage implements PackageContents {
     static async loadPackage(folder: vscode.Uri): Promise<PackageContents | null | undefined> {
         try {
             let { stdout } = await execSwift(
-                ["package", "describe", ...swiftpmSDKFlags(), "--type", "json"],
+                ["package", "describe", "--type", "json"],
                 {
                     cwd: folder.fsPath,
-                }
+                },
+                true
             );
             // remove lines from `swift package describe` until we find a "{"
             while (!stdout.startsWith("{")) {
--- a/src/SwiftPackage.ts
+++ b/src/SwiftPackage.ts
@@ -152,7 +152,7 @@ export class SwiftPackage implements PackageContents {
     static async loadPackage(folder: vscode.Uri): Promise<PackageContents | null | undefined> {
         try {
             let { stdout } = await execSwift(
-                ["package", "describe", "--type", "json", ...swiftpmSDKFlags()],
+                ["package", "describe", ...swiftpmSDKFlags(), "--type", "json"],
                 {
                     cwd: folder.fsPath,
                 }
--- a/src/SwiftPackage.ts
+++ b/src/SwiftPackage.ts
@@ -14,8 +14,7 @@
 
 import * as vscode from "vscode";
 import * as fs from "fs/promises";
-import { execSwift } from "./utilities/utilities";
-import { swiftpmSDKFlags } from "./SwiftTaskProvider";
+import { execSwift, swiftpmSDKFlags } from "./utilities/utilities";
 
 /** Swift Package Manager contents */
 export interface PackageContents {
--- a/src/SwiftPackage.ts
+++ b/src/SwiftPackage.ts
@@ -15,6 +15,7 @@
 import * as vscode from "vscode";
 import * as fs from "fs/promises";
 import { execSwift } from "./utilities/utilities";
+import { swiftpmSDKFlags } from "./SwiftTaskProvider";
 
 /** Swift Package Manager contents */
 export interface PackageContents {
@@ -150,9 +151,12 @@ export class SwiftPackage implements PackageContents {
      */
     static async loadPackage(folder: vscode.Uri): Promise<PackageContents | null | undefined> {
         try {
-            let { stdout } = await execSwift(["package", "describe", "--type", "json"], {
-                cwd: folder.fsPath,
-            });
+            let { stdout } = await execSwift(
+                ["package", "describe", "--type", "json", ...swiftpmSDKFlags()],
+                {
+                    cwd: folder.fsPath,
+                }
+            );
             // remove lines from `swift package describe` until we find a "{"
             while (!stdout.startsWith("{")) {
                 const firstNewLine = stdout.indexOf("\n");
--- a/src/SwiftTaskProvider.ts
+++ b/src/SwiftTaskProvider.ts
@@ -17,7 +17,7 @@ import { WorkspaceContext } from "./WorkspaceContext";
 import { FolderContext } from "./FolderContext";
 import { Product } from "./SwiftPackage";
 import configuration from "./configuration";
-import { getSwiftExecutable, swiftpmSDKFlags } from "./utilities/utilities";
+import { getSwiftExecutable, withSwiftSDKFlags } from "./utilities/utilities";
 import { Version } from "./utilities/version";
 
 /**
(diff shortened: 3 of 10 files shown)


adam-fowler:
Is this Windows specific?

stevapple:
No, of course. The concept of SDK works on every platform and is, IMO, more useful on macOS actually.

This will also prepare for cross-compilation support which is directly depended on standalone SDKs.

adam-fowler:
Ok. I think this needs to go into an advanced section of the settings (along with the custom runtime settings), so should be tagged `swift.advanced.SDK`. I already have had people adding bugs where they are putting random values into the sourcekit-lsp options.

src/SwiftPackage.ts
```diff
@@ -150,9 +151,12 @@ export class SwiftPackage implements PackageContents {
      */
     static async loadPackage(folder: vscode.Uri): Promise<PackageContents | null | undefined> {
         try {
-            let { stdout } = await execSwift(["package", "describe", "--type", "json"], {
-                cwd: folder.fsPath,
-            });
+            let { stdout } = await execSwift(
+                ["package", "describe", "--type", "json", ...swiftpmSDKFlags()],
```
adam-fowler, line 155: If the sdk flags need added to every call of `swift` why not add them in the `execSwift` instead of at every call site.

src/SwiftTaskProvider.ts
```diff
@@ -60,22 +60,39 @@ function testDiscoveryFlag(ctx: FolderContext): string[] {
...
+    }
+    return [];
 }
 
 /**
  * Creates a {@link vscode.Task Task} to build all targets in this package.
  */
 export function createBuildAllTask(folderContext: FolderContext): vscode.Task {
-    const additionalArgs: string[] = [];
+    const additionalArgs = [...swiftpmSDKFlags(), ...platformDebugBuildOptions()];
```
adam-fowler, line 92: As with `swiftExec` is it worth adding the sdk flags inside `createSwiftTask` instead of at every call site

package.json
```diff
@@ -152,6 +152,12 @@
             "description": "Output additional diagnostics to the Swift Output View.",
             "order": 8
           },
+          "swift.SDK": {
+            "type": "string",
+            "default": "",
+            "markdownDescription": "The path of the preferred Swift SDK.  The default is:\n- the result of `xcrun --sdk macosx --show-sdk-path` on macOS;\n- the value of `%SDKROOT%` on Windows.",
```
adam-fowler, line 158:
```suggestion
            "markdownDescription": "The path of the Swift SDK.  The default is to use the SDK that came packaged with your swift installation",
```

package.json
```diff
@@ -152,6 +152,12 @@
             "description": "Output additional diagnostics to the Swift Output View.",
             "order": 8
           },
+          "swift.SDK": {
```
adam-fowler, line 155:
```suggestion
          "swift.advanced.SDK": {
```
This is an advanced feature that most users of the extension will not use so can we flag it as such. "advanced" flags will get grouped together.

src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -231,13 +232,14 @@ export class LanguageClientManager {
         const serverPathConfig = lspConfig.serverPath;
         const serverPath =
             serverPathConfig.length > 0 ? serverPathConfig : getSwiftExecutable("sourcekit-lsp");
+        const sdkArguments = swiftDriverSDKFlags(true);
```
adam-fowler, line 235: How come the sdk flag needs passed via `-Xswiftc` here?

src/toolchain/toolchain.ts
```diff
@@ -106,7 +132,14 @@ export class SwiftToolchain {
                     return undefined;
                 }
                 const platformPath = path.dirname(path.dirname(path.dirname(sdkroot)));
-                const data = await fs.readFile(path.join(platformPath, "Info.plist"), "utf8");
+                const platformManifest = path.join(platformPath, "Info.plist");
+                if ((await pathExists(platformManifest)) !== true) {
+                    await vscode.window.showWarningMessage(
+                        "XCTest not found due to irregular library layout."
```
adam-fowler: Can we add something about the test explorer not working because of this

src/SwiftPackage.ts
stevapple: SwiftPM and Swift Driver use different styles of SDK flag, so we need to distinguish them from each other. Also, there’re some simple commands (like `swift package clean` and `swift --version`) that doesn’t depend on SDK.

In the long run, we should distinguish host SDK and target SDK, which should be passed accordingly to tasks.

src/SwiftTaskProvider.ts
stevapple: The same reason: different Swift tasks requires different SDK flags, so it’s not wise to hardcode.

package.json
stevapple: Oops, macOS SDKs never comes with Swift. It is always distributed standalone, and managed by `xcrun` by default. And on Linux we don’t quite use the concept of SDK, there’s also no "default SDK".

src/sourcekit-lsp/LanguageClientManager.ts
stevapple: SourceKit-LSP doesn’t support SwiftPM-styled SDK flags, so we need to directly pass the flags to `swiftc`.

package.json
stevapple: I guess this is not so useful in solving the problem.

However, it costs too high for us to move a config from Advanced to non-Advanced. It’s also not so good to hard-code a config as advanced.

Can we break the configurations into groups like CodeLLDB does? The grouped configurations can be adjusted at any time without breaking any user settings, and it’s much more obvious than a plain "prefix" like `Swift > Advanced: …`.

src/SwiftPackage.ts
adam-fowler: I just feel there must be a cleaner way to do this. Anyone adding new calls to `swift` will need to remember to add these.

package.json
adam-fowler: We could break the configuration into groups like Code-LLDB. I previously didn't because it required a later version of VSCode. But the next release of vscode-swift will require 1.65.0 so we can move to using groups. Shall we leave that to a later PR which breaks off an advanced section and a Sourcekit-LSP section

stevapple: I think we should do this by 0.5.0 release, so mark as resolved now:)

src/SwiftPackage.ts
stevapple: I tried to do so, but unfortunately should give up because of some bugs in SwiftPM (related to Argument Parser, I guess) that requires certain order of arguments, so we cannot simply append it to the end.

stevapple: I will keep the WIP branch until the issue is addressed.

package.json
adam-fowler: I don't think this is resolved. If you want to add details about what the default is, it cannot be run this command to find out. It needs to be more descriptive. If you want to add more details, I have a settings documentation file in PR #240 which has more detail about each setting. It will need to be extended with the new settings

stevapple: I think this would definitely need extra documentation there. So change the description here to:

> The path of the SDK to compile against. The default SDK is determined by the environment on macOS and Windows.

Okay?

adam-fowler: @stevapple Need to mention Linux where it is packaged with the swift install

stevapple: On Linux no SDK will be set by default.

Setting an SDK also means to make it a sysroot for clang, which may cause troubles on Linux. We’re just including Swift libraries in Linux toolchain, and SDK means much more.

adam-fowler: Yes I understand but the text doesn't mention this. At least say this variable has no effect on Linux

stevapple: You can build a package with C/C++ (eg. `llbuild`) by `swift build -v` on each platform to verify this. On macOS and Windows, `-sdk` is set for `swiftc` and `-isysroot` is set for `clang`; on Linux neither will be set.

adam-fowler: How about

> The path of the SDK to compile against. The default SDK is determined by the environment on macOS and Windows. On Linux this setting will have no effect.

stevapple: Also wrong. On Linux explicitly-set SDK will still be respected.

No default ≠ No support

adam-fowler: Is setting the sdk on linux actually any use?

stevapple: Setting an SDK has two general purposes:
- Point to an SDK where SDK is necessary (macOS/Windows) and when the default SDK cannot be found, eg. when Xcode is missing on macOS;
- For cross-compilation.

On Linux it’s obviously the latter.

src/SwiftPackage.ts
stevapple: I managed to land it in the latest commit👀 Hopefully this won’t trigger the `swift-package` bug.

stevapple:
> #261 has been merged. Do you want to move the SDK settings into a separate "Advanced" configuration group

Done.

adam-fowler (approved):
Looking good now. 
One minor nit, can you just add some comments to the `withSwiftSDKFlags` function detailing the issues with placement of the `-sdk` parameter. I think you mentioned somewhere there was a related swift bug. If you could include the link to that in your comments as well.

stevapple:
The comment is added.

> I think you mentioned somewhere there was a related swift bug. If you could include the link to that in your comments as well.

We’re now migrating from JIRA to GitHub issues, so there’s no way to submit bugs now. I will do that when issues are open for SwiftPM repo.

adam-fowler:
> We’re now migrating from JIRA to GitHub issues, so there’s no way to submit bugs now. I will do that when issues are open for SwiftPM repo.

Oh yeah forgot about that.

PR #258, opened 2022-04-18
Support custom runtime with injection

stevapple:
- Support custom runtime library path through configuration;
- Inject custom runtime path whenever needed;
- Switch to `swift -print-target-info` for compiler information;
- Move `swift.diagnostics` configuration to "Advanced".


stevapple:
I have verified this works with Windows. @adam-fowler Can you test for Linux / Darwin? This should have minor impact on non-Windows platform, but I need to ensure that it works.

adam-fowler:
Is there any advantage to having this running on non-Windows platforms?

stevapple:
…Hard to say actually.

It has no effect on the compiler, and is only necessary for platforms without RPath. In most cases, RPath will find the correct library on macOS and Linux. Is there any platform you can remember that doesn’t support RPath?

src/debugger/launch.ts
```diff
@@ -88,6 +89,7 @@ function createExecutableConfigurations(ctx: FolderContext): vscode.DebugConfigu
                 args: [],
                 cwd: folder,
                 preLaunchTask: `swift: Build Debug ${product.name}${nameSuffix}`,
+                env: swiftRuntimePathEnv({}),
```
adam-fowler, line 92: Any reason you are passing `{}` instead of `process.env`?

src/debugger/launch.ts
```diff
@@ -128,7 +131,11 @@ export function createTestConfiguration(
         folder = `${workspaceFolder}}/${ctx.relativePath}`;
         nameSuffix = ` (${ctx.relativePath})`;
     }
-    const testEnv = configuration.testEnvironmentVariables;
+    // respect user configuration if conflicts with injected runtime path
+    const testEnv = {
+        ...swiftRuntimePathEnv(),
+        ...configuration.testEnvironmentVariables,
```
adam-fowler, line 137: Because `configuration.testEnvironmentVariables` is last its variables will take precedence in any variable name clashes. Do you want that?

src/debugger/launch.ts
```diff
@@ -161,8 +168,8 @@ export function createTestConfiguration(
             program: `${folder}/.build/debug/${ctx.swiftPackage.name}PackageTests.xctest`,
             cwd: folder,
             env: {
-                Path: `${xcTestPath};${process.env.Path}`,
                 ...testEnv,
+                Path: `${xcTestPath};${testEnv.Path}`,
```
adam-fowler, line 172: By doing this you are assuming `testEnv` has a path variable. On macOS and Linux it doesn't.

src/toolchain/toolchain.ts
```diff
@@ -54,6 +76,12 @@ export class SwiftToolchain {
 
     logDiagnostics(channel: SwiftOutputChannel) {
         channel.logDiagnostic(`Toolchain Path: ${this.toolchainPath}`);
+        if (this.defaultTarget) {
+            channel.logDiagnostic(`Default Target Triple: ${this.defaultTarget}`);
```
adam-fowler, line 80: It is not completely obvious what "Default Target Triple" is

package.json
```diff
@@ -96,14 +96,20 @@
...
+          },
           "swift.buildArguments": {
             "type": "array",
             "default": [],
             "items": {
               "type": "string"
             },
             "markdownDescription": "Arguments to pass to `swift build`. Keys and values should be provided as separate entries. If you have created a copy of the build task in `tasks.json` then these build arguments will not be propogated to that task.",
-            "order": 2
+            "order": 11
```
adam-fowler, line 112: You can probably undo all the order setting as this will go into the advanced settings subsection as discussed in the sdk PR

package.json
```diff
@@ -96,14 +96,20 @@
             "markdownDescription": "The path of the folder containing the Swift executables. The default is to look in **$PATH**.",
             "order": 1
           },
+          "swift.runtimePath": {
+            "type": "string",
+            "default": "",
+            "description": "The path of the folder containing the Swift runtime libraries. This will only affect runtime behavior when libraries are not discovered by RPath.",
```
adam-fowler, line 102: Wording of this can be improved. Should we use something similar to the sdk description

src/utilities/utilities.ts
```diff
@@ -39,6 +39,9 @@ export async function execFile(
         `Exec: ${executable} ${args.join(" ")}`,
         folderContext.name
     );
+    if (configuration.runtimePath.length > 0) {
+        options.env = { ...options.env, ...swiftRuntimePathEnv(options.env) };
+    }
```
adam-fowler, line 44: `options.env` may be undefined here, if that is the case you should use `process.env` as the basis for runtime path construction

src/utilities/utilities.ts
```diff
@@ -62,6 +65,9 @@ export async function execFileStreamOutput(
...
```
adam-fowler, line 70: `options.env` may be undefined here, if that is the case you should use `process.env` as the basis for runtime path construction

adam-fowler (commented):
In general this is ok
There are a number of cases where environment variables are getting lost
Plus documentation needs to be a little clearer

src/debugger/launch.ts
stevapple: Since this is going to be persisted in `launch.json`, I tend to have the minimal list of paths. Regular Swift executables won’t depend on anything but the Swift runtime (Windows stuffs are always available regardless of `%Path%`). If the user need more, he can change the config manually or through "Test Environment Variables".

src/debugger/launch.ts
stevapple: Yes, it is left intentionally. We should always allow the user to overwrite the path if they wish.

src/utilities/utilities.ts
stevapple: If `options.env` is undefined, `swiftRuntimePathEnv()` will use the default value, which is `process.env`.

src/utilities/utilities.ts
stevapple: The same as last one.

src/debugger/launch.ts
stevapple: Isn’t this nested in `if (process.platform == "win32")`?

src/toolchain/toolchain.ts
stevapple: Well, it’s just for logging now — this is what `swift --version` outputs for the second line, indicating the default build target the driver will use. We can possibly eliminate "triple" here if you think that’s better.

Targets are tied to SDKs, so it’s actually more related to #256. I put it here because it uses `-print-target-info` output.

adam-fowler:
Just to verify, this is runtime that should be available when running the built application, not when building? Just asking because you are setting it up for both execFile and launch configurations.

stevapple:
It is used to support the toolchain too.

adam-fowler:
> It is used to support the toolchain too.

If it is used in the toolchain shouldn't you be passing it to the build tasks as well.

Instead of trying to catch every process run by the extension it might be more simple to just catch when the runtime path setting is changed and use
```ts
process.env.Path = `runtime path:${process.env.Path}`
```
You might need to clean up the path to remove old settings, but that shouldn't be too hard

stevapple:
> If it is used in the toolchain shouldn't you be passing it to the build tasks as well.

Build tasks are created by `createSwiftTask`, which uses `execSwift` and `execFile`, so I think it’s already passed.



> Instead of trying to catch every process run by the extension it might be more simple to just catch when the runtime path setting is changed and use
> 
> ```ts
> process.env.Path = `runtime path:${process.env.Path}`
> ```
> 
> You might need to clean up the path to remove old settings, but that shouldn't be too hard

Since we directly inject it into `execFile`, which is basically what we use to run any tool, I think it’s acceptable. Modifying the current environment is not so ideal and I think it may unexpectedly affect other things (eg. other extension?), compared to passing them on subprocesses.

src/toolchain/toolchain.ts
adam-fowler: Don't think triple means much to many people

adam-fowler:
> Build tasks are created by `createSwiftTask`, which uses `execSwift` and `execFile`, so I think it’s already passed.

Tasks do not use `execSwift`. Tasks execute shell processes via the `ShellExecution` object see `createSwiftTask`.

So far we have only really identified a Windows use for this variable, we can ensure the additional path is only available while the swift extension is active, and it looks as if that additional path will have minimal affect to other processes I think we can go for the simpler setting of the environment variable globally while the swift extension is running.

src/debugger/launch.ts
stevapple: I don’t know if this is the best way, but the reverse should be absolutely wrong👀 Users cannot change runtime path settings easily because it is the milestone of all Swift tools, so if it has the highest priority the users will lose the control to test environment.

adam-fowler: ok, that's cool

stevapple:
> Tasks do not use `execSwift`. Tasks execute shell processes via the `ShellExecution` object see `createSwiftTask`.

I see. Fixed now.

src/toolchain/toolchain.ts
stevapple: It’s useless for most people — but it shouldn’t be visible to most people too. The diagnostic output is disabled by default.

adam-fowler:
Ok, this is better. There is still one issue: generation of the `launch.json` occurs at extension initialisation and is only generated if it doesn't exist, or the `program`, `cwd` or `preLaunchTask` sections are different. It doesn't re-generate the launch configs if the `env` is different as the user can use the `env` section to add their own environment variables. 

So if you edit the runtime path you need to catch it has been edited and re-generate the launch.json, while not trashing any environment the user may have setup in there. 

Also you probably want to only re-generate once the user has finished editing the path. I got around this when editing the swift path (which requires a reload of the extension) by displaying a dialog saying do you want to restart and when they press ok restart. https://github.com/swift-server/vscode-swift/blob/4e27f8e885a54424fb890e22924893befcd20863/src/WorkspaceContext.ts#L50-L55

stevapple:
With the final commit, now we should fully support unpackaged Swift toolchain/SDK/runtime on Windows.

Personally I've in favor of using Registry to substitute for strict library layout conventions, but this still needs to be discussed and is not implemented yet.

adam-fowler:
On macOS with runtime path set to `/usr/path/`

Debugging application DYLD_LIBRARY_PATH = `/usr/path/:`
When running tests DYLD_LIBRARY_PATH = `/usr/path/:/usr/path/:undefined`
When debugging tests DYLD_LIBRARY_PATH = nil

None of them are set to `/usr/path/` as you would expect

src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -248,7 +254,7 @@ export class LanguageClientManager {
             if (toolchainPath) {
                 // eslint-disable-next-line @typescript-eslint/naming-convention
                 sourcekit.options = {
-                    env: { ...process.env, SOURCEKIT_TOOLCHAIN_PATH: toolchainPath },
+                    env: { ...sourcekit.options?.env, SOURCEKIT_TOOLCHAIN_PATH: toolchainPath },
```
adam-fowler, line 257: Is the `?` after `options` required? You just set `source kit.options` further up

stevapple: We should use either `!` or `?` here, and they should work the same because it's guaranteed to be created (but cannot be recognized by TS).

src/debugger/launch.ts
```diff
@@ -35,6 +37,11 @@ export async function makeDebugConfigurations(ctx: FolderContext) {
     for (const config of configs) {
         const index = launchConfigs.findIndex(c => c.name === config.name);
         if (index !== -1) {
+            if (regenerate && launchConfigs[index] !== config) {
+                launchConfigs[index] = config;
```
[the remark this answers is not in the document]
stevapple: It's hard to actually prevent us from covering user edits because they're also free to edit the paths themselves... And the logic/consequences can be hard to describe.

To regenerate here makes the logic general and applicable to other cases that may change the base config. I intended to notify user about the risk of their config being covered by using "Regenerate" instead of "Update" here.

package.json
```diff
@@ -193,11 +187,23 @@
...
             "default": "",
             "description": "The path of the SDK to compile against. The default SDK is determined by the environment on macOS and Windows.",
-            "order": 1
+            "order": 2
+          },
+          "swift.diagnostics": {
+            "type": "boolean",
+            "default": false,
+            "description": "Output additional diagnostics to the Swift Output View.",
+            "order": 100
```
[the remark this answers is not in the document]
stevapple: I marked this as explicitly the last one👀 Newer `swift` configs should be placed before `swift.diagnostics`.

adam-fowler:
> All fixed

I'm still getting
Debugging application DYLD_LIBRARY_PATH = /usr/path/:

Not sure how you resolve this one though

src/debugger/launch.ts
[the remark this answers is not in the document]
stevapple: I find this hard to implement because it's hard to update nested value in TypeScript🤷‍♂️definitely missing key path in Swift.

stevapple:
@adam-fowler It's up to you whether to accept the latest commit... That's really a ugly one.

stevapple:
> I guess the original value of your `DYLD_LIBRARY_PATH` is retrieved as `""` instead of `undefined`. Will explicitly handle this.

I think I misjudged here. You’re getting `/usr/path/:` because we set `DYLD_LIBRARY_PATH=/usr/path/:${env:DYLD_LIBRARY_PATH}` but you don’t have `DYLD_LIBRARY_PATH` originally set. This is almost impossible to fix because we can’t assume the environment and thus have this line in `launch.json`.

src/debugger/launch.ts
```diff
@@ -35,6 +40,29 @@ export async function makeDebugConfigurations(ctx: FolderContext) {
     for (const config of configs) {
         const index = launchConfigs.findIndex(c => c.name === config.name);
         if (index !== -1) {
+            if (keysToUpdate && launchConfigs[index] !== config) {
```
adam-fowler, line 43: Can we break this update code into a separate function in utilities.ts. It seems quite useful

adam-fowler: Oh plus maybe add some comments detailing what it is doing

src/TestExplorer/TestRunner.ts
```diff
@@ -233,9 +233,10 @@ export class TestRunner {
                 token,
                 {
                     cwd: testBuildConfig.cwd,
-                    env: { ...process.env, ...testBuildConfig.env },
+                    env: { ...testBuildConfig.env, ...process.env },
```
adam-fowler, line 236: Why have you swapped these two around. This now means variables in `process.env` override the variables from `testBuildConfig.env`

src/utilities/utilities.ts
```diff
@@ -33,12 +68,16 @@ export async function execFile(
...
     options: cp.ExecFileOptions = {},
-    folderContext?: FolderContext
+    folderContext?: FolderContext,
+    customSwiftRuntime = true
 ): Promise<{ stdout: string; stderr: string }> {
     folderContext?.workspaceContext.outputChannel.logDiagnostic(
         `Exec: ${executable} ${args.join(" ")}`,
         folderContext.name
     );
+    if (configuration.runtimePath.length > 0 && customSwiftRuntime) {
```
adam-fowler, line 78: Do we need to check `configuration.runtimePath.length` here given the same check is in `swiftRuntimeEnv`

src/utilities/utilities.ts
```diff
@@ -56,12 +95,16 @@ export async function execFileStreamOutput(
...
```
adam-fowler, line 105: Do we need to check `configuration.runtimePath.length` here given the same check is in `swiftRuntimeEnv`

adam-fowler (commented):
A few more comments, but we are pretty much there once we break out the launch config into a more general function in utilities.ts

src/utilities/utilities.ts
stevapple: This will prevent us from initializing `"env": {}`

adam-fowler: ok fair enough

src/debugger/launch.ts
stevapple: 🤔This doesn’t appear as easy as it looks like… We’re dancing at the edge of TypeScript’s type system here, and I tried many possible generics but still cannot comfort the type checker.

stevapple: This means we can only specify single type (`vscode.DebugConfiguraion`) to use it, which I think is not so general to go into `utilities`. Just extract to a fileprivate function in `launch.ts` instead?

adam-fowler: Sounds good

adam-fowler (approved):
I found another issue. When I change the runtime path it only changes the launch.json for currently opened projects. If you load another project it won't edit its launch.json. Otherwise everything looks good.

I've got a local fix for this where the environment variable merge is done on every call to `makeDebugConfigurations`. I'll add a PR for this after merging this

Issue #260, opened 2022-04-18
Add LLDB init commands for running on Windows


adam-fowler:
It looks like Swift LLDB on Windows requires the swiftCore.dll to be loaded before running the debugger. The following lldb command does this
```
process load <path to swiftCore.dll>
```
This can be added in the `initCommands` section to the launch configs built in `createExecutableConfigurations` and `createTestConfiguration` or by setting the `initCommands` field in the LLDB configuration settings.

compnerd:
Correction to this - it was the missing SDK settings.  So this should be `settings set target.sdk <SDKROOT>`.

adam-fowler:
Resolved in #275

PR #262, opened 2022-04-19
Windows: support launching lldb on Windows

compnerd:
Although the Swift release of lldb does not currently function
(c.f. #260), the default debugger does function and would enable at
least partial debugging which is useful.


adam-fowler:
Is it worthwhile doing the work for #260 in this PR seeing as without it the debugger will be completely broken for Windows?

compnerd:
I don't think that it is completely broken.  You cannot use the lldb bundled in the Swift toolchain, but can use lldb to a certain extent.

adam-fowler:
This change is enabling the swift lldb so would break debugging on Windows, well until at least the work from #260 was done

adam-fowler (approved):

PR #266, opened 2022-04-21
Support custom host SDK (aka. `SDKROOT`)

stevapple:
This PR differentiates host SDK and target SDK, and adds support for setting custom host SDK through "Advanced" configuration.


stevapple:
@adam-fowler Are you able to reproduce the CI error? I didn't think this has touched the behavior if no custom SDK is set, and cannot reproduce in my environments (Windows/5.6.1 & macOS/5.6) either.

adam-fowler:
I can't see why this should fail during CI. Running tests locally works fine. Can you merge in latest changes as it is very hard to see which changes are specific to this PR.

stevapple:
> At the same time a couple of the swift package commands (`describe`, `unedit`) have had the runtime and compiler environment removed from them, as it is unnecessary. Any reason you didn't do this for `edit`.

Could you point out where I may be missing? It feels like any `swift-package` command won’t be affected by custom SDK flags.

I intentionally want to make:
- SDK (and the upcoming "Target") applied to Build Tasks and Run Script Task (and possible debugging on Windows)
- host SDK applied to all tasks except for debugging
- runtime path applied everywhere

adam-fowler:
> `SwiftPM needs to run the manifest on the host, that’s where host SDK should be explicitly used.

Which commands are we talking about here?
Also any reason we don't use `--sdk` to point to the host SDK, in this situation?

stevapple:
Considering compiling a Swift package on macOS without Xcode (wish this could happen, but just imagine) for iOS.

We need to explicitly set the host SDK, or the manifest won’t compile. And we also need to set the target SDK, or the target cannot be supported. So finally it becomes:
```
SDKROOT=/path/to/macos/sdk swift build --triple arm64-apple-ios --sdk /path/to/ios/sdk
```

stevapple:
> Also any reason we don't use `--sdk` to point to the host SDK, in this situation?

Host SDK is never decided by CLI options. There’s only two sources: `SDKROOT` value > `xcrun` output

adam-fowler:
So what are the reasons you might want to make the "host" sdk configurable? If it is just to compile `Package.swift` I can't see what that would be.

stevapple:
If you don’t need to cross-compile, you’re recommended to set host SDK instead of custom SDK. The latter should be used explicitly for cross-compilation.

Also, setting `SDKROOT` is required for using an alternative Windows toolchain now. There is no second way.

adam-fowler:
> If you don’t need to cross-compile, you’re recommended to set host SDK instead of custom SDK. The latter should be used explicitly for cross-compilation.
> 
The "host" SDK is explicitly set already (when you install Xcode, and as I understand it SDKROOT is set when installing Swift on Windows, I may be wrong there).

> Also, setting `SDKROOT` is required for using an alternative Windows toolchain now. There is no second way.

So does that mean you only use `--sdk` when doing cross compilation.

adam-fowler:
So on macOS what do we gain from setting SDKROOT when we can already set the path to the swift compiler? The only thing I can think of is xctest.

stevapple:
> So on macOS what do we gain from setting SDKROOT when we can already set the path to the swift compiler?

On macOS toolchain and SDK is not tied so closely. Using alternative toolchain doesn’t affect the SDK (thanks to ABI stability). You always need to set SDKROOT if you want to use an SDK outside your current Xcode installation.

stevapple:
> The only thing I can think of is xctest.

XCTest on macOS is part of the toolchain. Because it is found using `xcode-select`, we have no option to override it yet.

adam-fowler:
> You always need to set SDKROOT if you want to use an SDK outside your current Xcode installation.

But can't you use `--sdk` in that case?

stevapple:
@adam-fowler Any review? Better get this in 0.5.0

stevapple:
BTW, plugins in SwiftPM 5.6+ will also stick to host SDK.

src/toolchain/toolchain.ts
```diff
@@ -198,17 +198,22 @@ export class SwiftToolchain {
                 return path.join(stdout.trimEnd(), "usr", "bin");
             }
             case "win32": {
+                // look up runtime library directory for XCTest alternatively
+                const fallbackPath =
```
adam-fowler, line 202: Can we remove this change it is unrelated.

adam-fowler (commented):
In general this looks fine. I'm not sure about the terminology host SDK. I don't remember seeing it anywhere else. Given people who are changing this will know it as SDKROOT call it that.

I'm not including this in 0.5.0 though. As I don't want to rush it in.

src/toolchain/toolchain.ts
stevapple: This is just moved here from L207-209

stevapple:
> I'm not including this in 0.5.0 though. As I don't want to rush it in.

If you're not including this for 0.5.0, I'll split this pitch into two parts (fixing up target SDK & adding host SDK) and left the former merged first. The current behavior of custom target SDK has some mistakes that need to be addressed.

stevapple:
> I'm not sure about the terminology host SDK.

This terminology is derived from "host toolchain" in SwiftPM source code. I don't know where `SDKROOT` is from but it seems to be a more ancient and platform-specific concept and I would tend to keep the SwiftPM naming.

stevapple:
Given that SwiftPM uses the terminology "destination" to describe the target, I'll update `targetSDK` into `destinationSDK` instead.

adam-fowler:
> > I'm not sure about the terminology host SDK.
> 
> This terminology is derived from "host toolchain" in SwiftPM source code. I don't know where `SDKROOT` is from but it seems to be a more ancient and platform-specific concept and I would tend to keep the SwiftPM naming.

host toolchain is a different thing from host sdk

stevapple:
> host toolchain is a different thing from host sdk

There is a clear relationship between destination toolchain — destination SDK — destination target triple. Similarly the logic can be applied to host.

adam-fowler:
Now that #305 is merged, can we close this

PR #268, opened 2022-04-22
Fix finding swiftc in centos7 and amazonlinux2

adam-fowler:
`which` doesn't exist on base Centos7 or AmazonLinux2 so use `type -p`. Also `type` is a builtin on Centos7 so cannot be spawn using execFile so using exec instead.

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -48,7 +48,7 @@ interface SwiftTargetInfo {
 
 export class SwiftToolchain {
     constructor(
-        public toolchainPath: string,
+        public toolchainPath: string | undefined,
         public swiftVersionString: string,
         public swiftVersion: Version,
         public runtimePath?: string,
@@ -100,7 +100,7 @@ export class SwiftToolchain {
     /**
      * @returns path to Toolchain folder
      */
-    private static async getToolchainPath(): Promise<string> {
+    private static async getToolchainPath(): Promise<string | undefined> {
         if (configuration.path !== "") {
             return path.dirname(path.dirname(configuration.path));
         }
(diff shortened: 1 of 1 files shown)


0xTim (commented):
`type -p` seems to print `swiftc is /usr/bin/swiftc` but I don't see any changes in the code that would parse this out?

adam-fowler:
> `type -p` seems to print `swiftc is /usr/bin/swiftc` but I don't see any changes in the code that would parse this out?

type -p on macOS prints `swiftc is /usr/bin/swiftc` but all the linux versions I checked it printed just `/usr/bin/swiftc`

adam-fowler:
@0xTim I just see that jenkins is complaining. Which is odd.

adam-fowler:
Still can't work out the jenkins issue but did work out that `zsh` has a different output so running inside`sh` instead

adam-fowler:
Resolved. We cannot rely on the `-p` argument being available so parsing output instead. Also to ensure output is consistent always run it through `/bin/sh`

0xTim (approved):

src/toolchain/toolchain.ts
```diff
@@ -116,9 +116,17 @@ export class SwiftToolchain {
...
-                const swiftc = stdout.trimEnd();
-                return path.dirname(path.dirname(path.dirname(swiftc)));
+                // use `type swiftc` to find `swiftc`. Run inside /bin/sh to ensure
+                // we get consistent output as different shells output a different
+                // format. Tried running with `-p` but that is not available in /bin/sh
+                const { stdout } = await execFile("/bin/sh", ["-c", "type swiftc"]);
+                const swiftcMatch = /^swiftc is (.*)$/.exec(stdout);
+                if (swiftcMatch) {
+                    const swiftc = swiftcMatch[1];
+                    vscode.window.showInformationMessage("swiftc: ${swiftc}");
```
stevapple, line 126: Why do we need a message here (it's diverged with other platforms)?

Instead, maybe we can add this information to log output?

adam-fowler: damn that's debug info. I was running inside a remote container to test different variants of linux. I'll remove

stevapple: A few more comments:
1. Why was it `swiftc` instead of `swift` here? We never call `swiftc` directly in the plugin.
	In the rare case where the toolchain is hidden from `PATH` (the support is broken and I've recently submitted some pitches), `swift` and `swiftc` should both be exposed.
2. I guess we should also handle cases when `swift` cannot be found on Darwin and Windows.
3. Why not throw an error when `swift` cannot be found? The plugin will always crash then and the users are not informed of the actual reason.

adam-fowler: Not sure why we chose swiftc over swift, don't think its a huge issue. If swift doesn't exist the extension will already fail to load, so no need to do another check here. The main issue here is if `type swiftc` fails, because we are running on some version of linux that hasn't been tested which produces different output from `type`.

adam-fowler: Actually I'll do the throwing of the error, means we don't have an optional toolchain

PR #271, opened 2022-04-23
Merge debug configs always when generating

adam-fowler:
This is to fix the issue where the launch.json is only updated for projects that are open when editing the runtime path is set. 

- Added `launchConfigKeysToUpdate` which defines what parameters of launch config can be updated by new config.
- Clone old config and then merge new config into clone. If the old config and new merged config are not the same then update.
- Added option to not show dialogs

### DIFF
--- a/src/debugger/launch.ts
+++ b/src/debugger/launch.ts
@@ -16,7 +16,7 @@ import * as os from "os";
 import * as vscode from "vscode";
 import configuration from "../configuration";
 import { FolderContext } from "../FolderContext";
-import { swiftRuntimeEnv } from "../utilities/utilities";
+import { swiftLibraryPathVariable, swiftRuntimeEnv } from "../utilities/utilities";
 
 /**
  * Edit launch.json based on contents of Swift Package.
@@ -253,15 +253,7 @@ export function createDarwinTestConfiguration(
 }
 
 function launchConfigKeysToUpdate(): string[] {
-    const keysToUpdate = ["program", "cwd", "preLaunchTask"];
-    switch (process.platform) {
-        case "win32":
-            return [...keysToUpdate, "env.Path"];
-        case "darwin":
-            return [...keysToUpdate, "env.DYLD_LIBRARY_PATH"];
-        default:
-            return [...keysToUpdate, "env.LD_LIBRARY_PATH"];
-    }
+    return ["program", "cwd", "preLaunchTask", `env.${swiftLibraryPathVariable()}`];
 }
 
 /** Return the base configuration with (nested) keys updated with the new one. */
--- a/src/debugger/launch.ts
+++ b/src/debugger/launch.ts
@@ -31,7 +31,8 @@ export async function makeDebugConfigurations(ctx: FolderContext, yes = false) {
     }
     const wsLaunchSection = vscode.workspace.getConfiguration("launch", ctx.folder);
     const launchConfigs = wsLaunchSection.get<vscode.DebugConfiguration[]>("configurations") || [];
-    const keysToUpdate = launchConfigKeysToUpdate();
+    // list of keys that can be updated in config merge
+    const keysToUpdate = ["program", "cwd", "preLaunchTask", `env.${swiftLibraryPathKey()}`];
 
     const configs = createExecutableConfigurations(ctx);
     let edited = false;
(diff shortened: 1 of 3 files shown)


adam-fowler:
@stevapple what are your thoughts?

stevapple:
I learnt that there's no inout parameter in TS... Have you verified that this will work?

adam-fowler:
> I learnt that there's no inout parameter in TS... Have you verified that this will work?

Yes it is verified. There might not be any inout parameters in TS, but every object is a reference object, so if you pass an object into a function and update it inside that function the reference outside the function also gets those changes. That is why I have to do the deep copy before calling `updateConfigWithNewKeys`.

src/debugger/launch.ts
```diff
@@ -254,12 +252,24 @@ export function createDarwinTestConfiguration(
     };
 }
 
+function launchConfigKeysToUpdate(): string[] {
+    const keysToUpdate = ["program", "cwd", "preLaunchTask"];
```
stevapple, line 256: These keys are updated in addition to `env` when runtime path is updated, and the dialog doesn't reflect this.

src/debugger/launch.ts
```diff
@@ -23,42 +23,40 @@ import { swiftRuntimeEnv } from "../utilities/utilities";
...
+            // deep clone config and update with keys from calculated config
+            const newConfig: vscode.DebugConfiguration = JSON.parse(
+                JSON.stringify(launchConfigs[index])
+            );
+            updateConfigWithNewKeys(newConfig, config, keysToUpdate);
+
+            // if original config is different from new config
+            if (JSON.stringify(launchConfigs[index]) !== JSON.stringify(newConfig)) {
+                if (!yes) {
+                    const answer = await vscode.window.showErrorMessage(
```
stevapple, line 50: Given that we're only updating specific keys, I would suggest this to be a warning instead of an error.

src/debugger/launch.ts
adam-fowler: These should only get updated if there is a new version of the extension. If someone changed any of these they would break the launch config. I doubt these will change from this point on, but they have changed between earlier versions of the extension.

src/debugger/launch.ts
adam-fowler: Yeah makes sense

adam-fowler: Done, updated message slightly

src/debugger/launch.ts
stevapple: I mean the message here should also be adjusted (& simplified)👀

https://github.com/swift-server/vscode-swift/blob/2a33ee8c2ac46072d925abf618321cc3f8c4fc38/src/WorkspaceContext.ts#L97

0xTim (approved):

src/debugger/launch.ts
adam-fowler: Given the advance nature of this feature I think the message is ok

stevapple: Can we extract "runtimePathEnvKey" to be an exported function? It seems to be duplicated in multiple functions.

adam-fowler: makes sense

adam-fowler: > Can we extract "runtimePathEnvKey" to be an exported function? It seems to be duplicated in multiple files.

done

src/utilities/utilities.ts
```diff
@@ -47,11 +47,20 @@ export function swiftRuntimeEnv(
     };
     switch (process.platform) {
         case "win32":
-            return runtimeEnv("Path", ";");
+            return runtimeEnv(swiftLibraryPathKey(), ";");
+        default:
+            return runtimeEnv(swiftLibraryPathKey());
+    }
```
stevapple, line 53: We can fully eliminate `runtimeEnv` now:
```typescript
    const runtimePath = configuration.runtimePath;
    const key = swiftLibraryPathKey();
    const separator = process.platform === "win32" ? ";" : ":";
    switch (base) {
        case false:
            return { [key]: runtimePath };
        case true:
            return { [key]: `${runtimePath}${separator}\${env:${key}}` };
        default:
            return base[key]
                ? { [key]: `${runtimePath}${separator}${base[key]}` }
                : { [key]: runtimePath };
    }

File: src/debugger/launch.ts [hunk 4]
```diff
@@ -254,12 +252,16 @@ export function createDarwinTestConfiguration(
     };
 }
 
+function launchConfigKeysToUpdate(): string[] {
+    return ["program", "cwd", "preLaunchTask", `env.${swiftLibraryPathKey()}`];
+}
```
stevapple, line 257: Do we still need this function? At least it's expressive and clear enough to put inline.

PR #273, opened 2022-04-23
Remove unnecessary information dialog, throw error when we cannot find swift

adam-fowler:


src/toolchain/toolchain.ts
```diff
@@ -119,14 +119,13 @@ export class SwiftToolchain {
                 // use `type swiftc` to find `swiftc`. Run inside /bin/sh to ensure
                 // we get consistent output as different shells output a different
                 // format. Tried running with `-p` but that is not available in /bin/sh
-                const { stdout } = await execFile("/bin/sh", ["-c", "type swiftc"]);
-                const swiftcMatch = /^swiftc is (.*)$/.exec(stdout);
-                if (swiftcMatch) {
-                    const swiftc = swiftcMatch[1];
-                    vscode.window.showInformationMessage("swiftc: ${swiftc}");
-                    return path.dirname(path.dirname(path.dirname(swiftc)));
+                const { stdout } = await execFile("/bin/sh", ["-c", "type swift"]);
```
stevapple, line 122: There is actually another situation: `/usr/bin/swift` can be a symlink into `/path/to/toolchain/usr/bin/swift-frontend`. Everything can function well, but the toolchain path is wrong.

src/toolchain/toolchain.ts
```diff
@@ -119,14 +119,13 @@ export class SwiftToolchain {
                 // use `type swiftc` to find `swiftc`. Run inside /bin/sh to ensure
                 // we get consistent output as different shells output a different
                 // format. Tried running with `-p` but that is not available in /bin/sh
```
stevapple, line 121:
```suggestion
                // use `type swift` to find `swift`. Run inside /bin/sh to ensure
                // we get consistent output as different shells output a different
                // format. Tried running with `-p` but that is not available in /bin/sh
```

stevapple (changes requested):
Also needs to handle `swift` missing on Windows and macOS.

src/toolchain/toolchain.ts
adam-fowler: What situation is that?

adam-fowler: anyway fix is in

adam-fowler:

0xTim (approved):

PR #274, opened 2022-04-24
Respect `SWIFT_USE_OLD_DRIVER` environment variable

stevapple:


src/commands.ts
```diff
@@ -188,9 +188,13 @@ async function runSingleFile(ctx: WorkspaceContext) {
...
     }
 
-    // Swift scripts require new swift driver to work on Windows. Swift driver is available
-    // from v5.7 of Windows Swift
-    if (process.platform === "win32" && ctx.toolchain.swiftVersion < new Version(5, 7, 0)) {
+    // Swift interpreter requires new swift driver to work on Windows
+    // which is not available until Swift 5.7
+    if (
+        process.platform === "win32" &&
+        process.env.SWIFT_USE_OLD_DRIVER === undefined &&
```
compnerd, line 195: I don't see the point to this - there is no "old driver", we only have a singular driver on the toolchain - it is either swift-driver (>=5.7) or the C++ driver (<5.7)

stevapple: Users can always opt-out for the new driver, the current Windows toolchain layout is wrong and diverged from other platforms.

`swift` and `swiftc` should always point to `swift-frontend`, and if `swift-driver` is placed alongside it will be called by default. We were not doing this on Windows simply because there's a silly bug in finding `swift-driver`, which is going to be fixed in apple/swift#42619

compnerd: > Users can always opt-out for the new driver, the current Windows toolchain layout is wrong and diverged from other platforms.

No, the windows toolchain layout is perfectly fine and reasonable.  I do agree that it is different from the other platforms.  But the user does **NOT** have the option to opt out of the new driver on windows.  There is no requirement that everything is identical in the toolchain distribution across platforms.  In fact, we already have differences - Foundation and XCTest on macOS are fundamentally different.  Please do not make assumptions on what the toolchain on Windows is supposed to be - ask me.

> `swift` and `swiftc` should always point to `swift-frontend`, and if `swift-driver` is placed alongside it will be called by default. We were not doing this on Windows simply because there's a silly bug in finding `swift-driver`, which is going to be fixed in apple/swift#42619

No, they will **NEVER** do that on Windows.  Installation of symlinks is not stable and not easy.  I do not intend to support that configuration.  Furthermore, that is not a bug, this was something which I was aware of, but opted to ignore because it is not a supported configuration on Windows.

Again please stop assuming that the windows toolchain is supposed to be what you want it to be - the windows toolchain is different from the Linux toolchain and intentionally so.

stevapple: > In fact, we already have differences - Foundation and XCTest on macOS are fundamentally different.

macOS has its historical reason to have part of its toolchain functionality relying on Objective-C runtime. It is not common and not ideal. The community — mainly SSWG folks — are always trying to minor the gap of macOS toolchain and others.

> There is no requirement that everything is identical in the toolchain distribution across platforms.

To provide unified cross-platform developer experience, the toolchain’s interface and behavior should be as close across platforms as possible. I don’t see the point of introducing this difference.

> Installation of symlinks is not stable and not easy.

`swift.exe` and `swiftc.exe` are not required to be symlinks. They were copies of `swift-frontend.exe` and can remain in the future until we finally remove the legacy driver. I don’t think this will introduce any problem.

> Furthermore, that is not a bug, this was something which I was aware of, but opted to ignore because it is not a supported configuration on Windows.

Please **explain** why you think this shouldn’t be supported on Windows. Please don’t make assumptions on UX. Not everyone has the same need as you, and those who’re using advanced or deprecated features always know what they’re doing.

> Again please stop assuming that the windows toolchain is supposed to be what you want it to be

Please post on the forums in detail in what points is Windows toolchain different from others, and please be open to discussions and suggestions. Unix people don’t quite know about Windows toolchain, nor do they know you’re breaking the UX consistency they’ve built.

I don’t think everyone is as satisfied by it as you are.

compnerd: > > In fact, we already have differences - Foundation and XCTest on macOS are fundamentally different.
> 
> macOS has its historical reason to have part of its toolchain functionality relying on Objective-C runtime. It is not common and not ideal. The community — mainly SSWG folks — are always trying to minor the gap of macOS toolchain and others.

Just as Windows has historical differences and those will always continue to come up.  Ultimately, Windows is not Unix, and you cannot expect it to be Unix.

> > There is no requirement that everything is identical in the toolchain distribution across platforms.
> 
> To provide unified cross-platform developer experience, the toolchain’s interface and behavior should be as close across platforms as possible. I don’t see the point of introducing this difference.

We fundamentally disagree here.  It is okay to be different, the differences here are fundamental to the platform.  The arguments are similar for the most part, and behave identically whenever possible.  I don't see Unix providing UNC paths or network shares the same way.  Perhaps that should be introduced on the Unix side too?

> > Installation of symlinks is not stable and not easy.
> 
> `swift.exe` and `swiftc.exe` are not required to be symlinks. They were copies of `swift-frontend.exe` and can remain in the future until we finally remove the legacy driver. I don’t think this will introduce any problem.

That absolutely introduces problems.  It increases the size of the toolchain and the size of the download.  This is a problem.  As it is, the Windows toolchain is ~8G to download if you include the dependencies.  Since you speak so highly of the UX, I consider this part of the UX, and is a problem that I think is more import to resolve than the differences between Unix and Windows.

> > Furthermore, that is not a bug, this was something which I was aware of, but opted to ignore because it is not a supported configuration on Windows.
> 
> Please **explain** why you think this shouldn’t be supported on Windows. Please don’t make assumptions on UX. Not everyone has the same need as you, and those who’re using advanced or deprecated features always know what they’re doing.

Windows has always stated that it will aggressively move towards the newer functionality as feasible.  This is part of that.  This is not an assumption on UX.  This is not making the toolchain conform to my needs (if it were, I would have far more invasive changes that I would make my life easier for the cases that I care about).

> > Again please stop assuming that the windows toolchain is supposed to be what you want it to be
> 
> Please post on the forums in detail in what points is Windows toolchain different from others, and please be open to discussions and suggestions. Unix people don’t quite know about Windows toolchain, nor do they know you’re breaking the UX consistency they’ve built.

I have no intention of spending time on this.  If you want to write up a list of differences, I am happy to look over that and identify if there are pieces which are unintentional.  "Unix people" (though one would argue that I am a Unix person - I help maintain a Linux distribution and have been an active Linux contributor for many years) simply need to learn that Windows is not Unix, and that there are fundamentally differences that will come up when trying to write cross-platform code.

> I don’t think everyone is as satisfied by it as you are.

That is fine - they are welcome to help improve the Windows platform.  However, ultimately, I have spent an inordinate amount of personal time, money, and resources to bring the Windows support to this point, this is not paid work.  I have finite resources, and I need to prioritise the overall Windows support.  I have yet to see anyone willing to help with the core fundamental issues that exist across the various pieces (e.g. LLVM, LLDB, etc.).  Keeping the interfaces narrow is a consequence of that.  The UX here is unperturbed for the most part - the new driver is meant to be a replacement.  If there are features that are missing there, those should be addressed at the appropriate level.

There is effectively <1 engineer working for the Windows platform, but many engineers working on Linux support even.  There is always going to be a disparity in the platforms.  The way to help with that is to encourage more people to help with the fundamental compiler and low level runtime and library issues and to encourage corporate baking to help improve the Windows toolchain.  There are man competing priorities, and as the platform (port) owner, I have to make tradeoffs to support the platform as a whole.

stevapple: > It increases the size of the toolchain and the size of the download. This is a problem. As it is, the Windows toolchain is ~8G to download if you include the dependencies.

It ultimately increases the installed toolchain size by ~500MB, but there will be no download size increase for end-users because the installer can point multiple files to a single source. I don’t regard this as a really big sacrifice — I would say Swift is never doing so good at storage efficiency, and 500MB is even identical to the size of `.build` of a simple Swift package on macOS.

> Windows has always stated that it will aggressively move towards the newer functionality as feasible.

We’re not eliminating the new driver support by introducing "pluggable" `swift-driver.exe`. I don’t think Windows is a fan of breaking any programming interface, including CLI and API (fairly speaking, Apple is far more aggressive there). That is, we always encourage people to use the newest technology, but removing the old piece ahead of other platforms is still questionable.

> That is fine - they are welcome to help improve the Windows platform. However, ultimately, I have spent an inordinate amount of personal time, money, and resources to bring the Windows support to this point, this is not paid work. I have finite resources, and I need to prioritise the overall Windows support.

I wish they are welcomed, but you need to realize that the community (and even core team) is relying on you on **nearly any** Windows pitch in general, and more maintainer will eventually tire you up if the fact doesn’t change.

Discussing pre changes on the forums, pushing required CI testing, these can all help to free yourself to some extent. You and the core team are guiding the right direction, instead of making every decision by yourselves.

> I have yet to see anyone willing to help with the core fundamental issues that exist across the various pieces (e.g. LLVM, LLDB, etc.). Keeping the interfaces narrow is a consequence of that. The UX here is unperturbed for the most part - the new driver is meant to be a replacement.

I wish UX improvements are welcome too — I’m tired of replying to others _"You need to … here, this is the correct way to make it work on Windows"_, but instead I would say _"This is a known issue and will eventually get addressed, you need to … as a workaround"_. Doing correctly doesn’t mean doing well, and doing less (from user side) doesn’t necessarily mean doing wrong. Swift is initially designed to "make impossible tasks possible, and make difficult tasks simple", and I hope so for the Windows toolchain.

What if I regard the driver opt-out to be a feature? Remember that although Swift Driver is stated to be the drop-in replacement of `swift` and `swiftc` at the beginning, now it becomes a pure plugin of `swift-frontend`, and the developers are keeping this in mind. If you make any aggressive change, please mention it in the document or on the forums, or the compatibility will be easily broken by ignorant developers from other platforms.

> There is always going to be a disparity in the platforms. The way to help with that is to encourage more people to help with the fundamental compiler and low level runtime and library issues and to encourage corporate baking to help improve the Windows toolchain. There are man competing priorities, and as the platform (port) owner, I have to make tradeoffs to support the platform as a whole.

Joe Groff mentioned the idea of forming a "Windows workgroup" during DM, but I would argue that the best effort to bring new platform support is to let each other know what you’re doing and what you want. That’s why I pre-proposed the idea of "cross-compilation workgroup" or "toolchain support workgroup" in https://forums.swift.org/t/misplacement-of-runtime-libraries-on-linux/55888/10.

In Chinese there’s an old saying called "making carts behind closed doors", indicating people doing things on their own but seldom share and discuss with others. I believe this is kind of the situation on Swift ports. Given that "broader platform support" is among the goals of Swift 6, the situation needs to be changed.

compnerd: > > It increases the size of the toolchain and the size of the download. This is a problem. As it is, the Windows toolchain is ~8G to download if you include the dependencies.
> 
> It ultimately increases the installed toolchain size by ~500MB, but there will be no download size increase for end-users because the installer can point multiple files to a single source. I don’t regard this as a really big sacrifice — I would say Swift is never doing so good at storage efficiency, and 500MB is even identical to the size of `.build` of a simple Swift package on macOS.

This is still a problem IMO.  The installed size is most definitely a problem - the growing dependency on Swift for building Swift means that we need the Swift toolchain and that will preclude the ability to have inspectable builds as we will need to switch to builds on my personal machine due to sizes.

> > Windows has always stated that it will aggressively move towards the newer functionality as feasible.
> 
> We’re not eliminating the new driver support by introducing "pluggable" `swift-driver.exe`. I don’t think Windows is a fan of breaking any programming interface, including CLI and API (fairly speaking, Apple is far more aggressive there). That is, we always encourage people to use the newest technology, but removing the old piece ahead of other platforms is still questionable.

See the previous point, we cannot include the additional installed content.  This additionally adds complexity to the system - you now need to think about the additional ways that pieces can interact.  Because there is a single driver the version explicitly tells everything.

> > That is fine - they are welcome to help improve the Windows platform. However, ultimately, I have spent an inordinate amount of personal time, money, and resources to bring the Windows support to this point, this is not paid work. I have finite resources, and I need to prioritise the overall Windows support.
> 
> I wish they are welcomed, but you need to realize that the community (and even core team) is relying on you on **nearly any** Windows pitch in general, and more maintainer will eventually tire you up if the fact doesn’t change.

I have yet to see anyone contributing to the core pieces that need the time and attention.  Even the installer was something that eventually was something that I had to spend time learning, implementing, and subsequently repairing.  There are more complications to the installer that I have not yet resolved.  There were simplistic things like file writing being problematic that no one helped fix.  So I hope you will forgive me for being reluctant to believe that there is someone waiting in the shadows to help with these deep fundamental issues.

> Discussing pre changes on the forums, pushing required CI testing, these can all help to free yourself to some extent. You and the core team are guiding the right direction, instead of making every decision by yourselves.

I definitely desire the required CI testing, but that has been difficult to get.  I've even gotten other pieces of infrastructure working in order to aid with this, but this still is taking quite a bit of time and effort.

> > I have yet to see anyone willing to help with the core fundamental issues that exist across the various pieces (e.g. LLVM, LLDB, etc.). Keeping the interfaces narrow is a consequence of that. The UX here is unperturbed for the most part - the new driver is meant to be a replacement.
> 
> I wish UX improvements are welcome too — I’m tired of replying to others _"You need to … here, this is the correct way to make it work on Windows"_, but instead I would say _"This is a known issue and will eventually get addressed, you need to … as a workaround"_. Doing correctly doesn’t mean doing well, and doing less (from user side) doesn’t necessarily mean doing wrong. Swift is initially designed to "make impossible tasks possible, and make difficult tasks simple", and I hope so for the Windows toolchain.

I would definitely be interested in learning about those cases.  I doubt that there are many "this is the correct way to make it work on Windows" and many of them are just things that need to be done as a workaround.  Most of these are driven by lack of resources.  As a good example - who is actually working on enabling the CAS on Windows?  I expect it a matter of time before that becomes necessary.  Restricting the paths to a single path helps ensure that I can try to account for them as things change.

As an example, had anyone ever said "passing `-sdk %SDKROOT%` is the correct way to make it work on Windows" may have thought that, but was something simply that hadn't been completed.  It was needed for a long time because it needed to be wired through at the various places.

> What if I regard the driver opt-out to be a feature? Remember that although Swift Driver is stated to be the drop-in replacement of `swift` and `swiftc` at the beginning, now it becomes a pure plugin of `swift-frontend`, and the developers are keeping this in mind. If you make any aggressive change, please mention it in the document or on the forums, or the compatibility will be easily broken by ignorant developers from other platforms.

That is fair - a reminder for that when 5.7 snapshots are actually announced would be nice.  However, there is the small struggle of the amount of time between the releases makes it harder to remember all the details.  Many of these changes have to often be reverted temporarily.  It would be nice if people actually reported issues more actively - I rarely see feedback until _after_ the release.

If you are adding this feature for your own use, and you intend to use it with a local built toolchain, that is fine.  I don't have a problem with vscode-swift supporting this, just as long as it is made clear that this this is an "advanced feature" which requires a local toolchain build and not the official toolchain.

stevapple: > I have yet to see anyone contributing to the core pieces that need the time and attention. Even the installer was something that eventually was something that I had to spend time learning, implementing, and subsequently repairing. There are more complications to the installer that I have not yet resolved. There were simplistic things like file writing being problematic that no one helped fix.

That’s exactly what I say by "making carts (port) behind closed door". I seldom hear from you on the forums reporting the progress of Windows port, and the last time I remembered was NIO. It was sadly abandoned mainly due to lack of Swift ecosystem on Windows IMO. That’s why we’re pushing these low-level frameworks to use System.

I don’t think there’s no people willing to help on Windows port. We’ve got volunteers for `vscode-swift` and for `winget-pkgs`. Most people are believing "Hey, this is Saleem, he knows everything for Windows, let’s wait for his magic" and they do not realize they could help. I would like to hear you on the forums more frequently rather than only being flooded with your PRs in my feed. I don’t think discussing and asking for help on the forums is useless, especially when you meet some bugs in a specific piece that you weren’t so familiar with.

> As an example, had anyone ever said "passing `-sdk %SDKROOT%` is the correct way to make it work on Windows" may have thought that, but was something simply that hadn't been completed.

This reminded me of being asked to "instruct the removal of Swift before starting to build a new one", while that is also an unresolved issue. There’re many flaws across the Windows toolchain, not only in the compiler. For example, why are we needing `%SDKROOT%` itself? This environment variable was used to **override** the host SDK on macOS, and shouldn’t exist in most cases.

> Restricting the paths to a single path helps ensure that I can try to account for them as things change.

I don’t think you need to account for all these yourself. We have SR (now issues) to track any experience downside on Windows. We have PR reviews to ensure everything won’t go wrong by mistake. AFAIK many reviewers don’t actually know what you’re doing through the Windows patches, and they simply believe that you’re always doing the right thing by approving. You can be the one saying "hey folks, Swift 5.x is going to be released, and let’s discuss the goal for Windows and then make it into reality". Of course, this would require both simple setup for local Windows build and CI support for PRs.

stevapple: And from your words, I would suggesting not using standalone `swift-driver.exe` is mainly a consideration of CI capacity, which is more of a money-related thing. Legacy driver support requires explicit opt-out and is something to use at one’s risk, and I don’t think that’s really something worth paying effort into. We can just leave it there.

I don’t know too much about how money is organized and put into CIs, but I would like to call out on the forums to address others’ attention on the issue. Toolchain size is another problems that the community is complaining about, but unfortunately it’s not an easy task to shrink the size. The most significant thing would be eliminating Swift LLVM, which is still far to achieve.

adam-fowler: I think this conversation has gone way off subject, and it should probably be moved to the swift forums.

I am going to close this PR though because the extension is primarily targeted at end users of swift not developers of swift. The likely hood of end users wanting to run the old swift driver on windows swift 5.7 is virtually 0% and all it is doing is protecting against running swift scripts (a minor element of the extension).

I am happy to include customisations so development versions of swift can be used but this one isn't worth it and it is certainly not worth the discussion above.

gwynne:
I concur with Adam- there is no use for this setting on Windows and never will be, as @compnerd pointed out at the beginning. Acknowledging the word of the toolchain’s maintainer in the future would probably save everyone a lot of time.

PR #276, opened 2022-04-25
toolchain: adjust the path for XCTest

compnerd:
XCTest parallel installs the XCTest runtime, and we should dynamically
select the last path component.  The following mapping applies:

  i686: bin32
  x86_64: bin64
  armv7: bin32a
  aarch64: bin64a

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -65,7 +65,12 @@ export class SwiftToolchain {
         const runtimePath = await this.getRuntimePath(targetInfo);
         const defaultSDK = await this.getDefaultSDK();
         const customSDK = this.getCustomSDK();
-        const xcTestPath = await this.getXCTestPath(runtimePath, defaultSDK);
+        const xcTestPath = await this.getXCTestPath(
+            targetInfo,
+            swiftVersion,
+            runtimePath,
+            defaultSDK
+        );
         return new SwiftToolchain(
             toolchainPath,
             targetInfo.compilerVersion,
@@ -182,6 +187,8 @@ export class SwiftToolchain {
      * @returns path to folder where xctest can be found
      */
     private static async getXCTestPath(
+        targetInfo: SwiftTargetInfo,
+        swiftVersion: Version,
         runtimePath: string | undefined,
         sdkroot: string | undefined
     ): Promise<string | undefined> {
@@ -212,14 +219,39 @@ export class SwiftToolchain {
                 if (!version) {
                     throw Error("Info.plist is missing the XCTEST_VERSION key.");
                 }
-                return path.join(
-                    platformPath,
-                    "Developer",
-                    "Library",
-                    `XCTest-${version}`,
-                    "usr",
-                    "bin"
-                );
+
+                if (swiftVersion >= new Version(5, 7, 0)) {
+                    let bindir: string;
+                    const arch = targetInfo.target?.triple.split("-", 1)[0];
+                    switch (arch) {
+                        case "x86_64":
+                            bindir = "bin64";
+                            break;
+                        case "i686":
+                            bindir = "bin32";
+                            break;
+                        case "aarch64":
+                            bindir = "bin64a";
+                            break;
+                        default:
+                            throw Error(`unsupported architecture ${arch}`);
+                    }
+                    return path.join(
+                        platformPath,
+                        "Developer",
+                        "Library",
+                        `XCTest-${version}`,
+                        bindir
+                    );
+                } else {
+                    return path.join(
+                        platformPath,
+                        "Developer",
+                        "Library",
+                        `XCTest-${version}`,
+                        "bin"
+                    );
+                }
             }
         }
         return undefined;


adam-fowler:
Is this a layout change for swift 5.7? Is the plan to recognise what platform we are running on and use the correct xctest.dll?

compnerd:
Yes, this is part of the 5.7 release.  The missing piece is exactly what you inferred: recognizing the architecture and mapping the path.  This is currently broken without this change, but it is also technically broken with this change if you want a different architecture.  So, maybe this should be okay to merge, but I dislike it because it's not checking the architecture.

stevapple:
We should support old toolchains as well, so I think the better way is to judge arch-specific path first, and if nothing found, fallback to `bin`.

compnerd:
@stevapple agreed, I just don't know how to do the check - I guess we could use the path APIs to check if the path exists.  That needs to be done before this is ready.

stevapple:
You can get current arch with `process.arch`: https://nodejs.org/api/process.html#processarch

In most cases this would be the arch to use. In the future, we will have target triple exposed from `Toolchain`.

compnerd:
I think that we want to use the target triple, not process.arch: I could be building a 32-bit binary with a 64-bit VSCode, which would not work.

stevapple:
I would argue to disable debugging and testing support for cross-compilation to another target by default. Most OSes don’t support running an executable of different arch or platform, and even on Windows this is limited (eg. we can run `x64` with `arm64` but not the revert).

Doing this on Windows also means we should provide separated runtimes for the host (used by the toolchain) and the target (used by the product). This adds up to complexity to some extent and we shouldn’t rush to support it.

compnerd:
I disagree entirely with you on this.  Windows and macOS are closer to each other than to Linux.

We can run x64, x86, and ARM64 on ARM64, but not the inverse, which also holds true for macOS AIUI.  XCTest is part of the SDK for precisely this reason.  There is no major complexity being added here other than the need to access the target triple.

stevapple:
Some points:

> XCTest is part of the SDK for precisely this reason.

On macOS, XCTest is a part of the toolchain instead of the SDK.

> There is no major complexity being added here other than the need to access the target triple.

We’ll be caching the target triple, this is a minimal work to access it. I think we should also hard-code the target arch in `launch.json`.

The major difficulty on Windows should be supporting different runtimes for the toolchain and the target. This needs a new configuration and should requires most effort in supporting cross debugging.

adam-fowler:
> We’ll be caching the target triple, this is a minimal work to access it

This is already cached, `WorkspaceContext.toolchain.defaultTarget` we just need to make it public, which I'll do now

adam-fowler:
> > We’ll be caching the target triple, this is a minimal work to access it
> 
> This is already cached, `WorkspaceContext.toolchain.defaultTarget` we just need to make it public, which I'll do now

Actually this is unnecessary. The point that xctest path is calculated is during the construction of the toolchain, so the triple can just be passed into `getXCTestPath`

src/toolchain/toolchain.ts
```diff
@@ -212,14 +214,28 @@ export class SwiftToolchain {
...
+                        break;
+                    case "aarch64":
+                        bindir = "bin64a";
+                        break;
+                    default:
+                        throw Error(`unsupported architecture ${arch}`);
+                }
+
+                const root = path.join(platformPath, "Developer", "Library", `XCTest-${version}`);
+                if (existsSync(path.join(root, "usr", bindir))) {
```
adam-fowler, line 235: I see you are basing this on the existence of the folder, and not swift version. Is this your preferable option? You could pass in `swiftVersion` calculated in `SwiftToolchain.create` into this and add a test 
```ts
if (swiftVersion >= new Version(5,7,0)) {
...
}
```
I'm not too fussed either way. The only thing I would say is the version check is clearer in code.

adam-fowler: Also if you don't want to use the swift version, could you use an async call from `fs/promises` instead of the blocking call. There is no version of `exists` but you could use `stat` I guess.

compnerd: I think that your point has more merit than you ascribe to it.  Really, practically speaking, the benefits aren't worth it here.  The `stat` check gives you compatibility with 1 (maybe 2) snapshots of 5.7, which are broken for other reasons.  Really, the check is only marginally more useful in theory and practically speaking, the version check is both easier to reason about and cleaner.  I'd argue that changing the code here is better.

stevapple (changes requested):
Some nits:
- `targetInfo` will never be undefined. Remove `| undefined` from definition;
- don’t import `fs`. We have `pathExists` that seals this functionality already.

src/toolchain/toolchain.ts
```diff
@@ -212,14 +213,40 @@ export class SwiftToolchain {
...
-                return path.join(
-                    platformPath,
-                    "Developer",
-                    "Library",
-                    `XCTest-${version}`,
-                    "usr",
-                    "bin"
-                );
+
+                const swiftVersion = await this.getSwiftVersion(targetInfo);
```
adam-fowler, line 217: Given we have already calculated the swift version in `SwiftToolchain.create` we could pass it into the function instead of parsing `targetInfo` again

adam-fowler (approved):
Looks good, one minor issue

src/toolchain/toolchain.ts
compnerd: 🤷‍♂️ sure, can do.

PR #284, opened 2022-04-30
toolchain: correct XCTest path on Windows

compnerd:
The shuffling and restructuring accidentally lost an entry in the path.
This corrects that which should repair part of the execution problems.


adam-fowler (approved):
oh dear

PR #285, opened 2022-04-30
Fixed issue with Windows Path in tests

adam-fowler:
When adding xcTestPath to the testEnv path it is possible for testEnv.Path to be undefined. In that case we need to drop back to using process.env.Path

### DIFF
--- a/src/debugger/launch.ts
+++ b/src/debugger/launch.ts
@@ -181,7 +181,7 @@ export function createTestConfiguration(
             return null;
         }
         if (xcTestPath !== runtimePath) {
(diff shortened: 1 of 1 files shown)


adam-fowler:
@stevapple @compnerd can one of you verify this.

compnerd (approved):
Thanks!  This does appear to fix the debugging and execution for me.

adam-fowler:
Hey @0xTim apparently the approval from @compnerd isn't enough to unblock merging. Can you approve?

0xTim (approved):
Yep!

PR #288, opened 2022-04-30
Add "build path" setting in oder to implement the enhancement from https://github.com/swift-server/vscode-swift/issues/283

smumriak:
Proposed change implements the "build path" setting for the extension and adds necessary logic to use this configuration location for swift package manager commands, sourcekit-lsp server and extension logic for things that need knowledge of swiftpm output.


src/utilities/utilities.ts
```diff
@@ -188,6 +188,29 @@ export function swiftpmSDKFlags(): string[] {
...
+    }
+}
+
+/**
+ * Get build path from configuration if exists or return a fallback .build directory in given workspace
+ * @param filesystem path to workspace that will be used as a fallback loacation with .build directory
+ */
+export function buildDirectoryFromWorkspacePath(workspacePath: string): string {
+    if (configuration.buildPath && configuration.buildPath.length > 0) {
+        return configuration.buildPath;
```
adam-fowler, line 208: Is it worthwhile supporting relative paths as well?

assets/test/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "swift.buildPath": "/some/other/full/test/path"
```
adam-fowler, line 2: Is this meant to be included in the PR?

src/utilities/utilities.ts
```diff
@@ -188,6 +188,29 @@ export function swiftpmSDKFlags(): string[] {
     return [];
 }
 
+/**
+ * Get build path flags to be passed to swift package manager and sourcekit-lsp server
+ */
+export function buildPathFlags(): string[] {
```
adam-fowler, line 194: Is it worthwhile combining this with the sdk flags functions. And instead have `configSwiftPMFlags` and `configSwiftDriverFlags` functions which are filled with swift build flags setup from the configuration.

@stevapple what's your thoughts. I know the `sdk` argument has special placement issues.

adam-fowler (commented):
In general looks good, I've left some questions.

src/utilities/utilities.ts
stevapple: > Is it worthwhile combining this with the sdk flags functions. And instead have configSwiftPMFlags and configSwiftDriverFlags functions which are filled with swift build flags setup from the configuration.

Agree on the initiative, but since some flags are unnecessary (eg. `--sdk` for `swift package`) it would be hard to do so.

Maybe better extracting these functions into `./utils/flags.ts`?

> I know the `sdk` argument has special placement issues.

I skimmed over the PR and think it won't break here.

adam-fowler: > Maybe better extracting these functions into ./utils/flags.ts?

Lets leave that for another PR

src/utilities/utilities.ts
smumriak: The file is created as a result of running tests. I figures if we are saving test assets anyway it might have sense to keep it in repository. I can remove it if you feel like it should not be in the repository

assets/test/.vscode/settings.json
smumriak: The file is created as a result of running tests. I figures if we are saving test assets anyway it might have sense to keep it in repository. I can remove it if you feel like it should not be in the repository

src/utilities/utilities.ts
smumriak: I'm not sure if i correctly understand your question . Are you asking if user can supply ".customBuildFolder" in this setting? It should work as is. The only thing I'm not sure is how vscode.Uri would resolve it. I can test it a bit later and if it's weird - we can just do logic that merger workspace path with relative path in the setting

assets/test/.vscode/settings.json
adam-fowler: It will effect other tests, so not sure we should have it here, maybe the build path tests should be reseting this once they are done.

src/utilities/utilities.ts
adam-fowler: Yeah I guess you're right with regards to parameters for swift calls, the relative path will work. launch.json configs will probably need prefixed with `${workspaceFolder:<folder name>}`

assets/test/.vscode/settings.json
smumriak: Yea, removing the actual setting is something the test teardown should perform, I'll add necessary code there. Tho I would still leave the empty settings.json, since there's also a test in WorkspaceContest.tests.ts that updates settings (tho that one is weird and I'm not sure it works correctly, since it is using `vscode.workspace(...).getConfiguration(...).update(...)` function synchronously, while it is supposed to be async)

src/utilities/utilities.ts
adam-fowler: We can leave this for now. When a PR for moving all the swift flags out to a separate file is done, tests can be added at that point.

assets/test/.vscode/settings.json
adam-fowler: An empty settings.json is fine

adam-fowler:
Hi @smumriak , do you have time to look at the issues brought up?

smumriak:
Hello @adam-fowler, sorry, those were two very busy weeks at my. I'll send updated PR tomorrow evening (looks like the environment variables PR was landed, so some conflicts are in place)

smumriak:
Updated revision:
- resolved conflicts
- added "swift package resolve" command as the one that needs SDK and build path arguments (it uses the checkouts directory of build path to do package resolution)
- settings.json is now empty file, test is correctly cleaning it up
- tested relative path in configuration - works as expected. launch.json is something we probably want to discuss further

stevapple:
> * added "swift package resolve" command as the one that needs SDK and build path arguments (it uses the checkouts directory of build path to do package resolution)

Don’t specialize `resolve` here. `--build-path` should be injected on every SwiftPM call, while destination flags can be eliminated in some cases.

smumriak:
> Don’t specialize resolve here. --build-path should be injected on every SwiftPM call, while destination flags can be eliminated in some cases.

i'm a bit confused. could you please elaborate more? what do you mean by "destination flags can be eliminated in some cases"?

stevapple:
IMO your implementation has the following problems:
- build path is SwiftPM-only, so we shouldn't inject it for `default` cases;
- build path should be injected on every SwiftPM call.

I would suggest the following:
```typescript
/**
 * Get modified swift arguments with configuration flags.
 *
 * @param args original commandline arguments
 */
 export function withSwiftFlags(args: string[]): string[] {
    switch (args[0]) {
        case "package": {
            const subcommand = args.splice(0, 2).concat(buildPathFlags());
            switch (subcommand[1]) {
                case "dump-symbol-graph":
                case "diagnose-api-breaking-changes":
                    // These two tools require building the package, so SDK
                    // flags are needed. Destination control flags are
                    // required to be placed before subcommand options.
                    return [...subcommand, ...swiftpmSDKFlags(), ...args];
                default:
                    // Other swift-package subcommands operate on the host,
                    // so it doesn't need to know about the destination.
                    return subcommand.concat(args);
            }
        }
        case "build":
        case "run":
        case "test": {
            const subcommand = args.splice(0, 1).concat(buildPathFlags());
            return [...subcommand, ...swiftpmSDKFlags(), args];
        }
        default:
            // We're not going to call the Swift compiler directly and the
            // settings are package-only, so do nothing here.
            return args;
    }
}
```

smumriak:
ah, I see what you mean. yea, you are right. let me update the PR

smumriak:
Updated

package.json
```diff
@@ -145,6 +145,11 @@
             "default": false,
             "markdownDescription": "**Experimental**: Run `swift build` in the background whenever a file is saved.",
             "order": 7
+          },
+          "swift.buildPath": {
+            "type": "string",
```
adam-fowler, line 150: We need to add a description for the setting. Also if you could add a more detailed description in `docs/settings.md`.

src/SwiftPackage.ts
```diff
@@ -216,7 +220,10 @@ export class SwiftPackage implements PackageContents {
      */
     public async loadWorkspaceState(): Promise<WorkspaceState | undefined> {
         try {
-            const uri = vscode.Uri.joinPath(this.folder, ".build", "workspace-state.json");
+            const uri = vscode.Uri.joinPath(
+                vscode.Uri.file(buildDirectoryFromWorkspacePath(this.folder.fsPath)),
+                "workspace-state.json"
+            );
             const contents = await fs.readFile(uri.fsPath, "utf8");
```
adam-fowler, line 227: This fails if the build directory is a relative path

src/ui/PackageDependencyProvider.ts
```diff
@@ -176,16 +176,14 @@ export class PackageDependenciesProvider implements vscode.TreeDataProvider<Tree
                 first.name.localeCompare(second.name)
             );
         }
+
+        const buildDirectory = buildDirectoryFromWorkspacePath(folderContext.folder.fsPath);
```
adam-fowler, line 180: Build directory can't be relative here.

src/utilities/utilities.ts
adam-fowler: I tested relative build paths and the `launch.json` configs look to work fine with a relative path. I have noted two cases where it does not work though. When loading `workspace-state.json` and when searching for files in package dependency view.

Would it be possible to add a second parameter `fullpath` to `buildDirectoryFromWorkspacePath` which indicates we want a non relative path. If it is set to true and `buildPath` doesn't start with a '/' then return `path.join(workspacePath, configuration.buildPath);`. In the two cases I mentioned above where it needs the full path, we should set this flag to true.

I don't want a full path all the time because I would rather not pollute the `launch.json` with details of where the project is installed.

adam-fowler:
One other thing that'll be needed is for you to rebuild the `launch.json` configurations whenever you change `configuration.buildPath`. You can see code for how to do that here https://github.com/swift-server/vscode-swift/blob/a50127896ab671671c175886a5d8ebd57be64247/src/WorkspaceContext.ts#L85-L103

src/utilities/utilities.ts
smumriak: Checking for "/" prefix is easy. what's not easy is to do the same for windows disks. like, needs a regex and, to be honest, i don't remember windows disk naming rules anymore. is it always single character? virtual disks can have more than one? can disk label contain numbers?

smumriak: damn windows. there are also network disks and paths to files on those can start with "\\"

smumriak: the good news is the fact that node's module "path" is smart enough to make this work

src/ui/PackageDependencyProvider.ts
smumriak: Basically anything that uses vscode-uri can not use relative path. This is tricky for future support, imo. Worth discussing how we could make this safer in future

smumriak:
Updates: 
- added docs
- added argument "absolute" (to be on-par with naming  from "path" node module) in "buildDirectoryFromWorkspacePath" function and set it to true whenever code needs to work with absolute paths (all places whrere vscode-uri) is used
- adjusted tests

adam-fowler:
@swift-server-bot test please

adam-fowler:

adam-fowler:
This all looks good, except I can't get the tests to work from inside VSCode. They return this error
```
Error: Unable to write to Workspace Settings because no workspace is opened. Please open a workspace first and try again.
```

smumriak:
Oh, i've been using `npm test` to run tests from terminal. Let me look into running tests from vscode

adam-fowler:
> Oh, i've been using `npm test` to run tests from terminal. Let me look into running tests from vscode

`npm test` fails for me as well. Keeps loading the previous workspace I had loaded

smumriak:
Interesting. are you running on a mac or linux?

adam-fowler:
> Interesting. are you running on a mac or linux?

Mac

smumriak:
messaged you in slack

adam-fowler:
@swift-server-bot add to whitelist

adam-fowler:

adam-fowler (approved):
I made a couple of changes to the settings readme, and this is now ready to merge

PR #294, opened 2022-05-02
Prepare plugin for cross-compilation support

stevapple:
- Add "Cross compilation" config group;
- Group `destination` configs;
- Function renames;
- Update settings document.


stevapple:
@adam-fowler What do you think?

Since renaming `swift.SDK` to `swift.destination.SDK` will cause compatibility breakage, we'd better fix this layout before releasing it to public.

stevapple:
Superseded by #386

PR #297, opened 2022-05-08
Only set `target.sdk-path` if CodeLLDB setup Swift LLDB

adam-fowler:
If failed to setup CodeLLDB to use the Swift version of LLDB, then don't set setting specific to Swift LLDB. 
Also show error message when failing to setup CodeLLDB.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "swift-lang",
-  "version": "0.4.3",
+  "version": "0.5.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "swift-lang",
-      "version": "0.4.3",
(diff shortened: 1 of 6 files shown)


compnerd (commented):
This seems like a safe change.  However, do we know why/when CodeLLDB fails to setup the path?  I think that it may currently never setup on Windows and requires the user to do that, this the message may want to direct the user towards the configuration.

adam-fowler:
It does set it up at least that's what I've been led to believe. My situation was the swift lldb failed to start at all because it could not find python310.dll. Without this change you don't see it failed. 

When I started debugging tests I got the error, target.sdk_path is not a valid setting.

compnerd (approved):
SGTM

0xTim (approved):
LGTM

Issue #283, opened 2022-04-29
Support build path flag that for all vscode-swift operations


smumriak:
**Is your feature request related to a problem? Please describe.**
Swift Package Manager provides ability to customize the location of "build directory" via `--build-path` argument. Sourcekit-LSP extension used to provide setting called `sourcekit-lsp.serverArguments`, which was successfully adopted by vscode-swift. There's also `swift.buildArguments` setting. Both of them allow setting the build directory, but it is only used by build commands and by LSP itself. Other useful features like test view, test running, launch task generation and package resolve are using hardcoded assumption of `.build` directory instead.

**Describe the solution you'd like**
Provide a dedicated extension setting that will be used as an argument for `--build-path` of LSP server and all swift package manager commands spawned by vscode-swift

**Describe alternatives you've considered**
Keeping the inability of setting custom build directory.

adam-fowler:
There is also the alternative of including a `swift.commonArguments` setting which is a list of arguments applied to all swift calls.

smumriak:
While this is viable idea for everything related to execution of swift package manager commands and sourcekit language server, this will, unfortunately, not work for next cases:
- creating launch tasks. this is not a swift command, but it requires to know the location of build artifacts in to pass the executable to lldb
- generating tests for dependencies, since the extension code needs to know location of those dependencies checkouts locations in order to crawl them. this is actually the same of any dependency-related operation that is performed by extension itself

stevapple:
I think this is worth an option. Add to personal todos.

smumriak:
I'm sending the PR right now :)

stevapple:
> I'm sending the PR right now :)

Go ahead:) There is a `withSwiftCompileFlags()` in `utils` that you can make use of.

markst:
👍

Might it possible to pass the Xcode DerivedData default path?
```
"swift.buildPath": "~/Library/Developer/Xcode/DerivedData",
```

to match that of what might be passed to `xcodebuild`?
```
xcodebuild -scheme Demo -configuration Debug -sdk iphonesimulator -derivedDataPath ~/Library/Developer/Xcode/DerivedData -allowProvisioningUpdates ARCHS=arm64 
```

adam-fowler:
> Might it possible to pass the Xcode DerivedData default path?

I don't think that would work. Xcode has a different structure to its build data. Even if it was structured the same you would have to include the project folder inside DerivedData.

adam-fowler:
[...]

PR #304, opened 2022-05-13
Pass swift build arguments to sourcekit-lsp

adam-fowler:
This should resolve an issue where swift build needs to do a full rebuild all the time.

### DIFF
--- a/src/sourcekit-lsp/LanguageClientManager.ts
+++ b/src/sourcekit-lsp/LanguageClientManager.ts
@@ -247,7 +247,7 @@ export class LanguageClientManager {
         const sdkArguments = [
             ...swiftDriverSDKFlags(true),
             ...platformDebugBuildOptions(),
-            ...configuration.buildArguments,
+            ...this.filterBuildArguments(configuration.buildArguments),
         ];
         const sourcekit: langclient.Executable = {
             command: serverPath,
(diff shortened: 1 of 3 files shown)


stevapple (changes requested):
I don't think it's right to do this. Ideally SourceKit-LSP should accept the same arguments as SwiftPM, but it's completely not the case now. Passing valid SwiftPM flags to SourceKit-LSP could easily make it crashed.

adam-fowler:
> I don't think it's right to do this. Ideally SourceKit-LSP should accept the same arguments as SwiftPM, but it's completely not the case now. Passing valid SwiftPM flags to SourceKit-LSP could easily make it crashed.

I asked @ahoppen who works on Sourcekit-LSP and he said it should be fine.

stevapple:
> I asked @ahoppen who works on Sourcekit-LSP and he said it should be fine.

The first piece of difference I ran into is the missing support of build destination control in SK-LSP. I don’t know if there’s more, but doing this has much higher and more serious risks than you may think of.

adam-fowler:
Are you saying if someone changes the build destination sourcekit lsp won't pick that up?

stevapple:
> Are you saying if someone changes the build destination sourcekit lsp won't pick that up?

Worse: it will crash immediately (if we directly pass SwiftPM destination flags to it).

That’s why we must handle cross-compilation flags one by one through settings.

adam-fowler:
Ok it looks like we can pass through `--build-path`, `-X...` family, as these should all be supported. If they aren't then that is a bug. The `--build-path` argument should eventually come from #283.

adam-fowler:
@stevapple i added the argument filter. This should cover issue you brought up.

stevapple:
For options we’re also supporting `--build-path=<path>` style. Can we add support for that too?🤔

adam-fowler:
> For options we’re also supporting `--build-path=<path>` style. Can we add support for that too?🤔

Given we are eventually going to get the build path from a separate parameter maybe we can ignore this and remove it from here.

0xTim (approved):
No objections on my part

adam-fowler:
> > For options we’re also supporting `--build-path=<path>` style. Can we add support for that too?🤔
> 
> Given we are eventually going to get the build path from a separate parameter maybe we can ignore this and remove it from here.

Changed my mind and added support for `--build-path=<path>` style

PR #305, opened 2022-05-13
Add setting for environment variables to send to Swift

adam-fowler:
Something like this @stevapple?

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -77,6 +77,12 @@ const configuration = {
     get buildArguments(): string[] {
         return vscode.workspace.getConfiguration("swift").get<string[]>("buildArguments", []);
     },
+    /** Environment variables to set when building */
+    get swiftEnvironmentVariables(): { [key: string]: string } {
+        return vscode.workspace
+            .getConfiguration("swift")
+            .get<{ [key: string]: string }>("swiftEnvironmentVariables", {});
+    },
     /** include build errors in problems view */
     get problemMatchCompileErrors(): boolean {
         return vscode.workspace
@@ -100,10 +106,10 @@ const configuration = {
         return vscode.workspace.getConfiguration("swift").get<boolean>("diagnostics", false);
     },
     /** Environment variables to set when running tests */
-    get testEnvironmentVariables(): Record<string, string> {
+    get testEnvironmentVariables(): { [key: string]: string } {
         return vscode.workspace
             .getConfiguration("swift")
-            .get<Record<string, string>>("testEnvironmentVariables", {});
+            .get<{ [key: string]: string }>("testEnvironmentVariables", {});
     },
 };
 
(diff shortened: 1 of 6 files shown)


src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -247,7 +247,7 @@ export class LanguageClientManager {
         const sourcekit: langclient.Executable = {
             command: serverPath,
             args: lspConfig.serverArguments.concat(sdkArguments),
-            options: { env: { ...process.env, ...swiftRuntimeEnv() } },
+            options: { env: { ...configuration.swiftEnvironmentVariables, ...swiftRuntimeEnv() } },
```
stevapple, line 250: `process.env` should also be passed here

src/configuration.ts
```diff
@@ -77,6 +77,12 @@ const configuration = {
     get buildArguments(): string[] {
         return vscode.workspace.getConfiguration("swift").get<string[]>("buildArguments", []);
     },
+    /** Environment variables to set when building */
+    get swiftEnvironmentVariables(): { [key: string]: string } {
```
stevapple, line 81: I guess we can extract `Object.keys(swiftEnvironmentVariables).length > 0` in to a new computed property?

stevapple (changes requested):
LGTM otherwise

stevapple:
I think this could close #266 and I'll continue the remaining work based on #294

src/sourcekit-lsp/LanguageClientManager.ts
adam-fowler: Just found we need to do the same with `execSwift`. It looks like we completely clear all other environment variables by setting one with `env` flags.

adam-fowler: Fixed

src/configuration.ts
adam-fowler: Not sure this is totally necessary, I do the same in execFile to make it more obvious why we are setting runtimePath env.

src/utilities/utilities.ts
```diff
@@ -79,8 +79,15 @@ export async function execFile(
...
     );
-    if (configuration.runtimePath.length > 0 && customSwiftRuntime) {
-        options.env = { ...options.env, ...swiftRuntimeEnv(options.env) };
+    if (customSwiftRuntime) {
+        const runtimeEnv = swiftRuntimeEnv(options.env);
+        if (runtimeEnv && Object.keys(runtimeEnv).length > 0) {
+            options.env = { ...options.env, ...runtimeEnv };
+        }
+    }
+    // it options contain environment keys add them to process environment keys
```
0xTim, line 88: Not spelling
```suggestion
    // if options contain environment keys add them to process environment keys
```

0xTim (approved):
One minor spelling, otherwise LGTM

src/utilities/utilities.ts
adam-fowler: This code got deleted so no need to fix comment. It is now setup the same as in `execFileStreamOutput`

PR #307, opened 2022-05-16
Treat Windows Swift LLDB setup fail differently

adam-fowler:
If Windows Swift LLDB setup fails it will most likely be due to LLDB failing to run because $PYTHONHOME has been setup incorrectly. This means Swift LLDB will fail to run the debugger and should not be setup. 

The code change here just means failure to setup Swift LLDB on Windows will fail silently and will not attempt to guess the location of the Swift LLDB DLL.

### DIFF
--- a/src/WorkspaceContext.ts
+++ b/src/WorkspaceContext.ts
@@ -284,9 +284,11 @@ export class WorkspaceContext implements vscode.Disposable {
     async setLLDBVersion() {
         const libPathResult = await getLLDBLibPath(this.toolchain);
         if (!libPathResult.success) {
-            const errorMessage = libPathResult.failure
-                ? `Error: ${getErrorDescription(libPathResult.failure)}`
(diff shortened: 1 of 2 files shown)


compnerd (approved):
While the change itself makes sense (once one understands the reasoning behind it), it feels as though the current comments don't convey the actual reasoning which is important to understand IMO.

Extending the comment in lldb.ts to explain that this is to prevent the setup of LLDB's liblldb.dll on Windows with older releases of the toolchain which were built improperly and would look for python explicitly in C:\hostedcache, which is unlikely to exist, rather than on the system would help others understand why the platform was singled out in this path.

I could also see this being a further test in the future - one where if the python installation exists, use it.  However, I think that preferring 5.7 on Windows is a better approach as a whole (each release still makes a considerable improvement to usability).

0xTim (approved):

PR #316, opened 2022-05-26
Add support for running a single LSP server that has multi-workspace support

adam-fowler:
When Swift 5.7 is released it will include a version of SourceKit-LSP which include multi-workspace support. This will allow us to run a single server and add and remove workspace folders to it when they are added/removed to the VSCode workspace. This is an alternative to what we have now where the server is restarted every time we focus on a new workspace folder.

This PR keeps the original restart solution as it is needed for Swift 5.6 and earlier, but will use the single server solution if the version of Swift is 5.7 or later.


adam-fowler:

0xTim (approved):
🎉

PR #323, opened 2022-06-02
Active folder and settings tests

adam-fowler:
- Fix setting workspace folder when running tests from VSCode
- Reset any config settings setup in tests
- Add test for setting swift path
- Add test for adding new FolderContext based on the active text document


smumriak:
Could someone please review this change? maybe @stevapple? this unblocks a lot of possibilities for testing

stevapple (approved):
I didn’t see any problem, but I don’t have access so we should still wait for an org member to approve:)

0xTim (approved):
LGTM

PR #362, opened 2022-07-16
Add language featues section to README

adam-fowler:
This has been added so we can tell the user to build their app before expect sourcekit-lsp to work.
It seems a pretty common issue, so adding to the README seems a good idea

@0xTim Not sure about the wording, don't know if you have any better ideas

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,10 @@ For the extension to work, you must have Swift installed on your system. Please 
 
 ## Features
 
+### Language features
+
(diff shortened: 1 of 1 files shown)


0xTim (approved):
Yeah looks good to me. I wonder if we could kick off a build in the background when the project opens and some messaging in the UI to say `building code for symbols` or similar. This would match what Xcode does in a way

adam-fowler:
> Yeah looks good to me. I wonder if we could kick off a build in the background when the project opens and some messaging in the UI to say `building code for symbols` or similar. This would match what Xcode does in a way

We could kick off a build in the background whenever a project opens, but if it fails symbol completion will not work. So it may still cause confusion. Although we could message something about that. Anyway should do that in another PR

PR #372, opened 2022-07-24
change icon to `package`

aelam:
It seems `package` is more accurate than `archive`

https://code.visualstudio.com/api/references/icons-in-labels#icon-listing 
Screen Shot 2022-07-25 at 7 34 49
Screen Shot 2022-07-25 at 7 35 02

### DIFF
--- a/src/ui/PackageDependencyProvider.ts
+++ b/src/ui/PackageDependencyProvider.ts
@@ -52,7 +52,7 @@ export class PackageNode {
         item.iconPath =
             this.type === "editing"
                 ? new vscode.ThemeIcon("edit")
-                : new vscode.ThemeIcon("archive");
+                : new vscode.ThemeIcon("package");
         item.contextValue = this.type;
         return item;
     }


0xTim:
@swift-server-bot test

adam-fowler:

adam-fowler:
Just because the icon is called `package` doesn't make it correct. It probably looks better, but at the same time it has an annoying space before it as well. If we could get rid of that I'd be happy

aelam:
> Just because the icon is called `package` doesn't make it correct. It probably looks better, but at the same time it has an annoying space before it as well. If we could get rid of that I'd be happy

what do you mean `annoying space`

adam-fowler:
> what do you mean `annoying space`

There is a space before the icon

aelam:
it's not a space, I just screenshot them from https://code.visualstudio.com/api/references/icons-in-labels#icon-listing
I didn't consider the alignment

adam-fowler:
It looked like it added a space when I used. I may have been wrong

adam-fowler (approved):

PR #376, opened 2022-07-28
fix invalid JSON in devcontainer.json example

kelvin13:
trailing comma is not allowed in JSON (although VSCode can tolerate it)


adam-fowler (approved):
👍

PR #379, opened 2022-07-30
show editing package version like `editing 1.2.3`

aelam:
show editing version like  `editing 1.2.3`

blocker #378

### DIFF
--- a/src/ui/PackageDependencyProvider.ts
+++ b/src/ui/PackageDependencyProvider.ts
@@ -21,7 +21,7 @@ import { WorkspaceContext } from "../WorkspaceContext";
 import { FolderEvent } from "../WorkspaceContext";
 import { FolderContext } from "../FolderContext";
 import contextKeys from "../contextKeys";
-import { WorkspaceState } from "../SwiftPackage";
+import { WorkspaceState, WorkspaceStateDependency } from "../SwiftPackage";
 
 /**
  * References:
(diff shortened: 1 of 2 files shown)


src/ui/PackageDependencyProvider.ts
```diff
@@ -292,7 +292,13 @@ export class PackageDependenciesProvider implements vscode.TreeDataProvider<Tree
     private dependencyDisplayVersion(dependency: WorkspaceStateDependency): string {
         const type = this.dependencyType(dependency);
         if (type === "editing") {
-            return "editing"; // ?TODO: get version from `baseOn` node for showing `editing 1.2.3`
+            if (dependency.basedOn) {
```
adam-fowler, line 295: If a user is editing a package using a local version of the dependency ie they ran 
```
swift package edit <dependency> -path <local-version-folder>
```

 I don't think we can say it is based on a particular version. We can only say it is `editing <version>` when they have  called 

```
swift package edit <dependency>
```

aelam: editing is ok to me no strong opinion about it 

but I'm trying to understand the real use case of edit 

When is the right scenario that we should use edit?
-  it has been a dependency already
- it can't be a local dependency before you edit 


So im thinking the scenario is that 
* If it's local dependency you can change whatever freely which is nothing about edit
* if you edit remote dependency, it means the engineer wants to make some change on current remote version/branch, then they get a chance to fix the existing issues of current version (or switching other version? to check)

adam-fowler: The two versions of `swift package edit` are for different situations 

1) without a path is generally used for testing a quick change to a dependency. I just need an editable copy of the current version of that dependency. Including the version in the editing string makes sense here as your changes are based of a specific version of the dependency.
2) With a path argument is used when I have been developing a feature in one of the dependencies independently of the current package and I want to test that new feature in the current package. I can then say I would rather use this version of my dependency over the version downloaded from the repository. In this case because there is no versioning link between the two versions of the dependency it doesn't make sense to include the version number in the editing string

Does that make sense?

adam-fowler:

src/ui/PackageDependencyProvider.ts
aelam: yes. fully understand these cases. 
if we really want to show the editing version in the case 2, probably we should also show the real version of the dependency in case 1

Do we continue this PR to see the result. here I summarize a bit. 
1. show `editing 1.2.3` for case 1, show `editing` for case 2
2. show `editing 1.2.3` for case 1 and case 2. needs to read info `swift package describe`
3. leave them as `editing`. close this 

Which one do you prefer. 
If it's 3, I'd like to create another PR about showing menu of `use locally` by right click, so that `swift package edit` can be done in UI action

Issue #369, opened 2022-07-23
Add command to set Xcode tools


adam-fowler:
**Is your feature request related to a problem? Please describe.**
Add command that displays selection of Xcode installed on system and then set DEVELOPER_DIR in build environment variables based off selection

Find Xcode's installed
`mdfind "kMDItemCFBundleIdentifier == 'com.apple.dt.Xcode'"`
Get version name from install
`/usr/libexec/PlistBuddy -c "Print :CFBundleShortVersionString" ${XCODE}/Contents/Info.plist`

PR #381, opened 2022-07-31
Add quick switcher between iOS and macOS targets

complexspaces:
This PR adds a command palette action to the extension that allows a user to quickly switch between the macOS and iOS SDK targets. It's only available if the host is macOS because the iOS SDK is not (officially) available for distribution on other operating systems.

The goal of these changes are to make the extension a more reasonable replacement for XCode by getting closer to the cross-compiling convenience that XCode offers through its UI. 

https://user-images.githubusercontent.com/68570223/182044190-ccabae14-b421-4a9d-aa5b-66412b91f4fe.mov

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,11 @@
         "category": "Swift"
       },
       {
+        "command": "swift.switchPlatform",
+        "title": "Switch Target",
+        "category": "Swift"
+      },
+      {
         "command": "swift.resetPackage",
         "title": "Reset Package Dependencies",
         "icon": "$(clear-all)",
@@ -244,6 +249,10 @@
         {
           "command": "swift.cleanBuild",
           "when": "swift.hasPackage"
+        },
+        {
+          "command": "swift.switchPlatform",
+          "when": "isMac"
         },
         {
           "command": "swift.resetPackage",
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -19,8 +19,11 @@ import { FolderEvent, WorkspaceContext } from "./WorkspaceContext";
 import { createSwiftTask, SwiftTaskProvider } from "./SwiftTaskProvider";
 import { FolderContext } from "./FolderContext";
 import { PackageNode } from "./ui/PackageDependencyProvider";
+import { withQuickPick } from "./ui/QuickPick";
 import { execSwift } from "./utilities/utilities";
 import { Version } from "./utilities/version";
+import { DarwinCompatibleTarget, SwiftToolchain } from "./toolchain/toolchain";
+import configuration from "./configuration";
 
 /**
  * References:
@@ -428,6 +431,34 @@ function openInExternalEditor(packageNode: PackageNode) {
     }
 }
 
+interface DarwinQuickPickTarget extends vscode.QuickPickItem {
+    value: DarwinCompatibleTarget;
+    label: string;
+}
+
+/**
+ * Switches the target SDK to the platform selected in a QuickPick UI.
+ */
+async function switchPlatform() {
+    const onSelect = async (picked: DarwinQuickPickTarget) => {
+        const sdkForTarget = await SwiftToolchain.getSdkForTarget(picked.value);
+        if (sdkForTarget) {
+            configuration.sdk = sdkForTarget;
+        } else {
+            vscode.window.showErrorMessage("Unable to obtain requested SDK path");
+        }
+    };
+
+    await withQuickPick<DarwinQuickPickTarget>(
+        "Select a new target",
+        [
+            { value: DarwinCompatibleTarget.macOS, label: "macOS" },
+            { value: DarwinCompatibleTarget.iOS, label: "iOS" },
+        ],
+        onSelect
+    );
+}
+
 function updateAfterError(result: boolean, folderContext: FolderContext) {
     const triggerResolvedUpdatedEvent = folderContext.hasResolveErrors;
     // set has resolve errors flag
@@ -450,6 +481,8 @@ export function register(ctx: WorkspaceContext) {
         ),
         vscode.commands.registerCommand("swift.updateDependencies", () => updateDependencies(ctx)),
         vscode.commands.registerCommand("swift.cleanBuild", () => cleanBuild(ctx)),
+        // Note: This is only available on macOS (gated in `package.json`) because its the only OS that has the iOS SDK available.
+        vscode.commands.registerCommand("swift.switchPlatform", () => switchPlatform()),
         vscode.commands.registerCommand("swift.resetPackage", () => resetPackage(ctx)),
         vscode.commands.registerCommand("swift.runSingle", () => runSingleFile(ctx)),
         vscode.commands.registerCommand("swift.openPackage", () => openPackage(ctx)),
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -46,6 +46,15 @@ interface SwiftTargetInfo {
     [name: string]: string | object | undefined;
 }
 
+/**
+ * A Swift compilation target that can be compiled to
+ * from macOS. These are similar to XCode's target list.
+ */
+export enum DarwinCompatibleTarget {
+    iOS,
+    macOS,
+}
+
 export class SwiftToolchain {
     constructor(
         public swiftFolderPath: string,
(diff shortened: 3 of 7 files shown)


complexspaces:
I was hoping that this would work in the context of a standalone Swift file (no SPM package present), but it doesn't seem to. I don't know if this is because my changes are wrong or because the extension doesn't currently support this. Single files seem to only work for the host's SDK.

stevapple:
@complexspaces Thanks for your PR and I think it’s a good idea.

As for cross-compiling a standalone file, I didn’t add the support because scripts are mostly intended to be run on the host machine directly, but it’s certainly useful to cross-compile a file without running it immediately. Perhaps we can add an action to generate a `task.json` entry for the current script?

---

I initially designed the cross-compilation ability with `swift build` arguments, but it ends up making the build command too long and introducing divergence between SwiftPM and SourceKit-LSP. Recently I’ve thought about redesigning the feature with autogenerated `destination.json` by the plugin. If you have any suggestions you could just point out.

package.json
```diff
@@ -44,6 +44,11 @@
         "title": "Clean Build",
         "category": "Swift"
       },
+      {
```
adam-fowler, line 47: You need to add the command to `commandPalette` as well. Add a "when" section to only enable for "macOS"

src/commands.ts
```diff
@@ -428,6 +430,43 @@ function openInExternalEditor(packageNode: PackageNode) {
...
 
+interface IDawrinCompatibleTarget extends vscode.QuickPickItem {
+    value: DarwinCompatibleTarget;
+    label: string;
+}
+
+async function handleQuickSwitch() {
+    const picker = vscode.window.createQuickPick<IDawrinCompatibleTarget>();
+
+    picker.placeholder = "Select a new target";
```
adam-fowler, line 441: I would probably move the QuickPick code outside of `commands.ts`. It would be good to generalise it if possible. I have another task #369 that'll need one. Maybe add a generalised version inside the UI folder. Provide list of menu entries which include a label and a function to call when selected (something like that).

src/commands.ts
```diff
@@ -428,6 +430,43 @@ function openInExternalEditor(packageNode: PackageNode) {
     }
 }
 
+interface IDawrinCompatibleTarget extends vscode.QuickPickItem {
+    value: DarwinCompatibleTarget;
+    label: string;
+}
+
+async function handleQuickSwitch() {
```
adam-fowler, line 438: QuickSwitch is a very vague function name, SwitchPlatform maybe

src/utilities/utilities.ts
```diff
@@ -251,6 +251,36 @@ export function swiftDriverSDKFlags(indirect = false): string[] {
     return indirect ? args.flatMap(arg => ["-Xswiftc", arg]) : args;
 }
 
+/**
+ * Get target flags for swiftc
+ *
+ * @param indirect whether to pass the flags by -Xswiftc
+ */
+export function swiftDriverTargetFlags(indirect = false): string[] {
```
adam-fowler, line 259: Not sure about setting the `target` command line parameter based on the sdk parameter. It seems a little fragile although I don't see a quick and easy way to set it otherwise.

src/toolchain/toolchain.ts
```diff
@@ -46,6 +46,11 @@ interface SwiftTargetInfo {
     [name: string]: string | object | undefined;
 }
 
+export enum DarwinCompatibleTarget {
```
adam-fowler, line 49: do we want to support building for the simulator as well as device.

adam-fowler (commented):
Thank you for the contribution. In general this looks good. A few comments

src/toolchain/toolchain.ts
complexspaces: I wasn't sure about how to expose the simulator, hence why I didn't add it to start. Do you think it should be (in the switcher UI), labeled something like `iOS (Simulator)`? I have not personally written any code that required simulator guards so I can't speak for how useful this is.

complexspaces:
> As for cross-compiling a standalone file, I didn’t add the support because scripts are mostly intended to be run on the host machine directly, but it’s certainly useful to cross-compile a file without running it immediately. Perhaps we can add an action to generate a task.json entry for the current script?

@stevapple My use case is probably fairly unique, but I have a large code repository where individual Swift files are compiled (mostly) on their own into standalone static libraries before being linked into larger Rust library builds. If it helps, the organization looks like this:
```
.
├── rust-crate-1/
│   ├── src/
│   │   ├── lib.rs
│   │   └── swift/
│   │       └── Supporting.swift
│   └── Cargo.toml
└── rust-crate-2/
    ├── src/
    │   ├── lib.rs
    │   └── swift/
    │       └── Supporting.Swift
    └── Cargo.toml
```

One of my goals with this PR was to support a structure like this where an IDE such as XCode is unable to work with individual files. It works OK with just macOS today, but we have the need to also have iOS-specific Swift code that uses things like `UIKit`. So, I'm not looking to ever run the file directly, but just to get a smooth editing experience. Does that make sense for the most part? Given that, I don't think a `task.json` would be the correct approach.

src/utilities/utilities.ts
complexspaces: I agree that this is a bit fragile, but I wasn't sure of another way to do this without exposing SDK versioning information to the user as well, something that wouldn't necessary 95% of the time because XCode's SDK uses symlinks.

src/commands.ts
complexspaces: Yeah, that's entirely my bad 😓. I'll rename this based off your suggestion and add some documentation.

src/commands.ts
complexspaces: Good call, thanks! I extracted most of this into a generalized `withQuickPick` that handles the boilerplate.

package.json
complexspaces: TIL, thank you for pointing that out. I moved the macOS gate here based on [the docs](https://code.visualstudio.com/api/references/when-clause-contexts) and removed the `darwin` check from `command.ts` as a result.

src/utilities/utilities.ts
adam-fowler: What do you think about adding/removing the target parameter from the build arguments setting when you choose the platform? It would be clearer to the user what is happening. The current version is adding hidden arguments which might confuse the user. I don't have an issue with exposing the underlying details to the user

src/toolchain/toolchain.ts
adam-fowler: We can give this a miss just now

src/commands.ts
```diff
@@ -428,6 +431,34 @@ function openInExternalEditor(packageNode: PackageNode) {
...
+/**
+ * Switches the target SDK to the platform selected in a QuickPick UI.
+ */
+async function switchPlatform() {
+    const onSelect = async (picked: IDawrinCompatibleTarget) => {
+        const sdkForTarget = await SwiftToolchain.getSdkForTarget(picked.value);
+        if (sdkForTarget) {
+            configuration.sdk = sdkForTarget;
+        } else {
+            console.warn("unable to obtain new SDK path");
```
adam-fowler, line 448: You should probably show an error dialog here, users will not see the console output
```suggestion
            vscode.window.showErrorMessage("Unable to obtain SDK path")
```

package.json
```diff
@@ -44,6 +44,11 @@
...
+        "command": "swift.quickSwitchTarget",
```
adam-fowler, line 48: Given you changed the function name, you might want to do the same to the command and title.

src/utilities/utilities.ts
adam-fowler: If we move to using destination files that'll make this a lot simpler

adam-fowler:
I stole your quickPick in #384. It works really well. Hopefully we can combine the work from both #384 and this PR.

src/toolchain/toolchain.ts
```diff
@@ -205,6 +210,27 @@ export class SwiftToolchain {
...
+        switch (target) {
+            case DarwinCompatibleTarget.macOS:
+                sdkType = "macosx";
+                break;
+            case DarwinCompatibleTarget.iOS:
+                sdkType = "iphoneos";
+                break;
+        }
+
+        const { stdout } = await execFile("xcrun", ["--sdk", sdkType, "--show-sdk-path"]);
```
adam-fowler, line 230: Can you include the following as an argument in the `execFile` call. Means we can pass `DEVELOPER_DIR` to the `xcrun` call and get the sdk from non-default Xcode installs 
```
options: {
    env: {...process.env, ...configuration.swiftEnvironmentVariables},
}
```

package.json
complexspaces: That's my bad, sorry. These should be unified now.

src/toolchain/toolchain.ts
complexspaces: Sure, no problem. That seems reasonable to me. I've added the environment variable forwarding to my changes.

src/commands.ts
complexspaces: That's a better idea, thanks. I forgot how buried away the console is when you aren't actively debugging an extension.

src/commands.ts
```diff
@@ -428,6 +431,34 @@ function openInExternalEditor(packageNode: PackageNode) {
     }
 }
 
+interface IDawrinCompatibleTarget extends vscode.QuickPickItem {
```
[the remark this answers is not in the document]
complexspaces: Thank you for catching that. It's slightly embarrassing that made it through two commits unnoticed 😅 but this now has a more friendly name.

src/toolchain/toolchain.ts
complexspaces: Sorry, I'm not familiar with that phrase 😓. Does "give this a miss" (here and other instances in this PR) mean to not worry about it for now?

adam-fowler: "give this a miss" means "lets not bother". We can look at this later. I know @stevapple is looking at supporting `destination.json` files, which could clean this up.

adam-fowler: "give this a miss" means "lets not bother"

adam-fowler:

adam-fowler (approved):

complexspaces:
Thank you for working with me on this one, I'm happy to see it merged 🎉

adam-fowler:
@complexspaces Although I just noticed after merging, building doesn't work. I know you can't build a complete iOS package, but it is still useful to build.

PR #382, opened 2022-08-01
traverse local depedencies from package.swift

aelam:
1. local/remote/edited dependency has remote/edited dependencies, Package.resolved covers them
2. remote/edited dependency has a local dependency, the local dependency must have been declared in root Package.swift
3. local dependency has a local dependency, traverse it and find the local dependencies only recursively
4. pins include all remote and edited packages for 1, 2

3: requires traverse, 4: reads from `package.resolved`

### DIFF
--- a/src/ui/PackageDependencyProvider.ts
+++ b/src/ui/PackageDependencyProvider.ts
@@ -196,7 +196,7 @@ export class PackageDependenciesProvider implements vscode.TreeDataProvider<Tree
      * Why tranverse is necessary here?
      *  * If we have an implicit local dependency of a dependency, you may not be able to see it in either `Package.swift` or `Package.resolved` unless tranversing from root Package.swift.
      * Why not using `swift package show-dependencies`?
-     *  * it costs more time and it triggers the file change of `Workspace-state.json` which is not necessary
+     *  * it costs more time and it triggers the file change of `workspace-state.json` which is not necessary
      * Why not using `workspace-state.json` directly?
      *  * `workspace-state.json` contains all necessary dependencies but it also contains dependencies that are not in use.
      * Here is the implementation details:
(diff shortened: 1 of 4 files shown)


src/ui/PackageDependencyProvider.ts
```diff
@@ -148,38 +155,130 @@ export class PackageDependenciesProvider implements vscode.TreeDataProvider<Tree
...
+        folderContext: FolderContext
+    ): Promise<Set<string>> {
+        const localDependencies = await this.getAllLocalDependencySet(
+            workspaceState,
+            folderContext
+        );
+        const remoteDependencies = this.getRemoteDependencySet(folderContext);
+
+        // merge the two sets of dependencies
+        const inUseDependencies = new Set<string>();
```
adam-fowler, line 212: This can be simplified to 
```
return new Set<string>([...localDependencies, ...remoteDependencies])
```

src/ui/PackageDependencyProvider.ts
```diff
@@ -148,38 +155,130 @@ export class PackageDependenciesProvider implements vscode.TreeDataProvider<Tree
...
+    }
+
+    private async getDependencyGraph(
+        workspaceState: WorkspaceState | undefined,
+        folderContext: FolderContext
+    ): Promise<PackageNode[]> {
+        if (!workspaceState) {
+            return [];
+        }
+        const inUseDependencies = await this.getInUseDependencies(workspaceState, folderContext);
```
adam-fowler, line 171: Can we add a comment saying why we have to do this

src/ui/PackageDependencyProvider.ts
```diff
@@ -148,38 +155,130 @@ export class PackageDependenciesProvider implements vscode.TreeDataProvider<Tree
...
+        localDependencies.forEach(dependency => inUseDependencies.add(dependency));
+        remoteDependencies.forEach(dependency => inUseDependencies.add(dependency));
+        return inUseDependencies;
+    }
+
+    private getRemoteDependencySet(folderContext: FolderContext | undefined): Set<string> {
+        return new Set<string>(folderContext?.swiftPackage.resolved?.pins.map(pin => pin.identity));
+    }
+
+    private async getAllLocalDependencySet(
```
adam-fowler, line 222: Function headers? You've added them for the other functions

src/ui/PackageDependencyProvider.ts
```diff
@@ -148,38 +155,130 @@ export class PackageDependenciesProvider implements vscode.TreeDataProvider<Tree
...
+    }
+
+    /**
+     * tranverse only local dependency tree
+     * @param dependencies depdencies of current node
+     * @param workspaceStateDependencies all dependencies in workspace-state.json
+     * @param showingDependencies result of dependencies that are showing in the tree
+     * @returns
+     */
+    private async getChildLocalDependencySet(
```
adam-fowler, line 245: Is it worth having `this.getChildLocalDependencySet` return a Set instead of passing a Set in to be filled. When you are parsing children of children you could combine the sets. It makes for slightly cleaner code.

adam-fowler (commented):
In general looks good. Just a few minor comments

src/ui/PackageDependencyProvider.ts
aelam: Maybe converting to a stack version?

src/ui/PackageDependencyProvider.ts
```diff
@@ -148,38 +155,138 @@ export class PackageDependenciesProvider implements vscode.TreeDataProvider<Tree
...
+    /**
+     * * Returns a set of all dependencies that are in use in the workspace.
+     * Why tranverse is necessary here?
+     *  * If we have an implicit local dependency of a dependency, you may not be able to see it in either `Package.swift` or `Package.resolved` unless tranversing from root Package.swift.
+     * Why not using `swift package show-dependencies`?
+     *  * it costs more time and it triggers the file change of `Workspace-state.json` which is not necessary
+     * Why not using `workspace-state.json` directly?
+     *  * `workspace-state.json` contains all necessary dependencies but it also contains dependencies that are not in use.
+     * Here is the implementation details:
+     * 1. local/remote/edited dependency has remote/edited dependencies, Package.resolved covers them
```
aelam, line 203: @adam-fowler 

After I tried more cases, I just found item 1 is wrong. originally I saw edited remote package is  in `Package.resolved` <-- but it's not . perhaps I prepared a wrong test case
the test case is 
1. A depends on B, B depends on C
2. `swift package edit C`, **C will dissapear from `Package.resolved`**

This means we can only check if C is in use by traversing from root or we just show `edited` dependencies out anyway. 

then we have to go back to the original discuss point. 
It seems traversing is the only way to get correct dependencies graph

Currently I added `getEditedDependencySet` which read edited packages from `workspace-state.json`, **but it's still wrong**. because once A is removed from `Package.swift`.  C is still in `workspace-state.json` as `edited`. so it will still be in the list  (deleting state.json could fix the issue)

lemme summarize a bit again 
1. tranverse from root, this is the best result [ no missing items, no extra items, but concern about performance]
2. load dependencies from workspace-state.json [no missing items, but sometimes extra items, no performance issue (deleting state.json could fix the issue)]
3. current PR, traversing local dependencies + reading editing from state.json + remote packages from pin [extra editing packages sometimes (deleting state.json could fix the issue)]

my preference is 1 >= 2 >= 3

What do you think

adam-fowler: 1. This is a real performance issue
2. Yeah workspace-state.json would be the best solution, but it isn't kept up to date. I don't know what the implications of deleting the workspace-state.json every time we resolve is. I have added this issue to the SwiftPM repo https://github.com/apple/swift-package-manager/issues/5727
3. I'm happy to go with this temporarily. It improves on our current situation.

adam-fowler (approved):

src/ui/PackageDependencyProvider.ts
aelam: I also created a PR https://github.com/apple/swift-package-manager/pull/5700/
if it can be accepted, workspace-state.json file could be watched in future

https://github.com/apple/swift-package-manager/issues/5727 I'm not sure if they will treat it as a bug, because they may still treat the unused dependencies as "`ManagedDepdency`",  hopefully, they can fix this issue

PR #384, opened 2022-08-09
Add quick pick to set Xcode DEVELOPER_DIR

adam-fowler:


adam-fowler:
@aelam does this work for you?

aelam:
@adam-fowler  will try it out this week. Thanks for mentioning me !

aelam:
Yes! It would be my honor!

adam-fowler:
> Yes! It would be my honor! 

You need to tell me what it is though

src/commands.ts
```diff
@@ -428,6 +431,50 @@ function openInExternalEditor(packageNode: PackageNode) {
...
+                // if DEVELOPER_DIR was set and the new folder is the default then
+                // delete variable
+                // eslint-disable-next-line @typescript-eslint/no-unused-vars
+                const { DEVELOPER_DIR, ...rest } = swiftEnv;
+                swiftEnv = rest;
+            }
+            configuration.swiftEnvironmentVariables = swiftEnv;
+            // if SDK is inside previous DEVELOPER_DIR then move to new DEVELOPER_DIR
+            if (
+                configuration.sdk.length > 0 &&
```
aelam, line 466: Here is convenient enough 

but I'm wondering if it's possible for the setting of swift.SDK could be like either way
1.  swift.SDK: "SDKs/MacOSX.sdk"  (`DEVELOPER_DIR` is the automatic prefix for it)
2. swift.SDK: "`$DEVELOPER_DIR`/SDKs/MacOSX.sdk".  (environment variable can be used in arguments too)

a bit out of scope here. you may find some ideas here

aelam:
wanglun02@gmail.com

src/commands.ts
adam-fowler: DEVELOPER_DIR doesn't exist on Linux or Windows. Also the SDK can exist outside the DEVELOPER_DIR if you are cross compiling for another platform.

src/toolchain/toolchain.ts
```diff
@@ -86,6 +86,25 @@ export class SwiftToolchain {
...
+    public static async getXcodeDeveloperDir(): Promise<string> {
+        const { stdout } = await execFile("xcode-select", ["-p"]);
+        return stdout.trimEnd();
+    }
+
+    /**
+     * Get list of Xcode versions intalled on mac
+     * @returns Folders for each Xcode install
+     */
+    public static async getXcodeInstalls(): Promise<string[]> {
```
aelam, line 101: [nits] `getInstalledXcodes` may be easier to understand?

src/commands.ts
```diff
@@ -428,6 +431,50 @@ function openInExternalEditor(packageNode: PackageNode) {
...
+ * @param workspaceContext
+ */
+async function selectXcodeDeveloperDir() {
+    const defaultXcode = await SwiftToolchain.getXcodeDeveloperDir();
+    const selectedXcode = configuration.swiftEnvironmentVariables.DEVELOPER_DIR;
+    const xcodes = await SwiftToolchain.getXcodeInstalls();
+    await withQuickPick(
+        selectedXcode ?? defaultXcode,
+        xcodes.map(xcode => {
+            const developerDir = `${xcode}/Contents/Developer`;
```
aelam, line 445: [nits] the UI and the actual value is a bit different
the list in UI is Xcode paths
but the actual values are the developer paths
it's a little bit different, not sure if engineers would get confused

src/commands.ts
aelam: Thanks for your answer

aelam:
I don't see any problems, it's convenient. 

when I check this PR, I found an issue #387 (not related to this one). Please check

0xTim (approved):
LGTM

PR #387, opened 2022-08-11
load taskProviders after folder is ready

aelam:
I noticed the task list came out kinda randomly. 
After I debugged the code, I found `provideTasks` is called in the very beginning before `folderContext` is ready

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -49,18 +49,6 @@ export async function activate(context: vscode.ExtensionContext): Promise<Api> {
         // listen for workspace folder changes and active text editor changes
         workspaceContext.setupEventListeners();
 
-        // Register task provider.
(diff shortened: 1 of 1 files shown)


adam-fowler:
@swift-server-bot please test this

adam-fowler:
> After I debugged the code, I found provideTasks is called in the very beginning before folderContext is ready

This is not strictly true. The task provider is setup which is then called whenever someone brings up the build task menu. I'm not sure what this fixes in the end. When you bring up the build task list before the projects have loaded, it will show nothing.  Once a project has loaded it will show build tasks for that project

aelam:
Sorry, this is a bug of a task explorer, there is no reload, the timing of loading the task list is only in the very beginning.

PR #389, opened 2022-08-15
Extend platform select

adam-fowler:
- Include tvOS and watchOS
- Enabling building of targets using SDKs
- Add warning about undefined results when using build

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -14,7 +14,6 @@
 
 import * as fs from "fs/promises";
 import * as path from "path";
-import { normalize } from "path";
 import * as plist from "plist";
 import * as vscode from "vscode";
 import configuration from "../configuration";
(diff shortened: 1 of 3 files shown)


src/utilities/utilities.ts
```diff
@@ -267,15 +272,17 @@ export function swiftDriverTargetFlags(indirect = false): string[] {
...
-        const version = sdkKind.substring(
-            // Trim the prefix
-            sdkKind.length - IPHONE_SDK_KIND.length,
-            // Trim the `.sdk` suffix
-            sdkKind.length - 4
-        );
-        args = ["-target", `arm64-apple-ios${version}`];
+    for (const target of targetMap) {
+        if (sdkKind.includes(target.name)) {
+            // Obtain the iOS version of the SDK.
```
0xTim, line 277:
```suggestion
            // Obtain the version of the SDK.
```
This is no longer just iOS

src/toolchain/toolchain.ts
```diff
@@ -249,7 +269,7 @@ export class SwiftToolchain {
                     return process.env.SDKROOT;
                 }
 
-                return this.getSdkForTarget(DarwinCompatibleTarget.macOS);
+                return this.getSDKPath("macosx");
```
0xTim, line 272: I'm assuming we're calling through to something here? Otherwise we should rename this to just `macOS`

0xTim (approved):
Couple of nits but overall looks good

src/toolchain/toolchain.ts
adam-fowler: This is the string that is passed to `xcrun` to get the sdk path. We don't have a darwin target macOS because there is no point setting the SDK for macOS as it is the default

PR #392, opened 2022-08-16
enable `edit locally`

aelam:
maybe this can be enabled this time?


adam-fowler:
I'll need to review how robust it is. 
- Does `edit` work if a folder is in the way?
- How well does it work with swift 5.6 vs 5.7 (Sourcekit-lsp 5.6 does some bad things in relation to local edit)

PR #396, opened 2022-08-22
Get correct swift path from swiftenv

adam-fowler:
swiftenv uses a shim shell script to run swift. This fix gets the real swift executable from the shim.

If we don't do this LLDB initialisation can fail, also it allows swiftenv to change the swift version without us knowing


0xTim (approved):
I'm a little uneasy about having tool specific stuff in here - ideally swiftenv would just fix this and expose Swift and LLDB in the path for us to pick up etc. However since it's currently the de facto standard and causing issues lets add it

adam-fowler:
> I'm a little uneasy about having tool specific stuff in here - ideally swiftenv would just fix this and expose Swift and LLDB in the path for us to pick up etc. However since it's currently the de facto standard and causing issues lets add it

I'm not happy with this either, but swiftenv is the de facto way of installing swift on linux at the moment and if we don't fix up the paths at initialisation there are a number of issues that can occur. I don't see swiftenv changing their methods either given they need these shell scripts to support local setting of swift version.

PR #403, opened 2022-08-30
Add right click sub-menu to editor window

adam-fowler:
Adds "Swift" sub-menu to editor right click menu. Entries include
- Run swift script *(when current file is not a snippet)*
- Run swift snippet *(when current file is a snippet)*
- Debug swift snippet *(when current file is a snippet)*
- Clean build *(when we have a package loaded)*

### DIFF
--- a/package.json
+++ b/package.json
@@ -320,7 +320,18 @@
       "swift.editor": [
         {
           "command": "swift.runScript",
-          "group": "1_file"
+          "group": "1_file",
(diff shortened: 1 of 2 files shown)


adam-fowler:
@0xTim Any thoughts on other functions that could go here?

0xTim (approved):
I think this makes sense. Dependency stuff should go in the dependencies view and running apps doesn't make sense in single files

PR #414, opened 2022-09-10
Swift package identity is case-insensitive

stevapple:
This PR fixes the bug of some dependency (eg. `Yams`) not being recognized due to case-sensitive identity comparison.

### DIFF
--- a/assets/test/identity-case/Package.resolved
+++ b/assets/test/identity-case/Package.resolved
@@ -1,0 +1,16 @@
+{
+  "object": {
+    "pins": [
+      {
+        "package": "Yams",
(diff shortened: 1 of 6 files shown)


adam-fowler (commented):
Your test passes without the case insensitive code changes. Do you want to make it actual test your changes. I'd be happy for you to make `PackageDependenciesProvider.getDependencyGraph` public so you can access it from tests.

stevapple:
@adam-fowler I added some assertions — and that’s what we can test now with a dedicated test suite for UI testing. Making `getDependencyGraph` public seems unnecessary for the case, because asserting on the UI (with `getChildren()`) is enough.

adam-fowler (approved):
looks good

PR #435, opened 2022-10-27
Calculate V1 resolved package identity correctly

adam-fowler:
Previous code was just lowercasing the package name, when the identity should be extracted from the URL

### DIFF
--- a/assets/test/identity-different/Package.swift
+++ b/assets/test/identity-different/Package.swift
@@ -1,0 +1,25 @@
+// swift-tools-version:5.4
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "identity-different",
+    products: [
+        // Products define the executables and libraries a package produces, and make them visible to other packages.
+        .library(
+            name: "identity-different",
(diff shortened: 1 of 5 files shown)


0xTim (approved):

stevapple (commented):
Let’s add a test case here to ensure we don’t break it accidentally?

assets/test/identity-different/Package.swift
```diff
@@ -0,0 +1,25 @@
...
+    name: "identity-different",
+    products: [
+        // Products define the executables and libraries a package produces, and make them visible to other packages.
+        .library(
+            name: "identity-different",
+            targets: ["identity-different"]),
+    ],
+    dependencies: [
+        // Dependencies declare other packages that this package depends on.
+        .package(name: "cmark", url: "https://github.com/apple/swift-cmark.git", .branch("main")),
```
stevapple, line 16:
```suggestion
        .package(name: "cmark", url: "https://github.com/apple/swift-cmark.git", .branch("gfm")),
```

The `main` branch is exactly a copy of `cmark`.

stevapple (changes requested):

PR #440, opened 2022-11-01
Read configuration from workspace folder and not workspace

adam-fowler:
In #400 I added the ability to set configuration values on a per folder basis. I didn't realise this wouldn't work automatically.

This PR implements loading configuration settings for each workspace folder. It also removes the ability to set it on a number of settings eg `buildArguments`, because they are used by SourceKit-LSP and it only has one set of arguments.


0xTim (approved):
Looks good to me!

PR #448, opened 2022-12-08
Reorder plugin loading to occur after package resolve

adam-fowler:
Previously part of the package load was to load the package plugins, but this seems to internally do a package resolve which means it can look like package loading takes longer than expected. This PR moves the plugin loading after the package resolve and adds a status item for the plugin loading. Now the order of operations is load, resolve, load plugins. 

This PR also adds helper function StatusItem.showStatusWhileRunning which shows a status item while a task/process is running.


0xTim (approved):

adam-fowler:
CI seems to be failing for an unrelated reason

PR #452, opened 2022-12-15
Cleanup test output parsing

adam-fowler:
Changes are mainly to reduce the size of the parsing functions and remove duplicated code.
Also add parsing of Test Suite messages, although we don't use them yet. 

Was hoping they'd give me a more solid idea of what target was being run as Linux test parsing is lacking some information and I have to make guesses about the test target a test is in. Unfortunately this hasn't helped. But this leaves test output parsing in a better place as I have a better understanding of the parsing state now.

### DIFF
--- a/src/TestExplorer/TestRunner.ts
+++ b/src/TestExplorer/TestRunner.ts
@@ -594,7 +594,7 @@ export class TestRunner {
         runState.suiteStack.pop();
     }
 
-    /** Flag we have start a test */
+    /** Flag we have started a test */
     private startTest(testIndex: number, runState: TestRunState) {
         if (testIndex !== -1) {
             this.testRun.started(this.testItems[testIndex]);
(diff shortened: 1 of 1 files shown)


src/TestExplorer/TestRunner.ts
```diff
@@ -495,79 +494,152 @@ export class TestRunner {
...
+    private passTestSuite(name: string, runState: TestRunState) {
+        runState.suiteStack.pop();
+    }
+
+    /** Flag a test suite has failed */
+    private failTestSuite(name: string, runState: TestRunState) {
+        runState.suiteStack.pop();
+    }
+
+    /** Flag we have start a test */
```
0xTim, line 597:
```suggestion
    /** Flag we have started a test */
```
Nit

0xTim (approved):
LGTM - one minor nit

PR #454, opened 2022-12-16
Add support for CMake projects

adam-fowler:
Triggers extension when a compile_commands.json exists at root of workspace. Because there is no package.swift in the folder most of the extension will do nothing, but it will start the LSP server so code completion will be available.

Also look for CMake projects in the subfolders of workspaces.

### DIFF
--- a/src/WorkspaceContext.ts
+++ b/src/WorkspaceContext.ts
@@ -277,8 +277,8 @@ export class WorkspaceContext implements vscode.Disposable {
      * @param folder folder being added
      */
     async addWorkspaceFolder(folder: vscode.WorkspaceFolder) {
-        // add folder if Package.swift exists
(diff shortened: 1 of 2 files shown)


adam-fowler:
See https://forums.swift.org/t/vs-code-swift-extension-and-cmake/62086

src/WorkspaceContext.ts
```diff
@@ -495,6 +495,17 @@ export class WorkspaceContext implements vscode.Disposable {
         }
     }
 
+    /**
+     * Return is folder is considered a valid root folder ie does it contain a SwiftPM
```
0xTim, line 499: Think this is a typo 
```suggestion
     * Return if folder is considered a valid root folder ie does it contain a SwiftPM
```

0xTim (approved):
One typo but LGTM

PR #461, opened 2023-01-04
Fix parsing of mac test output

adam-fowler:
Now we use `swift test` XCT output is sent to stdout


0xTim (commented):
The PR title and body don't appear to match the changes?

adam-fowler:
Sorry looks like I messed up the branch

PR #467, opened 2023-01-08
Attempt to fix CI issues

adam-fowler:
This improves the stability of tests, but they still fail occasionally
Also added a test to check CMake projects work


assets/test/cmake/compile_commands.json
```diff
@@ -0,0 +1,7 @@
+[
+{
+  "directory": "/Users/adamfowler/Developer/tools/vscode-swift/assets/test/cmake",
```
0xTim, line 3: Do we want these committed?

0xTim (commented):
Query about file with user paths in, otherwise LGTM

assets/test/cmake/compile_commands.json
adam-fowler: We need the compile_commands.json to trigger the extension for CMake projects. I'll edit it though

adam-fowler: I've removed most of the cmake project files as they were not needed

adam-fowler:

0xTim (approved):

PR #486, opened 2023-02-04
Workaround vscode task execution bug

adam-fowler:
If two tasks with same name but different scopes are executed at the same time the second one does not run. This showed up as a bug in the test explorer when you have more than two workspaces folders both with tests. Both packages would instigate a build but the second build would never run. The fix is to add a promise to wait until the first task has actually stated then the second one will start. 🤷 

I've also added a bunch of tests for the TaskManager and TaskQueue

### DIFF
--- a/assets/test/sleep.sh
+++ b/assets/test/sleep.sh
@@ -1,0 +1,2 @@
+sleep $1
+exit $2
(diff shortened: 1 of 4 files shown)


adam-fowler:
Tests are failing again. Bloody Jenkins seems to fail to setup X11 server correctly for VSCode to run .

0xTim (approved):

PR #487, opened 2023-02-05
Use ProcessExecution for swift tasks

adam-fowler:
Use ProcessExecution instead of ShellExecution. There is no need to setup a new shell to run each swift process.

### DIFF
--- a/src/SwiftTaskProvider.ts
+++ b/src/SwiftTaskProvider.ts
@@ -122,7 +122,7 @@ export async function getBuildAllTask(folderContext: FolderContext): Promise<vsc
     let task = tasks.find(
         task =>
             task.name === `swift: ${buildTaskName}` &&
-            (task.execution as vscode.ShellExecution).options?.cwd ===
+            (task.execution as vscode.ProcessExecution).options?.cwd ===
                 folderContext.folder.fsPath &&
             task.source === "Workspace"
     );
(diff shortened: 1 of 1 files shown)


compnerd (commented):
The only question this raises for me is, do we have a guarantee that `swift` is always going to be an absolute path?  By switching to `ProcessExecution`, we do gain efficiency but it means that we need to build the commands without any wildcards to expand and using absolute paths.  We already controlled the environment explicitly, so we shouldn't have to worry about that.

adam-fowler:
> The only question this raises for me is, do we have a guarantee that `swift` is always going to be an absolute path? By switching to `ProcessExecution`, we do gain efficiency but it means that we need to build the commands without any wildcards to expand and using absolute paths. We already controlled the environment explicitly, so we shouldn't have to worry about that.

The swift executable will either be found in the $PATH or has been set to an absolute path. We currently don't build any commands with wildcards.

compnerd:
> The swift executable will either be found in the $PATH or has been set to an absolute path. We currently don't build any commands with wildcards.

This is a problem.  AIUI, `ProcessExecution` will not honour `PATH`, it needs the absolute path to invoke.  We should be able to locate it once and cache that value and use that for the subsequent invocations.

adam-fowler:
This doesn't appear to be the case on macOS. I'm running with just `swift` and it works fine.

compnerd:
Oh, nice, in that case, LGTM!

compnerd (approved):

0xTim (approved):

PR #506, opened 2023-03-16
Experimental: Remove hard CodeLLDB dependency

adam-fowler:
Add a dialog that is displayed if the CodeLLDB dependency is not available. You can install the dependency from this dialog, set it to never appear again or just ignore it. This actually works out quite nicely.

Problems occur if you have the dependency installed but disabled. It cannot recognise this. Also if you go into a devcontainer without CodeLLDB but have it installed locally you cannot install it into the devcontainer. You need the user to go do it manually. Not too much of a problem if I could recognise you were in a container.

### DIFF
--- a/package.json
+++ b/package.json
@@ -328,6 +328,12 @@
             "description": "The path of the SDK to compile against (`--sdk` parameter). The default SDK is determined by the environment on macOS and Windows.",
             "order": 3
           },
+          "swift.skipCodeLLDBCheck": {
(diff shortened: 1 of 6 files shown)


adam-fowler:
If you want to test this you need to uninstall the released version of the swift extension otherwise you can't uninstall CodeLLDB

award999:
This work has been done in another PR :)

PR #510, opened 2023-03-20
Implemented GoToTestImplementationFile command

rizumita:
This PR implements the 'Go to Test or Implementation File' command.

This command provides the features
- Go to the implementation file if the test file is active.
- Go to the test file if the implementation file is active.
- Prompt to create a test or implementation file if it does not exist.


adam-fowler:
Hi, I'm sorry but I don't think we can accept this PR. I would like to keep the scope of this extension concentrated on core swift tools. I believe this belongs in a separate extension. It also imposes a certain structure (ie tests for `MySource.swift` are in `MySourceTests.swift`) to everyone's package which might not be how everyone works.

Issue #515, opened 2023-04-03
Activating Swift extension failed: Cannot read properties of null (reading 'nodeName')


compnerd:
**Describe the bug**
Activating the plugin fails on VSCode 1.77.0 with a local build of the toolchain.

**To Reproduce**
1. Build the Swift toolchain from sources via `build.ps1` in https://github.com/compnerd/swift-build
2. Setup the environment for development as per the docs in the repository.
3. Launch VSCode in a Swift repository (e.g. swift-package-manager)

**Expected behavior**
I expect that the extension actually activates properly (unclear if it would function properly, but that is what I wanted to verify).

**Environment**
 - OS: Windows 11
 - Swift version (output of `swift --version`)
   ```
   Swift version 5.9-dev (LLVM c3bcac25acd489f, Swift 18971c17846861c)
   Target: x86_64-unknown-windows-msvc
   ```
 - Visual Studio Code version: 1.77.0
 - vscode-swift version: 1.2.0

**Additional context**

compnerd:
![image](https://user-images.githubusercontent.com/63311/229589207-b7ff4055-934f-43aa-97a0-56cd819de113.png)

compnerd:
This is occurring due to `SwiftToolchain.getXCTestPath` failing.  In particular, the `plist.parse` failing due to the file format encoding (UTF16-LE/DOS LE).

PR #516, opened 2023-04-03
toolchain: ensure that we catch decoding failures

compnerd:
In the case the `Info.plist` is encoded improperly (e.g. UTF16-LE) we would fail to parse the file and then error out.  Catch the error and surface it properly for the user to follow along.

Fixes: #515

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -368,7 +368,15 @@ export class SwiftToolchain {
                     return undefined;
                 }
                 const data = await fs.readFile(platformManifest, "utf8");
-                const infoPlist = plist.parse(data) as unknown as InfoPlist;
+                let infoPlist;
+                try {
+                    infoPlist = plist.parse(data) as unknown as InfoPlist;
+                } catch (error) {
+                    vscode.window.showWarningMessage(
+                        `unable to parse ${platformManifest}: ${error}`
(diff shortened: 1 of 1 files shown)


adam-fowler:
Was this the issue you were seeing, a few weeks back?

compnerd:
@adam-fowler yes, it is!

src/toolchain/toolchain.ts
```diff
@@ -368,7 +368,15 @@ export class SwiftToolchain {
                     return undefined;
                 }
                 const data = await fs.readFile(platformManifest, "utf8");
-                const infoPlist = plist.parse(data) as unknown as InfoPlist;
+                let infoPlist;
+                try {
+                    infoPlist = plist.parse(data) as unknown as InfoPlist;
+                } catch (error) {
+                    vscode.window.showWarningMessage(
+                        `unable to parse ${platformManifest}: ${error}`
```
adam-fowler, line 376: Is there any suggestion we can make to help them fix the issue?

compnerd: I don't really know 😢 I think if we were able to detect the encoding that would be good.  However, this is a file that the user doesn't write, it is a distributed file, so its really a bad release.

adam-fowler (approved):
LGTM

PR #517, opened 2023-04-09
Run xctest directly if not running code coverage

adam-fowler:
Running via `swift test` means log output is not being received in a timely manner

Also send different kill signals if running tests via swift test or xctest


0xTim (approved):
LGTM

PR #520, opened 2023-04-11
Release v1.2.1

adam-fowler:


CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## 1.2.1 - 2023-04-12
```
0xTim, line 3:
```suggestion
## 1.2.1 - 2023-04-13
```

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+
+### Changed
+
+- Run `xctest` directly when running tests instead of via `swift test`.
```
0xTim, line 7: Is it worth mentioning that this is on macOS only?

0xTim (approved):

CHANGELOG.md
adam-fowler: Ah maybe I worded that wrongly. Maybe it should be `test executable` as it runs the built test executable on linux and windows and `xctest` on macOS which could be considered the `test executable`. The basic change is don't use `swift test` where we can avoid it.

PR #521, opened 2023-04-11
If LSP server shutdown times out still restart it

adam-fowler:
This fixes an issue on my Windows image where the LSP server never restarts

### DIFF
--- a/src/sourcekit-lsp/LanguageClientManager.ts
+++ b/src/sourcekit-lsp/LanguageClientManager.ts
@@ -353,7 +353,10 @@ export class LanguageClientManager {
                 .then(async () => {
                     await this.setupLanguageClient(workspaceFolder?.uri);
                 })
-                .catch(reason => {
+                .catch(async reason => {
+                    if (reason.message === "Stopping the server timed out") {
+                        await this.setupLanguageClient(workspaceFolder?.uri);
+                    }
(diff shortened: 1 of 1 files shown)


src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -349,7 +349,10 @@ export class LanguageClientManager {
                 .then(async () => {
                     await this.setupLanguageClient(workspaceFolder?.uri);
                 })
-                .catch(reason => {
+                .catch(async reason => {
+                    if (reason.message === "Stopping the server timed out") {
+                        await this.setupLanguageClient(workspaceFolder?.uri);
+                    }
```
compnerd, line 355: I don't follow how this guarantees that the "old" client is terminated.  It appears to simply spawn the new instance in the case of a timeout or not.  Am I missing something?

adam-fowler: I've found that VSCode is not registering that the old server has been killed even though it has. It is a bug in VSCode language client really.

tristanlabelle (approved):
I tried this fix locally and it seems good. I also don't end up up with multiple `sourcekit-lsp.exe` processes. Thanks for fixing this!

adam-fowler:

0xTim (approved):

PR #528, opened 2023-04-24
Add defaults settings for commonly used command plugin

adam-fowler:
Add defaults for lambda-runtime package, docc generate and preview, SwiftFormat
I didn't add any for swift-format as it requires you to add project specific arguments
@0xTim if you can think of any others that'd be great


0xTim (approved):
Nothing off the top of my head - Vapor may be introducing some for migrations etc but nothing yet

PR #531, opened 2023-04-30
BuildFlags and --scratch-path

adam-fowler:
This is rather a large PR for a very small change. In Swift 5.7 and later you should use `--scratch-path` instead of `--build-path`. Unfortunately previously I had no way of knowing what version of swift I was running when constructing the build path argument. So I moved all the build argument functions to a new class `BuildFlags` which is created by `SwiftToolchain` and has a reference to the toolchain, so it knows what version of swift we are using. 

This means I have to pass the `SwiftToolchain` about a lot more.


src/toolchain/BuildFlags.ts
```diff
@@ -0,0 +1,199 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VSCode Swift open source project
+//
+// Copyright (c) 2021 the VSCode Swift project authors
```
0xTim, line 5:
```suggestion
// Copyright (c) 2023 the VSCode Swift project authors
```

test/suite/BuildFlags.test.ts
```diff
@@ -0,0 +1,146 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VSCode Swift open source project
+//
+// Copyright (c) 2021-2022 the VSCode Swift project authors
```
0xTim, line 5:
```suggestion
// Copyright (c) 2023 the VSCode Swift project authors
```

0xTim (approved):
Couple of years that should be updated otherwise LGTM

PR #538, opened 2023-05-04
Management of swiftinterface files

adam-fowler:
Currently if you right click on the `import module` text you can jump to a cleaned up version of its swift interface file with documentation. Eventually I hope SourceKit-LSP will be able to do this for any symbol from a framework we don't have source code for. 

In the meantime this PR helps this process by first adding a readonly document provider and using this when displaying swift interface files. A middleware is added to the LSP client to redirect the swiftinterface files to the read-only document provider. This means of course the user can't edit the files but also has the added bonus that sourcekit-lsp will not try to compile them (and most likely complain). Secondly I have also associated the extension `.swiftinterface` with the swift language so we get proper highlighting.


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.3.0 - 2023-05-02
+## 1.3.0 - 2023-05-03
```
adam-fowler, line 3: This PR has caught my date update for the CHANGELOG

src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -209,17 +211,23 @@ export class LanguageClientManager {
      * @returns result of process
      */
     async useLanguageClient<Return>(process: {
-        (client: langclient.LanguageClient, cancellationToken: vscode.CancellationToken): Return;
-    }) {
-        if (!this.languageClient) {
+        (
```
adam-fowler, line 214: I cleaned up `useLanguageClient` a little because I was originally using this in the PR, but I'm not anymore. But the clean up is still valid I feel.

0xTim (approved):

PR #551, opened 2023-06-11
 Add configurable language support to mitigate conflict with clangd extension

shulong-apple:
Body:

In this Pull Request, I have added a new configuration option sourcekit-lsp.support_languages. This configuration allows users to specify the list of languages that the sourcekit-lsp extension should support.

The primary reason for this change is to resolve the conflict between the sourcekit-lsp and clangd extensions. The original cpptools-inactive option only considered the ms-vscode.cpptools extension. However, in a mixed Swift and Objective-C++ development environment, the Language Server Protocol (LSP) can often encounter issues, and clangd tends to be more stable than cpptools in these scenarios.

This update allows users to leverage clangd for Objective-C and Objective-C++ development while continuing to use sourcekit-lsp for Swift. It provides more flexibility for developers working in a mixed language environment and should improve the overall stability of the extensions.

Please review the changes and let me know your thoughts. Thank you.

### DIFF
--- a/package.json
+++ b/package.json
@@ -266,6 +266,7 @@
               "Disable when C/C++ extension is active"
             ],
             "description": "Add LSP functionality for C/C++ files. By default this is set to disable when the C/C++ extension is active.",
+            "markdownDeprecationMessage": "**Deprecated**: Please use `#sourcekit-lsp.supported-languages#` instead.",
             "order": 4
           },
           "sourcekit-lsp.trace.server": {
@@ -284,6 +285,28 @@
             "default": false,
             "description": "Disable the running of SourceKit-LSP.",
             "order": 6
+          },
+          "sourcekit-lsp.supported-languages": {
+            "type": "array",
+            "default": [
+              "swift",
+              "objective-c",
+              "objective-cpp",
+              "c",
+              "cpp"
+            ],
+            "description": "List of languages supported by SourceKit-LSP. This is used to determine when to start the SourceKit-LSP server.",
+            "items": {
+              "type": "string",
+              "enum": [
+                "swift",
+                "objective-c",
+                "objective-cpp",
+                "c",
+                "cpp"
+              ]
+            },
+            "order": 7
           }
         }
       },
--- a/package.json
+++ b/package.json
@@ -252,6 +252,28 @@
             "markdownDeprecationMessage": "**Deprecated**: Please use `#editor.inlayHints.enabled#` instead.",
             "order": 3
           },
+          "sourcekit-lsp.supported-languages": {
+            "type": "array",
+            "default": [
+              "swift",
+              "objective-c",
+              "objective-cpp",
+              "c",
+              "cpp"
+            ],
+            "description": "List of languages supported by SourceKit-LSP. This is used to determine whether SourceKit-LSP should provide language features for a particular file type.",
+            "items": {
+              "type": "string",
+              "enum": [
+                "swift",
+                "objective-c",
+                "objective-cpp",
+                "c",
+                "cpp"
+              ]
+            },
+            "order": 4
+          },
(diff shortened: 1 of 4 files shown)


adam-fowler:
SourceKit-LSP uses clangd in the background already. Do you get different results from running with only clangd extension and only sourcekit-lsp?

shulong-apple:
> SourceKit-LSP uses clangd in the background already. Do you get different results from running with only clangd extension and only sourcekit-lsp?

image

"Thank you for your response. I'm using the clangd extension (https://github.com/clangd/vscode-clangd) for C, C++, and Objective-C code completion. However, if I enable both the clangd and SSWG (Swift Server Work Group) extensions at the same time, I encounter the error mentioned earlier. In fact, the whole completion process generates errors. Similar issues have been reported here: https://www.lunarvim.org/docs/features/supported-languages/swift. I believe my pull request would help mitigate such conflicts and improve the overall stability for developers working in mixed language environments. I'm open to further discussion and willing to make necessary modifications based on your feedback."

shulong-apple:
"Additionally, I feel that the previous option support-c-cpp can be fully replaced by the new option introduced in this pull request. Apart from the ms-cpp tool extension, there are numerous other C and C++ extensions for which potential conflicts are unpredictable. Therefore, providing a configurable language support option can give users more control to manage their extensions and mitigate conflicts effectively."

package.json
```diff
@@ -252,6 +252,26 @@
             "markdownDeprecationMessage": "**Deprecated**: Please use `#editor.inlayHints.enabled#` instead.",
             "order": 3
           },
+          "sourcekit-lsp.supported-languages": {
+            "type": "array",
```
adam-fowler, line 256: Can you add a description please

package.json
```diff
@@ -252,6 +252,26 @@
...
+              "type": "string",
+              "enum": [
+                "swift",
+                "objective-c",
+                "objective-cpp",
+                "c",
+                "cpp"
+              ]
+            }
+          },
```
adam-fowler, line 274: Add `order` parameter to ensure the ordering of the setting.

package.json
```diff
@@ -252,6 +252,26 @@
...
           "sourcekit-lsp.support-c-cpp": {
             "type": "string",
             "default": "cpptools-inactive",
```
adam-fowler, line 277: Add a deprecation message to support-c-cpp setting. See sourcekit-lsp.inlayHints.enabled for how to do this.

shulong-apple:
Thank you for your feedback. I have updated the PR to include a description and made the changes you suggested

package.json
```diff
@@ -284,6 +285,28 @@
...
+              "type": "string",
+              "enum": [
+                "swift",
+                "objective-c",
+                "objective-cpp",
+                "c",
+                "cpp"
+              ]
+            },
+            "order": 7
```
adam-fowler, line 309: Can we change the order so this appears just before `sourcekit-lsp.support-c-cpp`

package.json
```diff
@@ -284,6 +285,28 @@
...
+          "sourcekit-lsp.supported-languages": {
+            "type": "array",
+            "default": [
+              "swift",
+              "objective-c",
+              "objective-cpp",
+              "c",
+              "cpp"
+            ],
+            "description": "List of languages supported by SourceKit-LSP. This is used to determine when to start the SourceKit-LSP server.",
```
adam-fowler, line 298: This isn't completely accurate. Maybe...
```suggestion
            "description": "List of languages supported by SourceKit-LSP. This is used to determine whether SourceKit-LSP should provide language features for a particular file type.",
```

adam-fowler (commented):
A couple of minor changes and I think we are good

adam-fowler (approved):
LGTM

0xTim (approved):

PR #563, opened 2023-07-03
Update problemMatcher to match Swift errors without a column number

tristanlabelle:
Compiler errors like this one didn't show up in the VS Code Problems pane:
```
<unknown>:0: error: cannot open file 'foobar.h': no such file or directory
```

This is because the `swiftc` problem matcher requires a line and a column number. This change updates the regex to support the lack of a column number, so that the error above will show up in the Problems pane. It shows up under a dummy file called `<unknown>`, but that's probably the best we can do with the VS Code problemsMatcher API. I tried making the first matching group not match when finding `<unknown>` but VS Code really expects something in the file group.

How tested: used a custom build task that prints error-like lines:

image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -542,7 +542,7 @@
         "fileLocation": "absolute",
         "pattern": [
           {
-            "regexp": "^(.*):(\\d+):(\\d+):\\s+(warning|error|note):\\s+(.*)$",
+            "regexp": "^(.*?):(\\d+)(?::(\\d+))?:\\s+(warning|error|note):\\s+(.*)$",
             "file": 1,
             "line": 2,
             "column": 3,


adam-fowler:
Where are errors in this format coming from? Not having a source file in the error makes it very hard to fix as there is no link to the error. I guess this is one degree better than not listing the error at all.

tristanlabelle:
This was from a `Packages.swift` building a C module, and the build was in a bad state so the file was missing. The reason why I think this fix is important even if it's not ideal to have a dummy file path is that our engineers are stopping to use the Problems tab in VS Code and losing trust in it because they can only reliably see all of their build errors by looking at the compiler output due to cases like this one.

I think the error is emitted from llvm? There's at least one Swift test that expects this format: https://github.com/apple/swift/blob/e12d53ecd2c6214cd5c129af9fab3f5650f923ee/test/Frontend/bad-vfs-overlays.swift#L4

tristanlabelle:
@adam-fowler Is this signoffable? I have an alternate way to fix this by creating a new VS Code extension defining an additional problem matcher and using it in our local builds on top of `$swiftc`, but since the error here comes from swiftc/LLVM, I think it makes more sense to have it in the Swift extension.

adam-fowler:
Sorry I've been away most of this week, so haven't had time to look into this.

PR #575, opened 2023-07-28
Allow getBuildAllTask() to return non-swift task

adam-fowler:
This allows the default build task to be something other than a swift process eg CMake.
The default build task is triggered by the test explorer and the background compilation

### DIFF
--- a/src/SwiftTaskProvider.ts
+++ b/src/SwiftTaskProvider.ts
@@ -133,20 +133,30 @@ export async function getBuildAllTask(folderContext: FolderContext): Promise<vsc
         buildTaskName += ` (${folderContext.relativePath})`;
     }
 
-    // search for build all task in task.json first
-    const workspaceTasks = await vscode.tasks.fetchTasks();
-    let task = workspaceTasks.find(task => {
-        if (task.source !== "Workspace" || task.name !== `swift: ${buildTaskName}`) {
+    const folderWorkingDir = folderContext.workspaceFolder.uri.fsPath;
+    // search for build all task in task.json first, that are valid for folder
+    const workspaceTasks = (await vscode.tasks.fetchTasks()).filter(task => {
+        if (task.source !== "Workspace" || task.scope !== folderContext.workspaceFolder) {
             return false;
         }
         const processExecutionOptions = (task.execution as vscode.ProcessExecution).options;
         const shellExecutionOptions = (task.execution as vscode.ShellExecution).options;
         let cwd = processExecutionOptions?.cwd ?? shellExecutionOptions?.cwd;
-        if (cwd === "${workspaceFolder}") {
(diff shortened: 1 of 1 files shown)


0xTim (approved):

src/SwiftTaskProvider.ts
```diff
@@ -128,26 +128,31 @@ export function createBuildAllTask(folderContext: FolderContext): vscode.Task {
...
-    let task = tasks.find(
-        task =>
-            task.name === `swift: ${buildTaskName}` &&
-            (task.execution as vscode.ProcessExecution).options?.cwd ===
-                folderContext.folder.fsPath &&
-            task.source === "Workspace"
-    );
+    const workspaceTasks = await vscode.tasks.fetchTasks();
+    let task = workspaceTasks.find(task => {
+        if (task.source !== "Workspace" || task.name !== `swift: ${buildTaskName}`) {
```
tristanlabelle, line 139: It looks like that won't work if our default build task is:

```json
		{
			"type": "cmake",
			"label": "App Build",
			"command": "build",
			"targets": [
				"all"
			],
			"problemMatcher": "$swiftc",
			"group": {
				"kind": "build",
				"isDefault": true
			},
			"detail": "Build App for selected configuration",
			"dependsOn": "Configure"
		},
```

adam-fowler: What, because the label is different?

adam-fowler: @tristanlabelle ?

fischman-bcny: @adam-fowler: @tristanlabelle is out for the week but yeah that's what I think he meant.
Is it not correct to search for a first kind:build/isDefault:true task?

adam-fowler: As I mentioned in the related issue, workspaces can have multiple packages in them. The default build task for a workspace might not compile the correct package. But I was thinking about and I can probably compare working directories as well to ensure a default task is the correct task for a package

adam-fowler: Maybe this change will work

tristanlabelle (approved):
I think that'll work!

src/SwiftTaskProvider.ts
tristanlabelle: Sorry for the wait! I think the latest will work. Thanks!

tristanlabelle:
I also now verified that it does work. Thanks!

tristanlabelle:
When might we see a new release that includes this fix?

adam-fowler:
Was thinking of doing one today/tomorrow. There are a couple of other things I'd like to get out

adam-fowler:
And it is released (1.5)

PR #588, opened 2023-08-22
Added debug adapter tracker to catch logging from debugger

adam-fowler:
While messing about with the llvm lldb-vscode DAP I found the `DebugAdapterTracker` class which allows us to catch every message going between VSCode and the debugger. This PR adds one to catch all the output to stdout and stderr by the debuggee and passes it to the test runner. It cleans up the last issue I really had with debugging tests where the output from tests didn't appear until you had finished running the tests.

### DIFF
--- a/src/debugger/logTracker.ts
+++ b/src/debugger/logTracker.ts
@@ -1,0 +1,84 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VSCode Swift open source project
+//
+// Copyright (c) 2023 the VSCode Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of VSCode Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import * as vscode from "vscode";
+
+/**
+ * Factory class for building LoggingDebugAdapterTracker
+ */
+export class LoggingDebugAdapterTrackerFactory implements vscode.DebugAdapterTrackerFactory {
+    createDebugAdapterTracker(
+        session: vscode.DebugSession
+    ): vscode.ProviderResult<vscode.DebugAdapterTracker> {
+        return new LoggingDebugAdapterTracker(session.id);
+    }
+}
+
+interface iOutputEventBody {
+    category: string;
+    output: string;
+}
+
+interface iDebugMessage {
+    seq: number;
+    type: string;
+    event: string;
+    body: iOutputEventBody;
+}
+
+/**
+ * Debug Adapter tracker that track debuggee output to stdout and stderr and returns it
+ */
+export class LoggingDebugAdapterTracker implements vscode.DebugAdapterTracker {
+    // keep a track of the logging debug trackers, so we can set the callback later on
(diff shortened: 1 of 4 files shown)


src/debugger/logTracker.ts
```diff
@@ -0,0 +1,84 @@
...
+        return new LoggingDebugAdapterTracker(session.id);
+    }
+}
+
+interface iOutputEventBody {
+    category: string;
+    output: string;
+}
+
+interface iDebugMessage {
```
compnerd, line 33: Is the leading I for interface here intentionally lowercase?

src/debugger/logTracker.ts
```diff
@@ -0,0 +1,84 @@
...
+    // keep a track of the logging debug trackers, so we can set the callback later on
+    private static debugSessionIdMap: { [id: string]: LoggingDebugAdapterTracker } = {};
+
+    private cb?: (output: string) => void;
+
+    constructor(public id: string) {
+        LoggingDebugAdapterTracker.debugSessionIdMap[id] = this;
+    }
+
+    static addDebugSessionCallback(session: vscode.DebugSession, cb: (log: string) => void) {
```
compnerd, line 53: set might be better than add as the previous callback will be dropped.

src/debugger/logTracker.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * The debug adapter has sent a Debug Adapter Protocol message to the editor. Check
+     * it is a output message and is not being sent to the console
+     */
+    onDidSendMessage(message: unknown): void {
+        const debugMessage = message as iDebugMessage;
+        if (
+            this.cb &&
```
compnerd, line 67: This can be handled by optional chaining I believe:
`this.cb?(…)`

adam-fowler: For some reason it didn't like it. Anyway I'd like to do the quick check that a callback has been set before doing all the string compares later on

src/debugger/logTracker.ts
adam-fowler: Never been really sure how to name interfaces in JS. I'm not a great fan of the `i` prefix but have seen it quite a bit. I might just get rid of the prefix.

src/debugger/logTracker.ts
compnerd: Oh, if it doesn't like it, then this is fine.  I could've sworn TS supported that syntax 🤷‍♂️

src/debugger/logTracker.ts
compnerd: That is common on Windows, but it tends to use `I` instead of `i`.

compnerd (approved):

src/debugger/logTracker.ts
```diff
@@ -0,0 +1,84 @@
...
+
+interface DebugMessage {
+    seq: number;
+    type: string;
+    event: string;
+    body: OutputEventBody;
+}
+
+/**
+ * Debug Adapter tracker that track debuggee output to stdout and stderr and returns it
```
0xTim, line 41:
```suggestion
 * Debug Adapter tracker that tracks debugger output to stdout and stderr and returns it
```

0xTim (approved):
Looks fine to me, one typo nit

PR #589, opened 2023-08-23
Use SwiftToolchain.getToolchainExecutable

adam-fowler:
Instead of getSwiftExecutable

This is cleaner, will return the actual executable path rather than the one in $PATH (/usr/bin/swift on macOS)


0xTim (approved):
Not a massive fan of replacing function calls with calls passing a string but no major objections

adam-fowler:
The original function had a default argument of "swift" so pretty much the same.

PR #590, opened 2023-08-24
Fix custom Swift path on Windows

stevapple:
The extension fails to find Swift on Windows due to missing `.exe` suffix.


adam-fowler (approved):
LGTM

adam-fowler:
Ah it wasn't good. You need to use backticks. Did you not test this?

stevapple:
My bad:( I worked around locally and submitted the patch directly through GitHub website…

PR #591, opened 2023-08-24
Use lldb-vscode debug adapter from Swift Toolchain

adam-fowler:
This is only available in the Windows toolchain at the moment. If you want to use it on other platforms you need to build a toolchain yourself and extract the lldb-vscode executable out of the build folder and copy into the bin folder alongside swift.

This is not enabled by default. You have to go into the settings and select `swift.debugger.useDebugAdapterFromToolchain`

### DIFF
--- a/package.json
+++ b/package.json
@@ -707,9 +707,6 @@
     },
     "breakpoints": [
       {
-        "language": "arm"
-      },
-      {
         "language": "asm"
       },
       {
@@ -734,7 +731,7 @@
     "debuggers": [
       {
         "type": "swift-lldb",
-        "label": "Native Swift LLDB Debugger",
+        "label": "Swift LLDB Debugger",
         "configurationAttributes": {
           "launch": {
             "required": [
--- a/package.json
+++ b/package.json
@@ -759,9 +759,14 @@
                 "default": "${workspaceRoot}"
               },
               "env": {
-                "type": "array",
-                "description": "Additional environment variables to set when launching the program. This is an array of strings that contains the variable name followed by an optional '=' character and the environment variable's value. Example:  [\"FOO=BAR\", \"BAZ\"]",
-                "default": []
+                "type": "object",
+                "patternProperties": {
+                  ".*": {
+                    "type": "string"
+                  }
+                },
+                "description": "Additional environment variables to set when launching the program.",
+                "default": {}
               },
               "stopOnEntry": {
                 "type": "boolean",
--- a/package.json
+++ b/package.json
@@ -405,13 +405,19 @@
         }
       },
       {
-        "title": "Debugger",
+        "title": "Debugger (Windows)",
         "properties": {
           "swift.debugger.useDebugAdapterFromToolchain": {
             "type": "boolean",
             "default": false,
             "description": "Use lldb-vscode packaged with Swift toolchain as your debug adapter. This is currently only available on Windows platforms",
             "order": 1
+          },
+          "swift.debugger.path": {
+            "type": "string",
+            "default": "",
+            "description": "Path to lldb-vscode debug adapter.",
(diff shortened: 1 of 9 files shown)


adam-fowler:
@compnerd

package.json
```diff
@@ -685,7 +697,239 @@
           "when": "swift.hasPackage"
         }
       ]
-    }
+    },
+    "breakpoints": [
+      {
+        "language": "arm"
```
compnerd, line 703: Seems like a typo?  I suspect it was copying the in-tree plugin, which lists Ada (which IMO we shouldn't bother with as I'm not sure if its ever been tested).

package.json
```diff
@@ -685,7 +697,239 @@
...
+        "language": "rust"
+      },
+      {
+        "language": "swift"
+      }
+    ],
+    "debuggers": [
+      {
+        "type": "swift-lldb",
+        "label": "Native Swift LLDB Debugger",
```
compnerd, line 730: I'm not sure about the title, it seems a bit confusing.

src/WorkspaceContext.ts
```diff
@@ -136,6 +137,30 @@ export class WorkspaceContext implements vscode.Disposable {
                         }
                     });
             }
+            // on change of swift debugger type
+            if (event.affectsConfiguration("swift.debugger.useDebugAdapterFromToolchain")) {
+                if (configuration.debugger.useDebugAdapterFromToolchain) {
+                    if (!(await verifyDebugAdapterExists(this.toolchain))) {
```
compnerd, line 143: Might be nice to have a logged event to explain that the DAP was not found.

src/debugger/debugAdapterFactory.ts
```diff
@@ -0,0 +1,98 @@
...
+    );
+    return {
+        dispose: () => {
+            debugConfigProvider.dispose();
+            debugAdpaterFactory.dispose();
+        },
+    };
+}
+
+async function isFileExists(path: string): Promise<boolean> {
```
compnerd, line 61: This is a weird name "is file exists".

compnerd (approved):

src/debugger/debugAdapterFactory.ts
adam-fowler: Haha copy and paste from Stackoverflow. I'll give it a better name.

src/WorkspaceContext.ts
adam-fowler: This needs cleaned up a little. I can add a log event. Would be useful if I could disable settings under certain situations but VSCode doesn't seem to allow this.

package.json
adam-fowler: I'll get rid of the native. I had just added Swift to the original plugin name used

package.json
adam-fowler: Again just copied from the original package.json

compnerd:
Would there be a way to make CodeLLDB optionally installed so if we don't use it we don't have to have it installed?

adam-fowler:
> Would there be a way to make CodeLLDB optionally installed so if we don't use it we don't have to have it installed?

Not easily. I have looked into breaking the dependency and setting it up programmatically. They haven't been 100% successful so far.

src/WorkspaceContext.ts
adam-fowler: Added a log event

0xTim (approved):

tristanlabelle:
This is super exciting! I'm looking forward for a path to better debugging support through this!

adam-fowler:
@tristanlabelle there is a pre-release out which includes this PR.

adam-fowler:
Merging into pre-release/1.6 branch

PR #613, opened 2023-09-28
Fix debug config path formats on Windows

tristanlabelle:
Debug configuration paths were generated with mixed back and forward slashes and no `.exe` extensions. That worked for `lldb` but not for `swift-lldb`.

https://github.com/swift-server/vscode-swift/pull/609 scoped down and applied to `pre-release/1.6`

### DIFF
--- a/src/debugger/launch.ts
+++ b/src/debugger/launch.ts
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import * as os from "os";
+import * as path from "path";
 import * as vscode from "vscode";
 import configuration from "../configuration";
 import { FolderContext } from "../FolderContext";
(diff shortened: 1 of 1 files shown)


src/debugger/launch.ts
```diff
@@ -113,14 +114,15 @@ function createExecutableConfigurations(ctx: FolderContext): vscode.DebugConfigu
...
+    const binaryExtension = process.platform === "win32" ? ".exe" : "";
     return executableProducts.flatMap(product => {
         return [
             {
                 type: DebugAdapter.adapterName,
                 request: "launch",
                 sourceLanguages: ["swift"],
                 name: `Debug ${product.name}${nameSuffix}`,
-                program: `${buildDirectory}/debug/` + product.name,
+                program: path.join(buildDirectory, "debug", product.name + binaryExtension),
```
compnerd, line 125:
```suggestion
                program: path.join(buildDirectory, "debug", `${product.name}${binaryExtension}`),
```

compnerd: Extracting the construction seems better:

```javascript
const Configuration = Object.freeze({
  Debug: "debug",
  Release: "release",
})

function product(configuration, name) {
  const extension = process.platform === "win32" ? ".exe" : ""
  return path.join(buildDirectory, Configuration[configuration], `${name}${extension}`)
}
```

This would give us something like:
```javascript
product(Configuration.Release, "test")
```

adam-fowler (approved):
Looks good to me. You can add the initial suggestion from @compnerd if you like. We can do a clean up of some of these functions later once this branch is merged back into main.

adam-fowler:
This is in the v1.6.1 prerelease

PR #618, opened 2023-10-04
Fix version comparison

tristanlabelle:
Typescript does not support operator overloading. What was happening when we compared two Version objects was that `toString` was called on both sides and the strings were compared. Unfortunately the unit test wasn't catching this because the strings ended up being lexicographically ordered.

This probably fixes several issues, but the one I was after is the Windows xctest path in the debug configuration being incorrect with a Swift 5.11 build because the `>= new Version(5, 7, 0)` was returning false.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -240,7 +240,10 @@ async function runSwiftScript(ctx: WorkspaceContext) {
 
     // Swift scripts require new swift driver to work on Windows. Swift driver is available
     // from v5.7 of Windows Swift
-    if (process.platform === "win32" && ctx.toolchain.swiftVersion < new Version(5, 7, 0)) {
+    if (
+        process.platform === "win32" &&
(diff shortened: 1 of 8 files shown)


adam-fowler (approved):
Whoops that slightly embarrassing. Changes look good

tristanlabelle:
@adam-fowler This PR was merged, but I don't see the changes in [main](https://github.com/tristanlabelle/vscode-swift/blob/e73398764ba403fd97186ea3f3cad6eecf13c781/src/commands.ts#L206). What happened? 🤔

adam-fowler:
Err not sure what has happened here. I'll go check what I have locally.

adam-fowler:
It is on main. Just after 1.5.2 changelog commit.

tristanlabelle:
Hmm never mind then, not sure what happened here. I also verified this works on 1.7.0. Thanks!

PR #621, opened 2023-10-11
Add client-side support for macro expansions

fwcd:
This implements basic client-side support for the `sourcekit-lsp/macroExpansion` request introduced in

- https://github.com/apple/sourcekit-lsp/pull/892

The expansion is presented using an in-memory text document in a peeked editor:

image

### To do

- [x] Declare the request (`sourcekit-lsp/macroExpansion`)
- [x] Add a VSCode command for performing the request (`swift.expandMacro`)
- [x] Find a better presentation style than dumping the expansion into an info message popup

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -27,7 +27,7 @@ import { DarwinCompatibleTarget, SwiftToolchain } from "./toolchain/toolchain";
 import { debugSnippet, runSnippet } from "./SwiftSnippets";
 import { debugLaunchConfig, getLaunchConfiguration } from "./debugger/launch";
 import { execFile } from "./utilities/utilities";
-import { macroExpansionRequest } from "./sourcekit-lsp/lspExtensions";
+import { MacroExpansionParams, macroExpansionRequest } from "./sourcekit-lsp/lspExtensions";
 
 /**
  * References:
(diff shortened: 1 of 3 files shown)


adam-fowler:
The third todo in your list is going to be the stumbling block. I've looked into finding a better way to present macros a while back and didn't find an easy solution. 

You mention rust-analyser in the related sourcekit-lsp PR. Do you know what they do to present macros?

fwcd:
They spawn another (regular) editor containing the expansion:

image

adam-fowler:
That's not really a great solution

fwcd:
I agree, haven't dug in too deep what APIs VSCode exposes in this regard, but I think it would be really cool if we could an "inline editor" for this, which incidentally would also approximate Xcode's macro expansion UI:

image

I think GitLens does or used to do something similar. We might still have to register a virtual document provider, but that should probably be doable.

fwcd:
The peeked editor approach seems to work well and looks nice too (check the PR description for a screenshot).

DougGregor:
> The peeked editor approach seems to work well and looks nice too (check the PR description for a screenshot).

That looks great, thank you!

src/commands.ts
```diff
@@ -477,6 +478,61 @@ function insertFunctionComment(workspaceContext: WorkspaceContext) {
...
+        };
+        const expansions = await client.sendRequest(macroExpansionRequest, params, token);
+        if (expansions && expansions.length >= 1) {
+            // Present the macro expansion using a custom in-memory URI scheme
+            // TODO: This will currently reregister the provider for this scheme
+            // every time a new macro expansion is requested. It looks like the
+            // old expansions will be closed correctly, but perhaps there's
+            // a more elegant solution to this?
+            const scheme = "swift-macro-expansion";
+            const provider = vscode.workspace.registerTextDocumentContentProvider(scheme, {
```
adam-fowler, line 501: We will need to manage the text document content providers slightly better. Currently you just keep creating them, without cleaning up the unused ones. Adding them to the workspaceContext subscriptions means they get removed when the workspace is closed, but not before then.

adam-fowler (commented):
It looks very nice. I didn't know about "editor.action.peekLocations"
In SourceKit-LSP you could add a CodeAction for macros to call the command you've setup here.

adam-fowler:
Closing this in favour of #945

PR #626, opened 2023-10-20
[lldb] Look for lldb-dap instead of lldb-vscode in the toolchain

JDevlieghere:
The LLDB Debug Adaptor Protocol implementation was renamed from lldb-vscode to lldb-dap [1]. This commit updates the VSCode plugin to look for lldb-dap instead of lldb-vscode in the toolchain.

[1] https://discourse.llvm.org/t/rfc-rename-lldb-vscode-to-lldb-dap/74075

### DIFF
--- a/package.json
+++ b/package.json
@@ -415,13 +415,13 @@
           "swift.debugger.useDebugAdapterFromToolchain": {
             "type": "boolean",
             "default": false,
-            "description": "Use lldb-vscode packaged with Swift toolchain as your debug adapter. This is currently only available on Windows platforms",
+            "description": "Use lldb-dap packaged with Swift toolchain as your debug adapter. This is currently only available on Windows platforms",
             "order": 1
           },
           "swift.debugger.path": {
             "type": "string",
             "default": "",
-            "description": "Path to lldb-vscode debug adapter.",
+            "description": "Path to lldb-vscode or lldb-dap debug adapter binary.",
             "order": 2
           }
         }
--- a/src/debugger/debugAdapter.ts
+++ b/src/debugger/debugAdapter.ts
@@ -19,7 +19,7 @@ import configuration from "../configuration";
 import contextKeys from "../contextKeys";
 
 /**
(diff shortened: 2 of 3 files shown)


JDevlieghere:
- I haven't cherrypicked the upstream commit to any of the Swift branches **yet**. 
 - I'd like to avoid having to add fallback logic to find both `lldb-dap `and `lldb-vscode`. It seems like the latter is only packaged with the Windows toolchain for now. If I cherrypick the rename to the 5.10 branch, does it sound reasonable to only look for `lldb-dap`? (@adam-fowler, @compnerd)

compnerd:
@JDevlieghere we are using the old name with the 5.9 toolchain already, we will need the fallback I fear.

adam-fowler:
> I'd like to avoid having to add fallback logic to find both lldb-dap and lldb-vscode

No it's not ideal, but as @compnerd says it is already being used in Windows. 

I don't know how many people are using it though. They could fall back to CodeLLDB for 5.9 maybe. The integration with lldb-vscode is only in a prerelease of the extension. Although changing it would mess with my release schedule slightly. I was hoping to do a full release with it being available. Managing multiple release branches is a pain. It would seem odd doing a full release of the extension with it using an executable not available in a swift release.

I would prefer if we had the fallback. It's not like the extension doesn't have tests for different swift versions elsewhere.

adam-fowler:
@compnerd we could do this based on swift version. If we are using 5.9 look for lldb-vscode, otherwise look for lldb-dap. That would mean we would need an updated 5.10 toolchain before this was included in a release though. What do you think?

compnerd:
Yeah, that seems reasonable. The 5.10 toolchain is waiting on @shahmishal as we can no longer generate the toolchain Azure due to the dependency on the swift toolchain due to macros.

shahmishal:
You can track the work here for Windows nightly toolchain - https://github.com/apple/swift/issues/69484

Issue #625, opened 2023-10-18
Rebuild times consistently take quite long in Build Task


dkolas:
**Describe the bug**
Re-building the project always takes a long time (30ish seconds) when done from Build task, but not from terminal (both without changing any files).  This is somewhat painful when running tests - rerunning the test takes a long time to start.

If I run the build via the build task repeatedly, it always takes approximately 30s:

```
 *  Executing task: swift build --build-tests 

Building for debugging...
Build complete! (29.70s)
 *  Terminal will be reused by tasks, press any key to close it. 

 *  Executing task: swift build --build-tests 

Building for debugging...
Build complete! (29.46s)
 *  Terminal will be reused by tasks, press any key to close it. 

 *  Executing task: swift build --build-tests 

Building for debugging...
Build complete! (28.35s)
 *  Terminal will be reused by tasks, press any key to close it.
```  

Note that the entire time is spent on some hidden step (no messages about compiling particular files are emitted).

If I simply run the build from a Terminal window within VSCode:

```
dkolas@terra:~/projects/nludb$ swift build --build-tests
Building for debugging...
Build complete! (29.50s)
dkolas@terra:~/projects/nludb$ swift build --build-tests
Building for debugging...
Build complete! (0.37s)
dkolas@terra:~/projects/nludb$ swift build --build-tests
Building for debugging...
Build complete! (0.36s)
dkolas@terra:~/projects/nludb$ swift build --build-tests
Building for debugging...
Build complete! (0.37s)
dkolas@terra:~/projects/nludb$ 
```

Note that after the first build, it takes less than 1s per build, which is more in line with expectations

I played around with various swift compiler options (--force-resolved-versions, --skip-update) but these didn't seem to have any effect, and it doesn't look like it's doing that stuff anyway since it's not in the output.

**To Reproduce**
I suspect this has something to do with either the number of files in the project or the number of dependencies, so it's unclear to me how to strip this down to an example without our code in it.

**Expected behavior**
Rebuilds (at least within some short wall-clock timeframe) should take nearly 0 time

**Environment**
 - OS: Windows 11 (vscode) connected to WSL Ubuntu 22.04.2
 - Swift version 5.9 (swift-5.9-RELEASE)
 - Visual Studio Code version: 1.83.1
 - vscode-swift version: 1.5.2

Thanks again for all your work on the project.

adam-fowler:
Yeah I've heard this being reported on Windows before.
[...]

PR #635, opened 2023-10-27
Improve WSL rebuild times

dkolas:
This PR fixes: #625

@adam-fowler Here's a PR to address the issue. This works for me in the debugging VSCode instance. It's unfortunately not quite as clean as I might have liked; would love your opinion on any ways to improve.

- The filter is only added in WSL environment in an attempt to not affect other environments.
- I would have strongly preferred to pass the env var override in as an env var to the task configuration. However the env var was still overwritten by VSCode on the path to execution, defeating the purpose of the override.
- Needed to change to shell execution from process execution to allow the env var prefix to work correctly.
- Had to change the ShellExecution from (process, args, ...) to (commandline...), because otherwise the invocation quotes the first parts together as the executable and it fails.

### DIFF
--- a/src/SwiftTaskProvider.ts
+++ b/src/SwiftTaskProvider.ts
@@ -266,10 +266,30 @@ export function createSwiftTask(
     } else {
         prefix = "";
     }
+    const envVarFilterPrefix = filterEnvironmentVariablePrefix();
+    if (envVarFilterPrefix !== null) {
+        prefix = prefix + envVarFilterPrefix;
+    }
     task.detail = `${prefix}swift ${args.join(" ")}`;
     task.group = config?.group;
     task.presentationOptions = config?.presentationOptions ?? {};
     return task;
+}
+
+/**
+ * Helper function to return a prefix for swift execution tasks
+ * when running in a WSL environment. This prevents swift's package
+ * manager manifest cache from being invalidated, speeding up incremental
+ * builds considerably. See https://github.com/swift-server/vscode-swift/issues/625
+ */
+function filterEnvironmentVariablePrefix(): string | undefined {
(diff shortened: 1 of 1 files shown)


adam-fowler:
I had a quick look at the changes here and just concatenating all the arguments into one string is not equivalent to sending them as an array. For instance how do you deal with arguments that include spaces

dkolas:
Indeed, it does not work if the arguments require shell-quoting.  If none do, then it's equivalent, but I now see that there's a plugin setting for arbitrary build arguments, so I guess that can't be guaranteed.

I'm a bit stuck on any other way to pass the arguments, though, without using the `commandline` version of the ShellExecution constructor. 

Things that don't seem to work (not necessarily unexepectedly, just trying to cover bases):
- putting the env var filter as the process and the swift executable as the first arg (bash can't resolve the env var filter as an executable file)
- Combining the env var filter and the swift process as the process (bash can't resolve the combination as an executable file)

Does the task have knowledge of the shell it's going to run in at definition time for proper quoting?  If so I haven't found it yet.

adam-fowler:
I had a quick look through the VSCode codebase and they go out of their way to ensure this environment variable is present when setting up a new terminal. This kinda makes me nervous of removing it. It appears to be related to the DevContainer support. 

I think the better solution would be to get swift to filter the environment variables.

adam-fowler:
I did try something similar but then realised the location of the swift executable is used to define the location of other relevant executables like sourcekit-lsp and lldb

dkolas:
Since this has been fixed upstream in the swift package manager:

https://github.com/apple/swift-package-manager/pull/7056

it doesn't seem to make sense to pursue this PR.

PR #638, opened 2023-11-20
Add a configuration to disable automatic test discovery

tristanlabelle:
We suspect that this is interfering with `swift build`'s access to the build database (either due to the `swift test` call or due to triggering workspace symbol searches) and causing many random build failures, so we should have a way to disable it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -248,6 +248,13 @@
             "default": "",
             "markdownDescription": "Path to the build directory passed to all swift package manager commands.",
             "order": 11
+          },
+          "swift.autoDiscoverTests": {
+            "type": "boolean",
+            "default": true,
+            "markdownDescription": "Automatically discover Swift tests and show them in the test explorer.",
+            "scope": "machine-overridable",
+            "order": 12
           }
         }
       },
--- a/src/TestExplorer/TestExplorer.ts
+++ b/src/TestExplorer/TestExplorer.ts
@@ -27,13 +27,14 @@ export class TestExplorer {
     static errorTestItemId = "#Error#";
     public controller: vscode.TestController;
     private lspFunctionParser?: LSPTestDiscovery;
-    private subscriptions: { dispose(): unknown }[];
+    private subscriptions: { dispose(): unknown }[] = [];
 
     constructor(public folderContext: FolderContext) {
         this.controller = vscode.tests.createTestController(
             folderContext.name,
             `${folderContext.name} Tests`
         );
+        this.subscriptions.push(this.controller);
 
         this.controller.resolveHandler = async item => {
             if (!item) {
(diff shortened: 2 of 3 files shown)


adam-fowler:
`swift test list` is called once a build has completed. So unless you run a build immediately after (like within a second) it will not clash. Automatic test discovery does not trigger workspace symbol searches it just intercepts any that sourcekit-lsp is already running. And anyway these only communicate with sourcekitd. 

I've found in the past the build database issues are related to sourcekit-lsp. See if you can replicate this issue without sourcekit-lsp running.

If

tristanlabelle:
I enabled diagnostics logging and I see the below being printed in the Swift output pane during our build: test discovery running and failing after doing operations on checked out repos.

We have people who have disabled LSP and are still hitting the random build issues, but people who disable the Swift extension altogether stop hitting them, so it points towards other process launches by the Swift extension interfering with the build.

This is the info I have so far. I'll try to gather more evidence by having people use a private build with this change.

```
15:24:19: <project name snip>: Test Discovery Failed: Updating https://github.com/thebrowsercompany/swift-composable-architecture
Updated https://github.com/thebrowsercompany/swift-composable-architecture (0.61s)
[similar lines snip]
Removing https://github.com/thebrowsercompany/swift-webdriver
[similar lines snip]
Creating working copy for https://github.com/SwiftGen/StencilSwiftKit.git
Working copy of https://github.com/SwiftGen/StencilSwiftKit.git resolved at 2.10.1
error: The operation could not be completed. The file doesn’t exist.
```

tristanlabelle:
Ah yes! The issue is specifically with this line https://github.com/swift-server/vscode-swift/blob/20e0907360cd891d3de72526d4c0084d7bc86414/src/TestExplorer/TestExplorer.ts#L82

The Swift starts a test list in the background when we open a VS Code window. Even though it passes `--skip-build`, it does result in SPM creating the checkouts for all dependencies. This is happening silently while we start a build, and then we end up with corrupted checkouts because the two SPM processes are racing.

I also tried having the extension pass `--skip-update`, but SPM will still create all the checkouts.

This is a high-impacting issue for us. Does adding this toggle seem reasonable, so we can use it as a workaround while we think about other ways to avoid overlapping checkouts?

adam-fowler:
Are you running with auto resolve disabled? This probably shouldn't run if so.

tristanlabelle:
We are disabling auto resolve for our repos, but the test discovery logic in the extension isn't conditional on that, and the resolution happens as part of `swift test list` due to what I think is an SPM bug: https://github.com/apple/swift-package-manager/issues/7113 , but I think it's still valuable to have this toggle because we can't work around a problem caused by a feature that is triggered automatically.

adam-fowler:
I have one more issue to resolve and a new build will be released. 

Does VSCode not indicate to you when a new version of the extension is available? I should probably attempt to keep the tags up to date at least.

adam-fowler:
I've tagged all the old versions and added a script to tag any new ones `npm run tag`

tristanlabelle:
> Does VSCode not indicate to you when a new version of the extension is available? I should probably attempt to keep the tags up to date at least.

VSCode will auto-update it and I'm not sure if it tells us. That's convenient in general except that it's practical to know when those deployments happen so we can make internal announcements about pending features being deployed. Also in this case we won't get updates at all because we're temporarily using a private build to work around the issue here, and that's off the update train.

PR #656, opened 2023-12-06
Add option to focus on problems after build error

adam-fowler:


package.json
```diff
@@ -243,10 +243,19 @@
             "markdownDescription": "**Experimental**: Run `swift build` in the background whenever a file is saved.",
             "order": 10
           },
-          "swift.focusOnProblems": {
```
0xTim: Is this considered a breaking change for those building on top of the extension?

0xTim (approved):
One query otherwise LGTM

package.json
adam-fowler: You are looking at the changes from one commit not the whole PR? Originally I implemented it as `focusOnProblems` Boolean and then changed my mind to use the enum.

PR #673, opened 2024-02-02
Activate the extension when starting debugging

tristanlabelle:
In Swift projects using CMake as the build system, the Swift extension doesn't auto-start when debugging, resulting in the below:

image

`onDebugInitialConfiguration` might also be useful since [the Java extension does this](https://github.com/microsoft/vscode-java-debug/blob/14fe7d5d0d41a397eaff4f3701860653850058fb/package.json#L33C6-L33C34), but I'm not sure how to test it and the current change solves the problem I'm after.

How tested: tried starting debugging before and after.


jumhyn-browser (approved):

compnerd (approved):

adam-fowler (approved):
Looks good. I think `onDebugInitialConfiguration` will cause the extension to activate when someone starts a debug sessions (but regardless of launch type).

tristanlabelle:
@adam-fowler I think `onDebugInitialConfiguration` is rather when you don't have a `launch.json` file, press F5, and the extension is supposed to contextually figure out a debug configuration (for example single-file debugging in a shell language)

PR #679, opened 2024-02-14
Limit when test explorer and test discovery is activated

adam-fowler:
- Don't add test explorer if package has no test targets, or is not a SwiftPM project
- Catch changes to Package.swift to enable/disable test explorer based on if package has test targets
- Don't run test discovery if package has no test targets
- Add `dontTriggerTestDiscovery` parameter to build task definition and set it to true when building executables

### DIFF
--- a/package.json
+++ b/package.json
@@ -694,6 +694,10 @@
             "description": "Disable any queued tasks while running the task. This includes the auto resolve when Packages.resolved is updated.",
             "type": "boolean"
           },
+          "dontTriggerTestDiscovery": {
+            "description": "Don't trigger the test discovery process.",
+            "type": "boolean"
+          },
           "macos": {
             "type": "object",
             "description": "macOS specific task configuration",
--- a/src/TestExplorer/TestExplorer.ts
+++ b/src/TestExplorer/TestExplorer.ts
@@ -58,7 +58,10 @@ export class TestExplorer {
                 event.exitCode === 0 &&
                 task.definition.dontTriggerTestDiscovery !== true
             ) {
-                this.discoverTestsInWorkspace();
+                // only run discover tests if the library has tests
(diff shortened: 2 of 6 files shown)


src/TestExplorer/TestExplorer.ts
```diff
@@ -80,11 +83,28 @@ export class TestExplorer {
...
+                            // discover tests will kick off a resolve if required
+                            if (!configuration.folder(folder.workspaceFolder).disableAutoResolve) {
+                                folder.testExplorer?.discoverTestsInWorkspace();
+                            }
+                        }
+                    }
+                    break;
+                case FolderEvent.packageUpdated:
+                    if (folder) {
+                        const testTargets = folder.swiftPackage.getTargets("test");
```
tristanlabelle, line 98: Suggestion: `const hasTestTargets = folder.swiftPackage.getTargets("test").length > 0`, since the rest of the code doesn't need the array.

package.json
```diff
@@ -694,6 +694,10 @@
             "description": "Disable any queued tasks while running the task. This includes the auto resolve when Packages.resolved is updated.",
             "type": "boolean"
           },
+          "dontTriggerTestDiscovery": {
```
tristanlabelle, line 697: Suggestion: name positively "triggerTestDiscovery", to avoid a double negation when assigning `false`.

tristanlabelle (approved):

0xTim (approved):
Nothing further to add!

src/TestExplorer/TestExplorer.ts
adam-fowler: Not sure it'll make a difference to performance but it make the code clearer so I've implemented it

package.json
adam-fowler: The default should be trigger test discovery, so I'm not keen on changing this

PR #680, opened 2024-02-16
Add setting to disable SwiftPM integration

adam-fowler:


package.json
```diff
@@ -264,6 +264,12 @@
             "default": "",
             "markdownDescription": "Path to the build directory passed to all swift package manager commands.",
             "order": 12
+          },
+          "swift.disableSwiftPackageManagerIntegration": {
```
tristanlabelle, line 268: Stylistic: naming options positively avoids a double negation, is more consistent across settings, and supports changing the default in the future (the negative naming is redundant information with the default value).

It's also more consistent with VS Code conventions. If you open `settings.json` and look at the completion for `disable`, there's a few extensions that registered settings with that term, but none of the built-in settings use the it, whereas `enable` is used consistently.

image

tristanlabelle (approved):
Thanks so much!

0xTim (approved):

PR #701, opened 2024-02-28
Use `workspace/tests` to get list of tests instead of `swift test list`.

adam-fowler:
- Generalised test explorer update for single file and whole workspace.
- Create test explorer update from `workspace/tests` LSP response and apply
- Create test explorer update from document symbols response for a single file and apply (Replacing previous update code in LSPTestDiscovery.ts)

I haven't include in this PR the creation of a test explorer update from `swift test list` just to keep the PR size down. I'll do that in another PR.

### DIFF
--- a/src/TestExplorer/TestExplorer.ts
+++ b/src/TestExplorer/TestExplorer.ts
@@ -22,6 +22,7 @@ import { Version } from "../utilities/version";
 import configuration from "../configuration";
 import { buildOptions, getBuildAllTask } from "../SwiftTaskProvider";
 import { SwiftExecOperation, TaskOperation } from "../TaskQueue";
+import { updateTestsFromClasses } from "./TestDiscovery";
 
 /** Build test explorer UI */
 export class TestExplorer {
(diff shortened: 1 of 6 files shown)


src/TestExplorer/TestExplorer.ts
```diff
@@ -205,12 +231,12 @@ export class TestExplorer {
                             classItem.children.forEach(funcItem => {
                                 const testName = `${targetItem.label}.${classItem.label}/${funcItem.label}`;
                                 if (
-                                    !results.find(item => item === testName) &&
+                                    !results.find(item => item === testName) /* &&
```
0xTim, line 234: Do we want to get rid of this?

src/TestExplorer/TestExplorer.ts
```diff
@@ -252,7 +278,7 @@ export class TestExplorer {
 
             // add items to target test item as the setActive call above may not have done this
             // because the test target item did not exist when it was called
-            this.lspFunctionParser?.addTestItems();
+            //this.lspFunctionParser?.addTestItems();
```
0xTim, line 281: As above

0xTim (approved):
One query around commented out code

src/TestExplorer/TestExplorer.ts
adam-fowler: Both of these will be deleted in the next PR which creates a test explorer update from `swift test list`.

src/TestExplorer/TestExplorer.ts
adam-fowler: see above

PR #706, opened 2024-02-29
Merge test parsing functions into one

adam-fowler:
Added `TestItemFinder` interface to define how you get a TestItem from the test id extracted from XCTest output
Added `TestRegex` interface which defines the regex to use when parsing the XCTest output.
Merge both `parseResultDarwin` and `parseResultNonDarwin` into one function `parseResult`.

I will also need the `TestItemFinder` when supporting parallel testing

### DIFF
--- a/src/TestExplorer/TestOutputParser.ts
+++ b/src/TestExplorer/TestOutputParser.ts
@@ -2,7 +2,7 @@
 //
 // This source file is part of the VSCode Swift open source project
 //
-// Copyright (c) 2021-2023 the VSCode Swift project authors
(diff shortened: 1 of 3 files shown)


0xTim (approved):
It's a shame the platforms are so different but LGTM!

adam-fowler:
They're not that different but just enough to require all this. The macOS output is more helpful as it includes the target name in the symbol name (something to do with it being Objective C based) while I only get a class and method name for Linux/Windows. 

A lot of the work is related to dealing with the edge case where two targets both have a test with the same class name and method name.

PR #714, opened 2024-03-06
Changes from Release 1.8.1

adam-fowler:
Support Package.resolved version 3 - needed to support Swift 5.10 packages

### DIFF
--- a/src/SwiftPackage.ts
+++ b/src/SwiftPackage.ts
@@ -78,7 +78,7 @@ export class PackageResolved {
                         pin.state
                     )
             );
-        } else if (this.version === 2) {
+        } else if (this.version === 2 || this.version === 3) {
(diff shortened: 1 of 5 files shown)


src/SwiftPackage.ts
```diff
@@ -78,7 +78,7 @@ export class PackageResolved {
                         pin.state
                     )
             );
-        } else if (this.version === 2) {
+        } else if (this.version === 2 || this.version === 3) {
```
0xTim, line 81: I'm assuming the actual JSON layout hasn't changed, just the version number?

0xTim (approved):

src/SwiftPackage.ts
adam-fowler: JSON layout of the bits we care about hasn't changed

PR #718, opened 2024-03-07
Add docker compose file for 5.10

adam-fowler:


adam-fowler:
Hi @yim-lee would we be able to get a 5.10 build machine for the VSCode extension. Thanks

docker/docker-compose.2204.510.yaml
```diff
@@ -0,0 +1,16 @@
+version: "3"
+
+services:
+
+  runtime-setup:
+    image: vscode-swift:22.04-5.10
+    build:
+      args:
+        ubuntu_version: "focal"
```
yim-lee, line 9:
```suggestion
        ubuntu_version: "jammy"
```

Ubuntu 22.04 should be `jammy`

yim-lee:
@adam-fowler CI for Swift 5.10 is set up!

Issue #754, opened 2024-04-18
Create new project command


matthewbastien:
**Is your feature request related to a problem? Please describe.**
This isn't related to a specific problem per-se. However, extensions for other languages such as Dart provide commands as a nice entry point into creating a new project. E.g. for Dart you can run the command ```Dart: New Project``` which will take you through a set of steps to create a new project based on a specific template:

<img width="600" alt="dart_new_project_wizard" src="https://github.com/swift-server/vscode-swift/assets/65295938/56127c2a-df90-45bf-ba95-0db77dd906ca">

VSCode's built-in git integration has a similar workflow for cloning repositories from GitHub using ```Git: Clone```.

**Describe the solution you'd like**
The Swift extension could provide a command to create a new project. The command would require the user to select a project template, a folder to create the project in, and a name for the project. It would then use Swift Package Manager to create the actual project on disk and open it within VSCode automatically.

**Describe alternatives you've considered**
The only alternative would be to invoke Swift Package Manager manually in the terminal.

**Additional context**
Swift Package Manager provides a command that can initialize a new project:
```
swift package init --type <type> --name <name>
```
The list of project types is available via the ```--help``` option:
```
❯ swift package init --help
OVERVIEW: Initialize a new package

USAGE: swift package init [--type <type>] [--name <name>]

OPTIONS:
  --type <type>           Package type: (default: library)
        library           - A package with a library.
        executable        - A package with an executable.
        tool              - A package with an executable that uses
                            Swift Argument Parser. Use this template if you
                            plan to have a rich set of command-line arguments.
        build-tool-plugin - A package that vends a build tool plugin.
        command-plugin    - A package that vends a command plugin.
        macro             - A package that vends a macro.
        empty             - An empty package with a Package.swift manifest.
  --name <name>           Provide custom package name
  --version               Show the version.
  -h, -help, --help       Show help information.
```

PR #755, opened 2024-04-19
Create new project

matthewbastien:
Implementation of the feature request #754.

The workflow is as follows:
1. User runs the command ```Swift: New Project```
2. A prompt appears to select the project type. Available project types are retrieved by parsing the output of ```swift package init --help```.
3. An open folder dialog appears to select the root folder for the new project.
4. A quick input appears to enter a name for the project.
5. The project folder is created on disk and ```swift package init``` is executed to create the project based on the given inputs.
6. A modal appears asking whether or not the user would like to open the newly created folder in the current window or a new one. An automatic action for this step can be configured with the ```swift.openAfterCreateProject``` setting. This is essentially the same modal used by the ```Git: Clone``` command.

I've also added a fairly basic test to make sure that the parsing of ```swift package init --help``` is functioning as intended.

Feedback is greatly appreciated! This is my first commit to this project and I definitely don't know all the ins and outs to the code base.


src/commands.ts
```diff
@@ -682,6 +834,7 @@ function updateAfterError(result: boolean, folderContext: FolderContext) {
  */
 export function register(ctx: WorkspaceContext) {
     ctx.subscriptions.push(
+        vscode.commands.registerCommand("swift.createProject", () => createProject(ctx)),
```
daveyc123, line 819: createProject() has a few instances where it returns undefined. What's the experience for the user in these cases? Should createProject be providing guidance/errors to the user in the form of notifications?

matthewbastien: The places where createProject() returns undefined correspond to the user cancelling dialogs. I don't think we need any notifications in this case.

daveyc123: 👍 What's the behaviour from vscode for exceptions (e.g. the execSwift or mkdir fail)

matthewbastien: It pops up a modal with the error message from the exception.

src/commands.ts
```diff
@@ -91,6 +92,139 @@ export async function updateDependencies(ctx: WorkspaceContext) {
...
+        validateInput(value) {
+            // Ensure there are no name collisions
+            if (existingNames.includes(value)) {
+                return "A file/folder with this name already exists";
+            }
+            return undefined;
+        },
+    });
+    if (!projectName) {
+        return undefined;
```
adam-fowler, line 153: Can we move this check inside the validateInput function of showInputBox

src/commands.ts
```diff
@@ -91,6 +92,139 @@ export async function updateDependencies(ctx: WorkspaceContext) {
...
+                }
+            );
+        },
+        1000
+    );
+
+    // Prompt the user whether or not they want to open the newly created project
+    const isWorkspaceOpened = !!vscode.workspace.workspaceFolders;
+    const config = vscode.workspace.getConfiguration("swift");
+    const openAfterCreate = config.get<string>("openAfterCreateProject");
```
adam-fowler, line 181: Can we add configuration settings to configuration.ts

src/commands.ts
```diff
@@ -91,6 +92,139 @@ export async function updateDependencies(ctx: WorkspaceContext) {
...
+        return undefined;
+    }
+
+    // Create the folder that will store the new project
+    const projectUri = vscode.Uri.joinPath(selectedFolder[0], projectName);
+    await fs.mkdir(projectUri.fsPath);
+
+    // Use swift package manager to initialize the swift project
+    await withDelayedProgress(
+        {
```
adam-fowler, line 162: Given there is no progress reported here, any reason you can't use `WorkspaceContext.statusItem.showStatusWhileRunning()`

src/commands.ts
adam-fowler: Have you verified what the message is for say `mkdir` failing? Would be good to make sure it is human readable.

src/toolchain/toolchain.ts
```diff
@@ -178,6 +187,47 @@ export class SwiftToolchain {
         return xcodes.trimEnd().split("\n");
     }
 
+    /**
+     * Get a list of new project templates from swift package manager
+     * @returns a {@link SwiftProjectTemplate} for each discovered project type
+     */
+    public async getProjectTemplates(): Promise<SwiftProjectTemplate[]> {
+        const { stdout } = await execSwift(["package", "init", "--help"], "default");
+        const lines = stdout.split(/\r?\n/g);
```
adam-fowler, line 196: This is some funky way to get project templates, but yeah can't see any other way of doing it.

I almost wonder if it is worthwhile rolling our own outside of SwiftPM. Would love to give users the ability to create their own templates. Maybe using Mustache templating to allow users to add in customisation. But that'd be for another task I think.

adam-fowler (commented):
In general looks pretty good, I've added a few comments

adam-fowler:

adam-fowler:

adam-fowler:
CI is returning a bunch of errors for older versions of swift. New features can probably be gated on the last three versions of swift. Get it working for 5.8 if it works for all the earlier swift versions as well you are done. Otherwise Add a new contextKey `newProjectAvailable` in contextKeys.ts and use the contextKey to control availability of the command. At startup you can check the swift version and set it to false if the swift version is less than 5.8

package.json
```diff
@@ -37,6 +37,11 @@
       }
     ],
     "commands": [
+      {
+        "command": "swift.createProject",
+        "title": "New Project",
```
award999, line 42: suggestion: "New" in title but "create" in ID, maybe make consistent?

suggestion: Typically have ellipsis suffix if the command will further prompt user for input

src/commands.ts
```diff
@@ -91,6 +92,139 @@ export async function updateDependencies(ctx: WorkspaceContext) {
...
+            action = "open";
+        } else if (result === openNewWindow) {
+            action = "openNewWindow";
+        } else if (result === addToWorkspace) {
+            action = "addToWorkspace";
+        }
+    }
+
+    if (action === "open") {
+        vscode.commands.executeCommand("vscode.openFolder", projectUri, { forceReuseWindow: true });
```
award999, line 219: suggestion: can we await or .catch to not leave a floating promise?

src/commands.ts
```diff
@@ -91,6 +92,139 @@ export async function updateDependencies(ctx: WorkspaceContext) {
...
+                ctx.toolchain,
+                {
+                    cwd: projectUri.fsPath,
+                }
+            );
+        },
+        1000
+    );
+
+    // Prompt the user whether or not they want to open the newly created project
```
award999, line 178: suggestion: maybe we can separate the creation from the opening of the project? The function is getting a little long

award999 (commented):
a few minor comments :)

src/commands.ts
adam-fowler: If we're going to do that maybe we should break it out of commands.swift and give it its own file

src/commands.ts
matthewbastien: Good question! I tried it out myself and it looks like the error message is fairly self explanatory:

Screenshot 2024-04-22 at 1 05 31 PM

package.json
```diff
@@ -160,6 +165,23 @@
       {
         "title": "Swift",
         "properties": {
+          "swift.openAfterCreateProject": {
```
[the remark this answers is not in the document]
matthewbastien: Done. I put it at the end of the settings since I can't imagine this being a high runner use case.

src/utilities/utilities.ts
```diff
@@ -321,3 +321,78 @@ export function expandFilePathTilda(filepath: string): string {
     }
     return filepath;
 }
+
+/**
+ * A wrapper around {@link vscode.Progress} that allows for delaying progress reporting.
+ */
+class ProgressWrapper<T> implements vscode.Progress<T> {
```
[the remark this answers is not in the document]
matthewbastien: Done.

src/commands.ts
matthewbastien: I've added some more validation here, but the check for undefined is still necessary for cancellation.

src/commands.ts
matthewbastien: Done.

src/commands.ts
matthewbastien: I'd expect that 99% of the time we would never even show the notification. In my testing it never showed up unless I reduced the timeout. However, if it does take longer than expected, I want it to be a prominent notification rather than something hidden away in the status bar. That way users know that something is actually happening in the background.

That being said, this does add extra code that will need to be maintained. I'm not too strongly attached to this opinion and can just use showStatusWhileRunning() instead if that makes more sense to you.

src/toolchain/toolchain.ts
matthewbastien: Yeah I agree. I'd honestly like an option in Swift Package Manager to get these as JSON rather than doing the crazy command line parsing. However, we still need to support Swift versions <6.

package.json
matthewbastien: Done.

src/commands.ts
matthewbastien: Good catch!

src/commands.ts
matthewbastien: I'm inclined to leave it where it is for now. I'm not sure when we'd even re-use these parts. The project creation aspect is literally just a mkdir() + execSwift() call. If we do end up re-using some of the code down the road then we can split it out.

matthewbastien:
> CI is returning a bunch of errors for older versions of swift. New features can probably be gated on the last three versions of swift. Get it working for 5.8 if it works for all the earlier swift versions as well you are done. Otherwise Add a new contextKey `newProjectAvailable` in contextKeys.ts and use the contextKey to control availability of the command. At startup you can check the swift version and set it to false if the swift version is less than 5.8

The output of `swift package init --help` changes pretty drastically in versions <5.8.0. I've added the context key for enablement of the command, but we still need to support users being able to create a project in cases where the Swift extension is not activated yet (e.g. when no workspace is open). As such, I've also added a `swift.isActivated` context key to detect this. An error is shown to the user if they try to use the create project command with an unsupported swift version.

adam-fowler (approved):
I think we are good here

Issue #762, opened 2024-04-21
Use `textDocument/tests` LSP request to augment test list on active document save


adam-fowler:
When a swift file is saved, make `textDocument/tests` LSP request and use the response to augment the test list generated from `workspace/tests`. 

Currently we are using a middleware to parse symbols from document symbols responses. This can be disabled for Swift 6 once the above is implemented. This is the code setting it up https://github.com/swift-server/vscode-swift/blob/dfba3243e6cb97faa45037d11478e2714d4d5e17/src/TestExplorer/TestExplorer.ts#L127-L133

plemarquand:
Closed by #767

Issue #761, opened 2024-04-21
Use hierarchical `workspace/tests` LSP request to get Test list


adam-fowler:
There is code is place to read the original flat representation format of `workspace/tests` but this has now been replaced by a hierarchical view that is closer in format to how tests are represented in VSCode. 

The request should be called from `TestExplorer. discoverTestsInWorkspace` https://github.com/swift-server/vscode-swift/blob/dfba3243e6cb97faa45037d11478e2714d4d5e17/src/TestExplorer/TestExplorer.ts#L170-L180. This is called at initialisation and after every compile, where a test file has been changed. Currently the LSP request call is commented out.

`LSPTestDiscovery.getWorkspaceTests` returns an intermediary format. You can probably avoid this and merge the response directly into the test list. Anyway the intermediary format is too restrictive for swift-testing.

plemarquand:
Closed by #767

Issue #757, opened 2024-04-19
Add support for swift-testing


daveyc123:
The extension should support discovering and executing tests defined via swift-testing. XCTests should continue to work alongside this functionality.

plemarquand:
I'll take this.

plemarquand:
This depends on #761 and #762

PR #767, opened 2024-04-24
Discover tests via LSP if available

plemarquand:
If the LSP is available and supports the two requests for listing tests then use those to discover tests in the workspace/document. If these requests are unsupported or if they fail, fall back to the existing methods.

This patch uses the existing code to leverage the LSP's workspace/tests method and adds the textDocument/tests for listing tests within documents as they change.

These requests produce a collection of TestClass structures, which are hierarchical. These are passed to one of the methods in TestDiscovery.ts which diffs against the existing tree and adds/removes tests. This is similar to how it already worked, but now supports arbitrary levels of nesting in the TestClass structure.

This patch also adds preliminary support for listing swift-testing tests through the LSP methods, as well as through `swift test --list-tests`. It does not support finding swift-testing tests through the document symbols.

Another small improvement is if listing tests fails via `swift test --list-tests` we'll attempt a `swift build --build-tests` before trying again. Today, if the user was on the 5.10 toolchain and no build had yet been performed they would get a somewhat cryptic error in the test explorer.

Swift-testing tests are filtered out of the list of available tests until code is added to run them. This will be added with #757.

- Adds tests for parsing the results of `swift test --list-tests`
- Adds tests for parsing tests via document symbols

This patch addresses both #761 and #762.

### DIFF
--- a/src/TestExplorer/LSPTestDiscovery.ts
+++ b/src/TestExplorer/LSPTestDiscovery.ts
@@ -44,15 +44,9 @@ export class LSPTestDiscovery {
         document: vscode.Uri
     ): Promise<TestDiscovery.TestClass[]> {
         return await this.languageClient.useLanguageClient(async (client, token) => {
-            const workspaceTestCaps = client.initializeResult?.capabilities.experimental;
-            if (!workspaceTestCaps) {
-                throw new Error("textDocument/tests requests not supported");
-            }
-            const textDocumentCaps = workspaceTestCaps[textDocumentTestsRequest.method];
-
             // Only use the lsp for this request if it supports the
             // textDocument/tests method, and is at least version 2.
-            if (textDocumentCaps?.version >= 2) {
+            if (this.checkExperimentalCapability(client, textDocumentTestsRequest.method, 2)) {
                 const testsInDocument = await client.sendRequest(
                     textDocumentTestsRequest,
                     { textDocument: { uri: document.toString() } },
@@ -60,7 +54,7 @@ export class LSPTestDiscovery {
                 );
                 return this.transform(client, swiftPackage, testsInDocument);
             } else {
-                throw new Error("textDocument/tests requests not supported");
+                throw new Error(`${textDocumentTestsRequest.method} requests not supported`);
             }
         });
     }
@@ -74,15 +68,9 @@ export class LSPTestDiscovery {
         workspaceRoot: vscode.Uri
     ): Promise<TestDiscovery.TestClass[]> {
         return await this.languageClient.useLanguageClient(async (client, token) => {
-            const workspaceTestCaps = client.initializeResult?.capabilities.experimental;
-            if (!workspaceTestCaps) {
-                throw new Error("textDocument/tests requests not supported");
-            }
-            const workspaceTestsCaps = workspaceTestCaps[workspaceTestsRequest.method];
-
             // Only use the lsp for this request if it supports the
             // workspace/tests method, and is at least version 2.
-            if (workspaceTestsCaps?.version >= 2) {
+            if (this.checkExperimentalCapability(client, workspaceTestsRequest.method, 2)) {
                 const tests = await client.sendRequest(workspaceTestsRequest, {}, token);
                 const testsInWorkspace = tests.filter(item =>
                     isPathInsidePath(
@@ -93,9 +81,26 @@ export class LSPTestDiscovery {
 
                 return this.transform(client, swiftPackage, testsInWorkspace);
             } else {
-                throw new Error("workspace/tests requests not supported");
+                throw new Error(`${workspaceTestsRequest.method} requests not supported`);
             }
         });
+    }
+
+    /**
+     * Returns `true` if the LSP supports the supplied `method` at or
+     * above the supplied `minVersion`.
+     */
+    private checkExperimentalCapability(
+        client: LanguageClient,
+        method: string,
+        minVersion: number
+    ) {
+        const experimentalCapability = client.initializeResult?.capabilities.experimental;
+        if (!experimentalCapability) {
+            throw new Error(`${method} requests not supported`);
+        }
+        const targetCapability = experimentalCapability[method];
+        return (targetCapability?.version ?? -1) >= minVersion;
     }
 
     /**
--- a/src/TestExplorer/LSPTestDiscovery.ts
+++ b/src/TestExplorer/LSPTestDiscovery.ts
@@ -52,7 +52,7 @@ export class LSPTestDiscovery {
                     { textDocument: { uri: document.toString() } },
                     token
                 );
-                return this.transform(client, swiftPackage, testsInDocument);
+                return this.transformToTestClass(client, swiftPackage, testsInDocument);
             } else {
                 throw new Error(`${textDocumentTestsRequest.method} requests not supported`);
             }
@@ -79,7 +79,7 @@ export class LSPTestDiscovery {
                     )
                 );
 
-                return this.transform(client, swiftPackage, testsInWorkspace);
+                return this.transformToTestClass(client, swiftPackage, testsInWorkspace);
             } else {
                 throw new Error(`${workspaceTestsRequest.method} requests not supported`);
             }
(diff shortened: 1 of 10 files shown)


plemarquand:
Note: This will only work with the latest sourcekit-lsp. To test you can clone the sourcekit-lsp repo, `swift build` and then set the `swift.sourcekit-lsp.serverPath` VSCode setting to point to your built sourcekit-lsp.

src/TestExplorer/LSPTestDiscovery.ts
```diff
@@ -29,104 +34,118 @@ import { LanguageClientManager } from "../sourcekit-lsp/LanguageClientManager";
...
-                    return {
-                        name: item.name,
-                        location: client.protocol2CodeConverter.asLocation(item.location),
-                        functions: functions,
-                    };
-                });
-            return classes;
+            const workspaceTestCaps = client.initializeResult?.capabilities.experimental;
+            if (!workspaceTestCaps) {
+                throw new Error("textDocument/tests requests not supported");
```
award999, line 79: todo: "workspace/tests"

suggestion: Since same logic for textDocument, we could extract this into it's own function so we can re-use if more experimental capabilities are added

src/TestExplorer/LSPTestDiscovery.ts
```diff
@@ -29,104 +34,118 @@ import { LanguageClientManager } from "../sourcekit-lsp/LanguageClientManager";
...
+        swiftPackage: SwiftPackage
+    ): string {
+        // XCTest: Target.TestClass/testCase
+        // swift-testing: TestClass/testCase()
+        //                TestClassOrStruct/NestedTestSuite/testCase()
+
+        let id: string = item.id;
+        if (item.style === "XCTest") {
+            const target = swiftPackage
+                .getTargets(TargetType.test)
```
award999, line 139: suggestion: maybe we fetch this list in "transform" so we're not filtering each time the map callback is called

plemarquand: Any given test might be in a different test target, based on its location. This means we need to check each test's target. A cached lookup of locations to test targets may improve performance a bit but there wont likely be a lot of test targets to look through in a given project.

src/TestExplorer/LSPTestDiscovery.ts
plemarquand: Good call, I've refactored out the common capability check into a method.

award999 (commented):
looks good to me, but probably should get another set of eyes

src/TestExplorer/LSPTestDiscovery.ts
```diff
@@ -29,104 +34,118 @@ import { LanguageClientManager } from "../sourcekit-lsp/LanguageClientManager";
...
+     * Returns `true` if the LSP supports the supplied `method` at or
+     * above the supplied `minVersion`.
+     */
+    private checkExperimentalCapability(
+        client: LanguageClient,
+        method: string,
+        minVersion: number
+    ) {
+        const experimentalCapability = client.initializeResult?.capabilities.experimental;
+        if (!experimentalCapability) {
```
adam-fowler, line 99: I did not know about this. It would be good if Microsoft actually documented the language client APIs. Is the availability of the workspace/tests and textDocument/tests something we can cache?

src/TestExplorer/LSPTestDiscovery.ts
```diff
@@ -29,104 +34,118 @@ import { LanguageClientManager } from "../sourcekit-lsp/LanguageClientManager";
...
+        }
+        const targetCapability = experimentalCapability[method];
+        return (targetCapability?.version ?? -1) >= minVersion;
+    }
+
+    /**
+     * Convert from `LSPTestItem[]` to `TestDiscovery.TestClass[]`,
+     * updating the format of the location.
+     */
+    private transform(
```
adam-fowler, line 110: This seems like a very generic method name. `transformToTestClass` maybe?

src/TestExplorer/TestExplorer.ts
```diff
@@ -147,18 +149,27 @@ export class TestExplorer {
             if (isPathInsidePath(uri.fsPath, folder.folder.fsPath)) {
```
adam-fowler, line 149: Do we want to run this on every document symbols response, or when the active file is saved?
EDIT: although I see what you are doing, you are falling back to document symbols if LSP fails

src/TestExplorer/TestDiscovery.ts
```diff
@@ -45,153 +31,123 @@ export interface TestTarget {
...
+            collection.delete(testItem.id);
         }
+    }
+
+    // Skip removing tests if the test explorer is empty
+    if (testController.items.size !== 0) {
+        testController.items.forEach(removeOldTests);
+    }
+
+    // Add/update the top level test items. upsertTestItem will decend the tree of children adding them as well.
```
adam-fowler, line 88:
```suggestion
    // Add/update the top level test items. upsertTestItem will descend the tree of children adding them as well.
```

src/TestExplorer/LSPTestDiscovery.ts
```diff
@@ -29,104 +34,118 @@ import { LanguageClientManager } from "../sourcekit-lsp/LanguageClientManager";
...
-                        location: client.protocol2CodeConverter.asLocation(item.location),
-                        functions: functions,
-                    };
-                });
-            return classes;
+            // Only use the lsp for this request if it supports the
+            // workspace/tests method, and is at least version 2.
+            if (this.checkExperimentalCapability(client, workspaceTestsRequest.method, 2)) {
+                const tests = await client.sendRequest(workspaceTestsRequest, {}, token);
+                const testsInWorkspace = tests.filter(item =>
```
adam-fowler, line 75: Have you found a situation where the tests returned weren't in the workspace? Also `updateTestsFromClasses` does this for individual workspace folders later on

src/TestExplorer/TestExplorer.ts
```diff
@@ -147,18 +149,27 @@ export class TestExplorer {
...
                 const target = folder.swiftPackage.getTarget(uri.fsPath);
                 if (target && target.type === "test") {
```
adam-fowler, line 151: I guess for swift-testing we will need to remove this condition given `swift-testing` tests can be found in any target.

adam-fowler (commented):
This looks great. Left a few comments about minor issues

src/TestExplorer/LSPTestDiscovery.ts
plemarquand: The capabilities are actually cached on the `client` right after the connection is established to the server.

adam-fowler: I actually meant caching in a boolean, instead of running this function. It's a small amount of code, but it gets run with every call

src/TestExplorer/TestExplorer.ts
plemarquand: I think we want to run this on every response, as I'll often be typing out a new test declaration and it will show up immediately instead of waiting until save.

And yep, if `getDocumentTests` fails because the LSP isn't supported we fall back to the old way.

adam-fowler: I guess `textDocument/tests` shouldn't be too heavy a request for the LSP server

src/TestExplorer/LSPTestDiscovery.ts
plemarquand: I think you're right. I kept this filter from the old code but I dont think the LSP can return tests outside the workspace it was initialized with. I'll remove it.

src/TestExplorer/LSPTestDiscovery.ts
plemarquand: Ah ok, I've cached the result in a Map keyed off the capability name + version.

src/TestExplorer/TestExplorer.ts
grynspan: At this time, we would only want to support running tests in test targets. We document that requirement [here](https://swiftpackageindex.com/apple/swift-testing/0.7.0/documentation/testing/definingtests#app-top) (see the first **Note** callout.)

adam-fowler (approved):

adam-fowler:
Before merging this I thought I'd test swift 5.10 and it looks like the document symbols loading isn't working

adam-fowler:
Ah I see what you have done. By converting to the `swift test list` output, you lose the location information that comes with the document symbols. This will need to be fixed before this PR is merged

PR #771, opened 2024-04-29
Add extra dash to llvm-cov arguments

plemarquand:
The llvm-cov that ships with Xcode requires two dashes on arguments.


src/coverage/LcovResults.ts
```diff
@@ -65,8 +65,8 @@ export class LcovResults implements vscode.Disposable {
                     "-format",
```
adam-fowler, line 65: What about `-format`?

adam-fowler (approved):

PR #773, opened 2024-04-29
Fix missing test locations on toolchains 5.10 and below

plemarquand:
If the LSP is unavailable `swift test list` will be used to enumerate the tests. This method cannot attach a location to the TestClasses.

The code pre #767 took this in to account, but #767 introduced a regression that would overwrite an existing item enumerated by document symbols which had a location with one that didn't.

### DIFF
--- a/src/TestExplorer/TestDiscovery.ts
+++ b/src/TestExplorer/TestDiscovery.ts
@@ -89,10 +89,6 @@ export function updateTests(
     testItems.forEach(testItem => {
         upsertTestItem(testController, testItem);
     });
-
-    testController.items.forEach(item => {
-        testController.items.add(item);
-    });
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/TestExplorer/TestDiscovery.ts
```diff
@@ -89,10 +89,6 @@ export function updateTests(
     testItems.forEach(testItem => {
         upsertTestItem(testController, testItem);
     });
-
-    testController.items.forEach(item => {
-        testController.items.add(item);
-    });
```
adam-fowler, line 95: Surprised I missed this. What was this meant to be doing?

adam-fowler (approved):
Change looks good.

src/TestExplorer/TestDiscovery.ts
plemarquand: I left this in accidentally, it was effectively a no-op

PR #775, opened 2024-05-01
Add support for swift-testing

plemarquand:
Support running swift-testing tests the same as XCTests. If a test run has both types, swift-testing tests will be run first followed by XCTests.

First a list of XCTests and swift-testing tests to run is parsed from the test request. Test type is determined by a tag on each `vscode.TestItem[]`, either `"XCTest"` or `"swift-testing"`.

swift-testing tests are launched by running the binary named <PackageName>PackageTests.swift-testing inside the build debug folder. This binary is run with the `--experimental-event-stream-output` flag which forwards test events (test started, complete, issue recorded, etc) to a named pipe. The `SwiftTestingOutputParser` watches this pipe for events as the tests are being run and translates them in to `ITestRunner` calls to record test progress in VSCode.

There are different named pipe reader implementations between macOS/Linux and Windows.

### DIFF
--- a/src/TestExplorer/TestDiscovery.ts
+++ b/src/TestExplorer/TestDiscovery.ts
@@ -16,7 +16,7 @@ import * as vscode from "vscode";
 import { FolderContext } from "../FolderContext";
 import { TargetType } from "../SwiftPackage";
 import { LSPTestItem } from "../sourcekit-lsp/lspExtensions";
-import { reduceTestItemChildren } from "../utilities/utilities";
+import { reduceTestItemChildren } from "./TestUtils";
 
 /** Test class definition */
 export interface TestClass extends Omit<Omit<LSPTestItem, "location">, "children"> {
@@ -177,12 +177,14 @@ function upsertTestItem(
     newItem.label = testItem.label;
     newItem.range = testItem.location?.range;
 
-    // Sort test items by path:line_number to provide a stable sort order.
-    // This prevents test items from changing their order as items are removed and readded.
-    if (testItem.location) {
-        const line = `${testItem.location.range.start.line}`.padStart(8, "0");
-        newItem.sortText = `${testItem.location.uri.fsPath}:${line}`;
-    } else {
+    if (testItem.sortText) {
+        newItem.sortText = testItem.sortText;
+    } else if (testItem.location) {
+        // Sort test items by path:line_number to provide a stable sort order.
+        // This prevents test items from changing their order as items are removed and readded.
+        // const line = `${testItem.location.range.start.line}`.padStart(8, "0");
+        // newItem.sortText = `${testItem.location.uri.fsPath}:${line}`;
+    } else if (!testItem.location) {
         // TestItems without a location should be sorted to the top.
         const zeros = ``.padStart(8, "0");
         newItem.sortText = `${zeros}:${testItem.label}`;
--- a/src/TestExplorer/TestDiscovery.ts
+++ b/src/TestExplorer/TestDiscovery.ts
@@ -16,6 +16,7 @@ import * as vscode from "vscode";
 import { FolderContext } from "../FolderContext";
 import { TargetType } from "../SwiftPackage";
 import { LSPTestItem } from "../sourcekit-lsp/lspExtensions";
+import { reduceTestItemChildren } from "../utilities/utilities";
 
 /** Test class definition */
 export interface TestClass extends Omit<Omit<LSPTestItem, "location">, "children"> {
@@ -76,7 +77,11 @@ export function updateTests(
             (!filterFile || testItem.uri?.fsPath === filterFile.fsPath)
         ) {
             const collection = testItem.parent ? testItem.parent.children : testController.items;
-            collection.delete(testItem.id);
+
+            // TODO: This needs to take in to account parameterized tests with no URI, when they're added.
+            if (testItem.children.size === 0) {
+                collection.delete(testItem.id);
+            }
         }
     }
 
@@ -114,6 +119,24 @@ function createIncomingTestLookup(
 }
 
 /**
+ * Merges the TestItems recursively from the `existingItem` in to the `newItem`
+ */
+function deepMergeTestItemChildren(existingItem: vscode.TestItem, newItem: vscode.TestItem) {
+    reduceTestItemChildren(
+        existingItem.children,
+        (collection, testItem: vscode.TestItem) => {
+            const existing = collection.get(testItem.id);
+            if (existing) {
+                deepMergeTestItemChildren(existing, testItem);
+            }
+            collection.add(testItem);
+            return collection;
+        },
+        newItem.children
+    );
+}
+
+/**
  * Updates the existing `vscode.TestItem` if it exists with the same ID as the `TestClass`,
  * otherwise creates an add a new one. The location on the returned vscode.TestItem is always updated.
  */
@@ -142,10 +165,28 @@ function upsertTestItem(
         newItem = existingItem;
     }
 
+    // At this point all the test items that should have been deleted are out of the tree.
+    // Its possible we're dropping a whole branch of test items on top of an existing one,
+    // and we want to merge these branches instead of the new one replacing the existing one.
+    if (existingItem) {
+        deepMergeTestItemChildren(existingItem, newItem);
+    }
+
     // Manually add the test style as a tag so we can filter by test type.
     newItem.tags = [{ id: testItem.style }, ...testItem.tags];
     newItem.label = testItem.label;
     newItem.range = testItem.location?.range;
+
+    // Sort test items by path:line_number to provide a stable sort order.
+    // This prevents test items from changing their order as items are removed and readded.
+    if (testItem.location) {
+        const line = `${testItem.location.range.start.line}`.padStart(8, "0");
+        newItem.sortText = `${testItem.location.uri.fsPath}:${line}`;
+    } else {
+        // TestItems without a location should be sorted to the top.
+        const zeros = ``.padStart(8, "0");
+        newItem.sortText = `${zeros}:${testItem.label}`;
+    }
 
     // Performs an upsert based on whether a test item exists in the collection with the same id.
     // If no parent is provided operate on the testController's root items.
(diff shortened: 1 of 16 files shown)


src/debugger/launch.ts
```diff
@@ -169,13 +169,113 @@ export function createSnippetConfiguration(
...
+    } else {
+        // On Linux, just run the .xctest executable from the configured build directory.
+        return {
+            ...baseConfig,
+            program: path.join(
+                buildDirectory,
+                "debug",
+                ctx.swiftPackage.name + "PackageTests.xctest"
+            ),
+            env: testEnv,
```
adam-fowler, line 267: This should be `swift-testing`. Also `--experimental-event-stream-output` isn't setup.

src/debugger/launch.ts
```diff
@@ -169,13 +169,113 @@ export function createSnippetConfiguration(
     };
 }
 
+export function createSwiftTestConfiguration(
```
adam-fowler, line 172: This seems very similar to `createXCTestConfiguration` can we combine these?

src/TestExplorer/TestRunner.ts
```diff
@@ -630,6 +733,113 @@ export class TestRunner {
...
+            if (testList.length > 0) {
+                testFilterArg = args.flatMap(arg => ["--filter", regexEscapedString(arg)]);
+            }
+
+            testBuildConfig.args = [...testBuildConfig.args, ...testFilterArg];
+            testBuildConfig.terminal = "console";
+
+            return testBuildConfig;
+        } else {
+            const testBuildConfig = createSwiftTestConfiguration(folderContext, fifoPipePath, true);
```
adam-fowler, line 824: `mkfifo` isn't called here. Is it cleaned up anywhere? You could use `TemporaryFolder.withTemporaryFile` to create a temporary file name, which will get deleted once you return found the supplied closure.

adam-fowler (commented):
Some initial comments

src/TestExplorer/TestRunner.ts
plemarquand: I ended up just `rm`ing the pipe at the end of the test execution because TemporaryFolder.withTemporaryFile's API didn't fit quite right for windows pipes.

src/debugger/launch.ts
plemarquand: Fixed

plemarquand:
Command line test arguments still need to be consolidated down to suites if all items in a suite are included in a a test run. Currently all the tests to be run are passed as individual arguments, which can overrun the command line length.

src/TestExplorer/TestRunner.ts
adam-fowler: Yeah there are cases where you exit the function before the fifo is deleted (if cancellation has been requested). Also some of those functions called might throw errors. I'd rather we did something safer with a similar style to `withTemporaryFile`. 
```ts
// do platform specific setup of fifo and run function and then remove fifo
withTemporaryFifo(fifoPath => {
    const testBuildConfig =
                await LaunchConfigurations.createLaunchConfigurationForSwiftTesting(
    etc
});
```

src/utilities/utilities.ts
```diff
@@ -321,3 +321,37 @@ export function expandFilePathTilda(filepath: string): string {
...
+        }
+    }
+    return result;
+}
+
+/**
+ * An implementation of `reduce()` that operates on a vscode.TestItemCollection,
+ * which only exposes `forEach` for iterating its items
+ */
+export function reduceTestItemChildren<U>(
```
adam-fowler, line 347: It's a little odd this is in utilities.ts. Is not one of the test explorer files better?

src/TestExplorer/TestDiscovery.ts
```diff
@@ -148,11 +165,29 @@ function upsertTestItem(
...
+    }
+
     // Manually add the test style as a tag so we can filter by test type.
     newItem.tags = [{ id: testItem.style }, ...testItem.tags];
     newItem.label = testItem.label;
     newItem.range = testItem.location?.range;
 
+    // Sort test items by path:line_number to provide a stable sort order.
+    // This prevents test items from changing their order as items are removed and readded.
+    if (testItem.location) {
```
adam-fowler, line 182: Is this necessary? In the UI it sorts them at least by line number once you set the location.

src/TestExplorer/TestParsers/SwiftTestingOutputParser.ts
```diff
@@ -0,0 +1,186 @@
...
+    private completionMap = new Map<number, boolean>();
+
+    /**
+     * Watches for test events on the named pipe at the supplied path.
+     * As events are read they are parsed and recorded in the test run state.
+     */
+    public async watch(
+        path: string,
+        runState: ITestRunState,
+        pipeReader?: INamedPipeReader
```
adam-fowler, line 119: Is this just for testing?

src/TestExplorer/TestRunner.ts
```diff
@@ -208,6 +157,8 @@ export class TestRunner {
             // will rebuild everything again
             if (testKind !== TestKind.coverage) {
                 const task = await getBuildAllTask(this.folderContext);
+                task.definition.dontTriggerTestDiscovery = true;
```
adam-fowler, line 160: Is this still needed in pre 6.0 where we are getting the test list from `swift test list`

src/TestExplorer/TestRunner.ts
```diff
@@ -233,138 +184,129 @@ export class TestRunner {
...
     }
 
-    /** Run test session without attaching to a debugger */
-    async runSession(
-        token: vscode.CancellationToken,
+    private async launchTests(
         testKind: TestKind,
+        token: vscode.CancellationToken,
+        stdout: stream.Writable,
+        stderr: stream.Writable,
```
adam-fowler, line 292: Calling them stdout and stderr is a little confusing as the first one is the output stream that is parsed and the second one is the output stream that is just passed onto the test runner output. On macOS the parsed output stream is stderr, on Linux it is stdout. Well as least for XCTest

src/TestExplorer/TestRunner.ts
```diff
@@ -233,138 +184,129 @@ export class TestRunner {
...
-                    );
+                    await this.runCoverageSession(token, stdout, stderr, testBuildConfig);
                     break;
                 case TestKind.parallel:
-                    await this.runParallelSession(
-                        token,
-                        parsedOutputStream,
-                        outputStream,
-                        testBuildConfig
-                    );
```
adam-fowler, line 352: Both parallel and coverage sessions don't seem to run swift testing tests

adam-fowler (commented):
Its pretty much there, a few more comments
I haven't looked at the tests in any detail yet

src/TestExplorer/TestRunner.ts
plemarquand: This will come in a subsequent patch as there are some changes in SwiftPM required to support our coverage workflow. Specifically we need the `--enable-code-coverage` flag to be added to `swift build` (see https://github.com/apple/swift-package-manager/pull/7518). 

I have the code written, I just need to test the SwiftPM update with a nightly toolchain, which I should be able to do shortly.

src/TestExplorer/TestParsers/SwiftTestingOutputParser.ts
plemarquand: Yep, see https://github.com/plemarquand/vscode-swift/blob/0214e8dd8e7f314b1f54fbb01ee5568e05fbcada/test/suite/testexplorer/SwiftTestingOutputParser.test.ts#L27 for the test implementation

src/TestExplorer/TestRunner.ts
plemarquand: I actually think this can be removed

src/TestExplorer/TestRunner.ts
grynspan: Agreed—that would help encapsulate the platform-specific FIFO code too.

src/TestExplorer/TestRunner.ts
```diff
@@ -233,105 +188,122 @@ export class TestRunner {
...
-            }
-        } else {
-            const testBuildConfig = createTestConfiguration(this.folderContext, true);
+
+                // Watch the pipe for JSONL output and parse the events into test explorer updates.
+                // The await simply waits for the watching to be configured.
+                await this.swiftTestOutputParser.watch(fifoPipePath, runState);
+
+                await this.launchTests(
+                    testKind,
```
adam-fowler, line 242: Given parallel test profile is specific to the XCTests (Swift testing already runs in parallel). Perhaps this should not use the parallel test path here.
```suggestion
                    testKind === TestKind.parallel ? TestKind.standard : testKind,
```

grynspan: swift-testing honours parallel/no-parallel options at the command line. The default is reversed from XCTest. Dunno if that affects your suggestion or not.

plemarquand: Good idea. Eventually we'll want to offer running swift-testing tests with no-parallel, but for now I think it makes sense to run them as their default.

adam-fowler: @grynspan this was more about getting swift-testing working when someone chooses parallel. The parallel path at the moment breaks swift-testing. Eventually we should have the parallel/non-parallel setting work for both XCTest and testing. 

Parallel for XCTest doesn't work great in VSCode (result only available at end, error parsing broken prior to Swift 6) so wasn't keen on making it the default. Meanwhile swift-testing parallel works great so would be a shame to not make it the default. Perhaps we should have two settings.

adam-fowler (approved):
I think we are good.
I see this is no longer a draft, has the related swift-testing PR been merged

grynspan:
I can tag 0.9 any time, although I've been trying to maintain a beginning-of-the-month cadence.

adam-fowler:
@grynspan We have a few other things to finish before doing the next release of the extension, so you can probably hold off for a two weeks or so before doing a swift-testing release.

PR #779, opened 2024-05-03
Use c99name for test target IDs

plemarquand:
The test target ID provided to XCTest needs to be a valid Swift identifier. The test target name is not guarenteed to be valid.

Instead, use the c99name, which is the name converted to a c identifier, which is also a valid swift identifier.


award999 (approved):
looks good to me

adam-fowler (approved):
LGTM

PR #786, opened 2024-05-09
Setup cwd correctly for plugin tasks

adam-fowler:
There are two different current working folders in a Swift Plugin task. One is the one inside the tasks definition. This should always be relative to the workspace folder, so the task.json can be shared across multiple machines. The other is the actual current working folder to run the executable in.

This PR splits the two.


award999 (approved):
LGTM. In a separate commit maybe we should add a test asset project if don't have one and add a test

award999 (approved):

PR #792, opened 2024-05-10
Add Code Snippets

plemarquand:
Augment the built in collection of Swift snippets with some common snippets that were not included. Also adds snippets for creating test suites and cases in both XCTest and swift-testing.

### DIFF
--- a/snippets/swift-testing.code-snippets
+++ b/snippets/swift-testing.code-snippets
@@ -3,7 +3,7 @@
     "prefix": "@Test",
     "body": [
       "@Test func ${1:name}() async throws {",
-      "\t$2",
+      "\t${2:statements}",
       "}"
     ],
     "description": "swift-testing: Test Function"
@@ -13,7 +13,7 @@
     "body": [
       "@Test(\"${1:display name}\")",
       "func ${2:name}() async throws {",
-      "\t$3",
+      "\t${3:statements}",
       "}"
     ],
     "description": "swift-testing: Test Function with Traits"
@@ -23,7 +23,7 @@
     "body": [
       "struct ${1:suiteName} {",
       "\t@Test func ${2:testName}() async throws {",
-      "\t\t$3",
+      "\t\t${3:statements}",
       "\t}",
       "}"
     ],
@@ -35,7 +35,7 @@
       "@Suite(\"${1:display name}\")",
       "struct ${2:name} {",
       "\t@Test func ${3:testName}() async throws {",
-      "\t\t$4",
+      "\t\t${4:statements}",
       "\t}",
       "}"
     ],
--- a/snippets/swift-testing.code-snippets
+++ b/snippets/swift-testing.code-snippets
@@ -6,7 +6,7 @@
       "\t$2",
       "}"
     ],
-    "description": "Swift Testing: Test Function"
+    "description": "swift-testing: Test Function"
   },
   "@Test(\"named\")": {
     "prefix": "@Test",
@@ -16,7 +16,7 @@
       "\t$3",
       "}"
     ],
-    "description": "Swift Testing: Test Function With Traits"
+    "description": "swift-testing: Test Function with Traits"
   },
   "@Suite": {
     "prefix": "@Suite",
@@ -27,7 +27,7 @@
       "\t}",
       "}"
     ],
-    "description": "Swift Testing: Test Function"
+    "description": "swift-testing: Test Suite"
   },
   "@Suite(\"named\")": {
     "prefix": "@Suite",
@@ -39,7 +39,7 @@
       "\t}",
       "}"
     ],
-    "description": "Swift Testing: Test Function With Traits"
+    "description": "swift-testing: Test Suite with Traits"
   },
   "tag": {
     "prefix": "tag",
@@ -48,6 +48,6 @@
       "\t@Tag static var ${1:tag}: Tag",
       "}"
     ],
-    "description": "Swift Testing: Tag Declaration"
+    "description": "swift-testing: Tag Declaration"
   }
 }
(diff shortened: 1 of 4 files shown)


snippets/swift-testing.code-snippets
```diff
@@ -0,0 +1,53 @@
...
+  "@Suite": {
+    "prefix": "@Suite",
+    "body": [
+      "struct ${1:suiteName} {",
+      "\t@Test func ${2:testName}() async throws {",
+      "\t\t$3",
+      "\t}",
+      "}"
+    ],
+    "description": "Swift Testing: Test Function"
```
stmontgomery:
```suggestion
    "description": "Swift Testing: Suite Struct"
```

plemarquand: Done

snippets/swift-testing.code-snippets
```diff
@@ -0,0 +1,53 @@
...
+    "body": [
+      "@Test(\"${1:display name}\")",
+      "func ${2:name}() async throws {",
+      "\t$3",
+      "}"
+    ],
+    "description": "swift-testing: Test Function with Traits"
+  },
+  "@Suite": {
+    "prefix": "@Suite",
```
stmontgomery, line 22: This one is a little different because the snippet body doesn't begin with `@`—it doesn't have the `@Suite` attribute explicitly, unlike the "suite _with traits_" snippet which does. Should its prefix be simpler like just "suite"?

snippets/xctest.code-snippets
```diff
@@ -0,0 +1,22 @@
+{
+  "XCTestSuite": {
+    "prefix": "testsuite",
+    "body": [
+      "class ${1:suiteName}: XCTestCase {",
+      "\tfunc test_${2:testName}() {",
```
stmontgomery, line 6: Underscore intentional? I'd typically suggest
```suggestion
      "\tfunc test${2:testName}() {",
```

snippets/xctest.code-snippets
```diff
@@ -0,0 +1,22 @@
...
+    "description": "XCTest: Test Suite"
+  },
+  "XCTest": {
+    "prefix": "test",
+    "body": [
+      "func test${1:Name}() {",
+      "\t$2",
+      "}"
+    ],
+    "description": "XCTest: Test Function"
```
stmontgomery, line 20:
```suggestion
    "description": "XCTest: Test Method"
```
"method" is generally the preferred term in the context of XCTest, since it only supports tests as methods. In contrast, swift-testing supports global functions so we've been using "function" there.

snippets/swift-testing.code-snippets
plemarquand: On the one hand if they both has the prefix `@Suite` you can start typing `@Suite` and then use the arrow keys to pick whether you want the named or unnamed version. But on the other hand you're right, you're not actually annotating with @Suite in the unnamed one. I'll try out both and see what feels better.

plemarquand: I've renamed the snippet name to be Suite, but kept the prefix as `@Suite`. This lets you easily pick between the named/unnamed with you type `@Suite` and still have the unnamed version come up in the list when you type `suite`

stmontgomery (approved):
My review is focused only on the swift-testing and xctest snippet files, but this LGTM!

snippets/swift.code-snippets
```diff
@@ -0,0 +1,123 @@
+{
+  "API Availability Check": {
```
adam-fowler, line 2: We need a @available version of well

adam-fowler (commented):
A number of the swift code snippets seem more work to use than actually doing the typing. eg defer, deinit, import, var, let, init

snippets/swift.code-snippets
```diff
@@ -0,0 +1,123 @@
...
+  "Deinitialization Declaration": {
+    "prefix": "deinit",
+    "body": [
+      "deinit {",
+      "\t${1:deferred statements}",
+      "}",
+    ],
+    "description": "Performs cleanup before an object is deallocated."
+  },
+  "Deprecated": {
```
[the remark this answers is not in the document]
plemarquand: `Deprecated` now just has a message. Added `Renamed`.

snippets/swift.code-snippets
plemarquand: Done

snippets/swift.code-snippets
```diff
@@ -0,0 +1,123 @@
+{
+  "API Availability Check": {
+    "prefix": "available",
```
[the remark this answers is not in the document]
plemarquand: I agree. I've removed the hardcoded `OS` string so code completion works a bit more naturally and you dont have to manually remove it after picking an option.

plemarquand:
@adam-fowler For the basic Swift ones I was trying to fill the gap between what what was missing in the VSCode Swift language support extension and what is present in Xcode. I agree they're not that useful day to day but could help a beginner, maybe. I'm not committed to keeping them if you'd rather omit them.

snippets/swift.code-snippets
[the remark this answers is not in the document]
adam-fowler: Given they are in Xcode we'll leave them `available` snippets work nice

snippets/swift-testing.code-snippets
```diff
@@ -0,0 +1,53 @@
+{
+  "@Test": {
+    "prefix": "@Test",
+    "body": [
+      "@Test func ${1:name}() async throws {",
+      "\t$2",
```
adam-fowler, line 6: To keep this consistent with standard snippets perhaps it should be ${2:statements}
Similarly for all the other test function snippets

plemarquand: Done

adam-fowler (approved):
👍

PR #795, opened 2024-05-10
Handle having multiple Swift toolchains on Windows

plemarquand:
It is possible for `which swift` to return multiple lines if the user has more than one Swift toolchain installed. Windows behavior is to use the first one, so we use it as well but also warn the user.

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -285,7 +285,13 @@ export class SwiftToolchain {
                     }
                     case "win32": {
                         const { stdout } = await execFile("where", ["swift"]);
-                        swift = stdout.trimEnd();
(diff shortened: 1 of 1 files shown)


adam-fowler:
@compnerd, @tristanlabelle  what are your thoughts on this?

compnerd:
`where` on windows will return all the paths. Warning about the ambiguity makes sense.

The runtime is not ABI stable, so anything other than the single version installation seems fragile - there's no guarantee that the right runtime is being used. There's a desire to support multiple versions, which is why the installer has gained support for that, but the work to support parallel installs is incomplete.

This seems to be something which would be nice to support, indicating ambiguity seems like a positive, but we still have more work to do.

award999 (approved):

adam-fowler (approved):

PR #796, opened 2024-05-11
Only disable build tasks in swift versions less than v6

adam-fowler:
SwiftPM in Swift 6 and later does not allow for multiple processes to access the build folder and stalls them until it is available. So we don't need to disable build tasks.

I also added a relative path to the `Build tasks disabled` text when it is needed in a similar manner to the actual build tasks.


plemarquand (approved):

plemarquand:
Looks good to me, this is a nice little user experience improvement

PR #797, opened 2024-05-13
Support parameterized swift-testing tests

plemarquand:
Swift-testing emits a `test` event at the beginning of a test run that enumerates all the parameterized test cases to be executed.

This patch waits for the `test` event and then generates `vscode.TestItem`s for each parameterized test execution, parenting them to their test. Then when we recieve a `runStarted` event we enqueue all the tests along with the newly created parameterized `TestItem`s.

Before a new test run starts we remove the existing parameterized `TestItems` so we can regenerate them, as there may be a different number of parameterized tests run with every execution.

### DIFF
--- a/src/TestExplorer/TestParsers/SwiftTestingOutputParser.ts
+++ b/src/TestExplorer/TestParsers/SwiftTestingOutputParser.ts
@@ -1,3 +1,4 @@
+import * as vscode from "vscode";
 import * as readline from "readline";
 import { Readable } from "stream";
 import {
@@ -6,6 +7,7 @@ import {
     WindowsNamedPipeReader,
 } from "./TestEventStreamReader";
 import { ITestRunState } from "./TestRunState";
+import { TestClass } from "../TestDiscovery";
 
 // All events produced by a swift-testing run will be one of these three types.
 export type SwiftTestEvent = MetadataRecord | TestRecord | EventRecord;
@@ -21,7 +23,7 @@ interface MetadataRecord extends VersionedRecord {
 
 interface TestRecord extends VersionedRecord {
     kind: "test";
-    payload: Test;
+    payload: TestSuite | TestFunction;
 }
 
 export type EventRecordPayload =
@@ -43,15 +45,23 @@ interface Metadata {
     [key: string]: object; // Currently unstructured content
 }
 
-interface Test {
-    kind: "suite" | "function" | "parameterizedFunction";
+interface TestBase {
     id: string;
     name: string;
     _testCases?: TestCase[];
     sourceLocation: SourceLocation;
 }
 
-interface TestCase {
+interface TestSuite extends TestBase {
+    kind: "suite";
+}
+
+interface TestFunction extends TestBase {
+    kind: "function";
+    isParameterized: boolean;
+}
+
+export interface TestCase {
     id: string;
     displayName: string;
 }
@@ -76,6 +86,11 @@ interface BaseEvent {
     testID: string;
 }
 
+interface TestCaseEvent {
+    sourceLocation: SourceLocation;
+    _testCase: TestCase;
+}
+
 interface TestStarted extends BaseEvent {
     kind: "testStarted";
 }
@@ -84,11 +99,11 @@ interface TestEnded extends BaseEvent {
     kind: "testEnded";
 }
 
-interface TestCaseStarted extends BaseEvent {
+interface TestCaseStarted extends BaseEvent, TestCaseEvent {
     kind: "testCaseStarted";
 }
 
-interface TestCaseEnded extends BaseEvent {
+interface TestCaseEnded extends BaseEvent, TestCaseEvent {
     kind: "testCaseEnded";
 }
 
@@ -96,7 +111,7 @@ interface TestSkipped extends BaseEvent {
     kind: "testSkipped";
 }
 
-interface IssueRecorded extends BaseEvent {
+interface IssueRecorded extends BaseEvent, TestCaseEvent {
     kind: "issueRecorded";
     issue: {
         sourceLocation: SourceLocation;
@@ -115,6 +130,12 @@ export interface SourceLocation {
 
 export class SwiftTestingOutputParser {
     private completionMap = new Map<number, boolean>();
+    private testCaseMap = new Map<string, Map<string, TestCase>>();
+
+    constructor(
+        public testRunStarted: () => void,
+        public addParameterizedTestCase: (testClass: TestClass, parentIndex: number) => void
+    ) {}
 
     /**
      * Watches for test events on the named pipe at the supplied path.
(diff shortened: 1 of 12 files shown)


adam-fowler:
This doesn't appear to work. SKLSP is returning parameterized tests as `funcName(_ :)`, while swift-testing is expecting `funcName(_:)` which doesn't include the space.

plemarquand:
@adam-fowler sorry I forgot to mention this depends on https://github.com/apple/sourcekit-lsp/pull/1209. For now to test you'd need to build your own sourcekit-lsp with that change included, as it looks like the nightly toolchain builds aren't working.

adam-fowler:
Outside of the fact I can't compile, other issues include I cannot run a test suite that includes a parameterised test, and deleting tests doesn't always delete the test in the test explorer.

src/TestExplorer/TestParsers/SwiftTestingOutputParser.ts
```diff
@@ -155,31 +176,131 @@ export class SwiftTestingOutputParser {
...
+    }
+
+    // Test cases do not have a unique ID if their arguments are not serializable
+    // with Codable. If they aren't, their id appears as `argumentIDs: nil`, and we
+    // fall back to using the testCase display name as the test case ID. This isn't
+    // ideal because its possible to have multiple test cases with the same display name,
+    // but until we have a better solution for identifying test cases it will have to do.
+    // SEE: rdar://119522099.
+    private createTestCaseId(testCase: TestCase): string {
+        return testCase.id === "argumentIDs: nil" ? testCase.displayName : testCase.id;
```
award999, line 191: question: Where does the "argumentIDs: nil" constant come from?

suggestion: Maybe add a comment so we know where we can check to make sure it is in sync

src/TestExplorer/TestParsers/SwiftTestingOutputParser.ts
```diff
@@ -155,31 +176,131 @@ export class SwiftTestingOutputParser {
...
+                    this.parameterizedFunctionTestCaseToTestClass(
+                        testCase,
+                        sourceLocationToVSCodeLocation(item.payload.sourceLocation),
+                        index
+                    )
+                )
+                .flatMap(testClass => (testClass ? [testClass] : []))
+                .forEach(testClass => this.addParameterizedTestCase(testClass, testIndex));
+        } else if (item.kind === "event") {
+            if (item.payload.kind === "runStarted") {
```
award999, line 246: suggestion: again maybe some info about where these payload kinds come from and what they mean

src/TestExplorer/TestRunner.ts
```diff
@@ -53,9 +54,157 @@ export enum TestKind {
...
+        } else {
+            this.queuedOutput.push(output);
+        }
+    }
+}
+
 /** Class used to run tests */
 export class TestRunner {
-    private testRun: vscode.TestRun;
+    // private testRun: vscode.TestRun;
```
award999, line 206: nit: remove commented out code

award999 (commented):
looking good to me

src/TestExplorer/TestParsers/SwiftTestingOutputParser.ts
plemarquand: This is the ID string for parameterized test cases that have unserializable arguments. Its a bit odd for sure, and liable to change in future swift-testing releases. The comment on this function describes the reason.

adam-fowler (commented):
The code all looks good. Couple of minor code nits
Functionality wise:
- If you delete a parameterised test it doesn't delete it from the test explorer
- Parameterized test items don't have a location
- If I ask one of them to run nothing happens. Ideally it would only run the test with that one case. If you can't do that I'm not sure if it is worth it keeping a test item for every case.

src/TestExplorer/TestRunner.ts
```diff
@@ -421,7 +582,6 @@ export class TestRunner {
                 {
                     cwd: testBuildConfig.cwd,
                     env: { ...process.env, ...testBuildConfig.env, SWT_SF_SYMBOLS_ENABLED: "0" },
-                    maxBuffer: 16 * 1024 * 1024,
```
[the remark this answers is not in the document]
plemarquand: Good catch, I removed this while debugging something else and didn't add it back.

We should be able to remove this soon; the next patch that adds coverage support also moves to using Swift Tasks for these calls.

src/TestExplorer/TestParsers/XCTestOutputParser.ts
```diff
@@ -228,6 +229,10 @@ export class XCTestOutputParser {
         runState.failedTest = undefined;
     }
 
+    private sourceLocationToVSCodeLocation(file: string, lineNumber: number): vscode.Location {
```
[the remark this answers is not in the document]
plemarquand: Done

src/TestExplorer/TestRunner.ts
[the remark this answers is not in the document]
plemarquand: I've added it back.

plemarquand:
> If you delete a parameterised test it doesn't delete it from the test explorer

Fixed, the code to prune the test item tree works bottom up and assumed items being pruned would have no children. Added a check to still prune if all children are parameterized test results.

> Tests with named parameters don't run. Looks like SKLSP is giving you the wrong name for the test eg

I couldn't reproduce this, trying on both macOS and in the Linux dev container. What were you seeing? The tests would run/complete but you wouldn't get new parameterized test results in the Test Explorer?

> Parameterized test items don't have a location

This is by design, since results don't really have a "location" per se. The parent TestItem should always be visible in the Test Explorer even scrolling a long list of items, and the tests location remains available off that. Its also a handy way to distinguish if an item in the TestItem hierarchy is a parameterized test result entry or not.

> If I ask one of them to run nothing happens. Ideally it would only run the test with that one case. If you can't do that I'm not sure if it is worth it keeping a test item for every case.

It isn't possible today, but we can look at it for a subsequent PR.

adam-fowler:
> I couldn't reproduce this, trying on both macOS and in the Linux dev container. What were you seeing? The tests would run/complete but you wouldn't get new parameterized test results in the Test Explorer?
> 
Ok. I think this was a symptom of the first issue. I had a function where the parameter didn't have a label and I edited it to include a label for the parameter, but it never deleted the test items and I was still running with those old test items. Your fix for the deletion of test items for parametized test functions fixes this also.

adam-fowler:
> > If I ask one of them to run nothing happens. Ideally it would only run the test with that one case. If you can't do that I'm not sure if it is worth it keeping a test item for every case.
> 
> It isn't possible today, but we can look at it for a subsequent PR.

Can we at least get it to run the test the parameter belongs to

plemarquand:
> > > If I ask one of them to run nothing happens. Ideally it would only run the test with that one case. If you can't do that I'm not sure if it is worth it keeping a test item for every case.
> > 
> > 
> > It isn't possible today, but we can look at it for a subsequent PR.
> 
> Can we at least get it to run the test the parameter belongs to

I can, but I worry having the behaviour change in a subsequent release would be confusing. In the future it would go from running all the test parameters to running the test with just one. 

Tests with parameter types that aren't Codable wouldn't be able to be run individually, and so they'd lose the play button (or rerun the test with all parameters, but I think that'd be even more confusing).

plemarquand:
@adam-fowler you were right that it didn't make sense for these parameterized test items to have run buttons at all. I've added a "runnable" tag to all the test items that can be run, associated that with the run profiles, and this omits the button on parameterized tests as you'd expect.

https://github.com/swift-server/vscode-swift/assets/667875/542210a0-e5b8-4472-af10-f67e8ec3d200

adam-fowler (approved):
We are good here

PR #807, opened 2024-05-17
Use native VSCode coverage APIs, cover swift-testing tests

plemarquand:
Move to the new VSCode coverage APIs that were introduced in VSCode version 1.88. This allows us to remove the coverage renderer and coverage report files. This functionality is now built in to the test explorer and VSCode's document rendering.

This patch also adds support for producing test coverage from swift-testing tests. XCTests and swift-testing tests are each run in their respective binaries and then the coverage data is merged into a profdata file before being exported to LCOV format.

Finally we transform the LCOV JSON into the coverage structures for VSCode and associate this data with the test run.


src/debugger/buildConfig.ts
```diff
@@ -0,0 +1,317 @@
...
+                    this.fifoPipePath,
+                ];
+
+                if (this.swiftVersionGreaterOrEqual(6, 0, 0)) {
+                    args = [...args, "--disable-xctest"];
+                }
+
+                return {
+                    ...this.baseConfig,
+                    program: this.swiftProgramPath,
```
adam-fowler, line 116: You can't use `swift` executable if you are debugging tests. You have to call xctest or swift-testing executable directly.

src/TestExplorer/TestRunner.ts
```diff
@@ -193,8 +190,15 @@ export class TestRunProxy {
...
 
-    public end() {
-        this.testRun?.end();
+    public async end() {
+        if (!this.testRun) {
+            return;
+        }
+
+        // Compute final coverage numbers if any coverage info has been captured during the run.
+        await this.coverage.computeCoverage(this.testRun);
```
adam-fowler, line 199: How about moving this to where the coverage `createRunProfile` is created instead of calling for every test run

adam-fowler (commented):
I'm not sure how ready this is for review, but I can get the coverage view to appear now which looks nice, I don't get inline rendering of coverage in the text editor though. Outside of coverage, parallel testing crashes VSCode which is quite impressive and debugging tests doesn't work anymore.

I think you are going to need to reimplement the two phase build and test instead of relying on `swift test` to do everything (outside of coverage testing), because you cannot use `swift test` when debugging tests.

src/debugger/buildConfig.ts
```diff
@@ -0,0 +1,425 @@
...
+import { TargetType } from "../SwiftPackage";
+import { Version } from "../utilities/version";
+import { TestKind, TestLibrary } from "../TestExplorer/TestRunner";
+
+/**
+ * Creates `vscode.DebugConfiguration`s for different combinations of
+ * testing library, test kind and platform. Use the static `swiftTestingConfig`
+ * and `xcTestConfig` functions to create
+ */
+export class DebugConfigurationFactory {
```
adam-fowler, line 32: Should this have a name to indicate it is for creating launch config for tests

plemarquand:

plemarquand:
@swift-server-bot test this please

plemarquand:
@adam-fowler thanks for taking just a through look. I think everything should be addressed.

adam-fowler (approved):
Looking good

PR #809, opened 2024-05-17
[DO NOT MERGE] Prototype for SPI integration

daveyc123:
This change is a prototype for SPI integration. It provides two new commands that leverage the new add package workflows within swiftpm
  * Add package
  * Browse Package Index...

Add package is fairly straight forward. It prompts the user for both the package URL and the version, and adds the package to their Package.swift (by calling swift package add...). An alternative here would be to have the LSP do the equivalent refactoring operation.

Browse Package Index opens a webview to the SPI (note in this patch it's using localhost) in an iframe. The intent is that the SPI can detect that it is running through the vscode webview and provide alternative UI for adding a package. This is accomplished through:
  * A window event listener between the frame and the webview in which messages that can be sent [1]
  * Access to the vscode api from the webview that post a message to run the command. This invokes the add package command detailed above

[1] - SPI usage
- SPI would need to detect that it is running in the vscode context and present alternative (or augmented UI). In the example given it could use window.name to check and see if it is equal to "vscode_iframe"
- When the user takes the action to add their package using this new UI, SPI needs to send a message to its parent. window.parent.postMessage( { action: 'addPackage', url: url, version: version }, '*')

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -249,6 +249,147 @@ export async function createNewProject(ctx: WorkspaceContext): Promise<void> {
 }
 
 /**
+ * Prompts the user to input project details and then executes `swift package init`
+ * to create the project.
+ */
+export async function browsePackageIndex(ctx: WorkspaceContext): Promise<void> {
+    // The context key `swift.browsePackageIndex` only works if the extension has been
+    // activated. As such, we also have to allow this command to run when no workspace is
+    // active. Show an error to the user if the command is unavailable.
+    if (!ctx.toolchain.swiftVersion.isGreaterThanOrEqual(new Version(6, 0, 0))) {
+        vscode.window.showErrorMessage(
+            "Creating a new swift project is only available starting in swift version 6.0.0."
+        );
+        return;
+    }
+
+    const panel = vscode.window.createWebviewPanel(
+        "swiftPackageIndex",
+        "Swift Package Index",
+        vscode.ViewColumn.One,
+        {
+            retainContextWhenHidden: true,
+            enableScripts: true,
+        }
+    );
+
+    panel.webview.onDidReceiveMessage(
+        message => {
+            switch (message.command) {
+                case "addSwiftPackage":
+                    addPackageDependency(ctx, message.url, message.version);
+                    return;
+                default:
+                    // Log message
+                    vscode.window.showErrorMessage(`An error occurred when adding packages`);
+                    return;
+            }
+        },
+        undefined,
+        ctx.subscriptions
+    );
+
+    panel.webview.html = `<html>
+        <iframe name="vscode_iframe" id="mainframe" src="http://localhost:8080/" style="position:fixed; top:0; left:0; bottom:0; right:0; width:100%; height:100%; border:none; margin:0; padding:0; overflow:hidden; z-index:999999;">
+            Your browser doesn't support iframes
+        </iframe>
+        <script>
+            const vscode = acquireVsCodeApi();
+            window.addEventListener('message', e => {
+                action = e.data.action;
+                if (action === "addPackage") {
+                    url = e.data.url
+                    version = e.data.version
+                    
+                    if (url && version) {
+                        vscode.postMessage({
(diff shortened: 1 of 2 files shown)


daveyc123:
@finestructure @daveverwer

src/commands.ts
```diff
@@ -248,6 +248,147 @@ export async function createNewProject(ctx: WorkspaceContext): Promise<void> {
...
+ * Prompts the user to input project details and then executes `swift package init`
+ * to create the project.
+ */
+export async function browsePackageIndex(ctx: WorkspaceContext): Promise<void> {
+    // The context key `swift.browsePackageIndex` only works if the extension has been
+    // activated. As such, we also have to allow this command to run when no workspace is
+    // active. Show an error to the user if the command is unavailable.
+    if (!ctx.toolchain.swiftVersion.isGreaterThanOrEqual(new Version(6, 0, 0))) {
+        vscode.window.showErrorMessage(
+            "Creating a new swift project is only available starting in swift version 6.0.0."
```
adam-fowler, line 261: Interesting error? If you do want to gate access to a feature you should use a context key to disable the command

adam-fowler (commented):
Thought I'd try and get this running but nothing shows up 😞 when I run browsePackageIndex

src/commands.ts
daveyc123: Fixed the wording of the text and added a context key to block the command.

daveyc123:
> Thought I'd try and get this running but nothing shows up 😞 when I run browsePackageIndex

Unfortunately it won't be functional until SPI has added support

adam-fowler:
But @daveyc123 it didn't even show the SPI website.

src/commands.ts
```diff
@@ -248,6 +248,147 @@ export async function createNewProject(ctx: WorkspaceContext): Promise<void> {
...
+    // activated. As such, we also have to allow this command to run when no workspace is
+    // active. Show an error to the user if the command is unavailable.
+    if (!ctx.toolchain.swiftVersion.isGreaterThanOrEqual(new Version(6, 0, 0))) {
+        vscode.window.showErrorMessage(
+            "Browsing the package index is only available starting in swift version 6.0.0."
+        );
+        return;
+    }
+
+    const panel = vscode.window.createWebviewPanel(
```
adam-fowler, line 266: Can we break out some of the web view stuff, into a general function that can be used by other systems. Currently when you right click on a package dependency and select `View Repository` it opens a web page in your default browser. Would be cool if we could do that inside VSCode instead in a similar manner to how this works.

src/commands.ts
```diff
@@ -248,6 +248,147 @@ export async function createNewProject(ctx: WorkspaceContext): Promise<void> {
...
 
+/**
+ * Prompts the user to input project details and then executes `swift package init`
+ * to create the project.
+ */
+export async function browsePackageIndex(ctx: WorkspaceContext): Promise<void> {
+    // The context key `swift.addPackageRefactoringAvailable` only works if the extension has been
+    // activated. As such, we also have to allow this command to run when no workspace is
+    // active. Show an error to the user if the command is unavailable.
+    if (!ctx.toolchain.swiftVersion.isGreaterThanOrEqual(new Version(6, 0, 0))) {
```
adam-fowler, line 259: Now you have the context key this shouldn't be necessary

src/commands.ts
```diff
@@ -248,6 +248,147 @@ export async function createNewProject(ctx: WorkspaceContext): Promise<void> {
...
+
+export async function addPackageDependency(
+    ctx: WorkspaceContext,
+    url?: string,
+    version?: string
+): Promise<void> {
+    // The context key `swift.addPackageRefactoringAvailable` only works if the extension has been
+    // activated. As such, we also have to allow this command to run when no workspace is
+    // active. Show an error to the user if the command is unavailable.
+    if (!ctx.toolchain.swiftVersion.isGreaterThanOrEqual(new Version(6, 0, 0))) {
```
adam-fowler, line 330: What's the point of having this command available when you have no package loaded?

PR #814, opened 2024-05-21
Switch to Build All task when listing tests fails

### LINKED ISSUE #808: TestExplorer should not be triggering builds using SwiftExecOperation
If the test list fails then a build is triggered. We can't use SwiftExecOperation for builds they can be long operations. Using SwiftExecOperation means you can't see the result of the build, it is a long process you have no control over. 

Use this instead
```ts
// get buildall task and execute it
const backgroundTask = await getBuildAllTask(this.folderContext);
if (!backgroundTask) {
    return;
}
try {
    await this.folderContext.taskQueue.queueOperation(new TaskOperation(backgroundTask));
} catch {
    // can ignore if running task fails
}    
```

plemarquand:
Use the buildAllTask instead of SwiftExecOperation to make long this behaviour the same as other builds in the extension.

Added a task cache since provideTasks can be called multiple times and the most recently returned task is held by VSCode. If we attempted created the build all task and hooked up the event listeners but then provideTasks was called again before the task was executed, the executed task would not be the one we were listening to.

Fixes #808


award999 (approved):
LGTM

PR #817, opened 2024-05-22
Toolchain selection

### LINKED ISSUE #756: Provide a UI workflow for selecting the swift toolchain
Currently the extension provides a mechanism to discover installed toolchains and automatically sets a toolchain (via the vscode configuration settings).

Users should have the ability to run a command and select from the list of installed toolchains. When a toolchain isn't installed, they should be guided on how to install one.

matthewbastien:
Fixes #756


src/toolchain/toolchain.ts
```diff
@@ -181,12 +182,84 @@ export class SwiftToolchain {
      * @returns Folders for each Xcode install
      */
     public static async getXcodeInstalls(): Promise<string[]> {
+        if (process.platform !== "darwin") {
```
plemarquand, line 185: on line 150 the method `getXcodeDeveloperDir` calls `xcode-select -p`. If I choose another Xcode as my toolchain from within VSCode this still points to my system default toolchain.

src/toolchain/toolchain.ts
```diff
@@ -181,12 +182,84 @@ export class SwiftToolchain {
...
+            throw new Error("Failed to retrieve Swiftly installations from disk.");
+        }
+    }
+
+    /**
+     * Checks common directories for available swift toolchain installations.
+     *
+     * @returns an array of toolchain paths
+     */
+    public static getToolchainInstalls(): Promise<string[]> {
```
plemarquand, line 235: Is this Darwin only?

src/toolchain/toolchain.ts
```diff
@@ -181,12 +182,84 @@ export class SwiftToolchain {
...
+                .filter(
+                    dirent =>
+                        dirent.name.startsWith("swift-") &&
+                        (dirent.isDirectory() || dirent.isSymbolicLink())
+                )
+                .map(dirent => path.join(dirent.path, dirent.name));
+        } catch {
+            // Assume that there are no installations here
+        }
+        return [];
```
plemarquand, line 260: nit: move the return to the `catch` to make it more clear that `[]` is returned for the error condition

src/ui/ToolchainSelection.ts
```diff
@@ -0,0 +1,198 @@
...
+            return {
+                type: "toolchain",
+                label: path.basename(xcodePath, ".app"),
+                detail: xcodePath,
+                toolchainPath,
+                swiftFolderPath: path.join(toolchainPath, "bin"),
+            };
+        });
+    const toolchains = (await SwiftToolchain.getToolchainInstalls())
+        .sort((a, b) => (a > b ? -1 : 1)) // Reverse order
```
plemarquand, line 122:
```suggestion
        .reverse()
```

src/ui/ToolchainSelection.ts
```diff
@@ -0,0 +1,198 @@
...
+        .map<SwiftToolchainItem>(toolchainPath => ({
+            type: "toolchain",
+            label: path.basename(toolchainPath, ".xctoolchain"),
+            detail: toolchainPath,
+            toolchainPath: path.join(toolchainPath, "usr"),
+            swiftFolderPath: path.join(toolchainPath, "usr", "bin"),
+        }));
+    const swiftlyToolchains = (await SwiftToolchain.getSwiftlyToolchainInstalls())
+        .sort((a, b) => (a > b ? -1 : 1)) // Reverse order
```
plemarquand, line 131:
```suggestion
        .reverse()
```

src/ui/ToolchainSelection.ts
```diff
@@ -0,0 +1,198 @@
...
+ * with the user's selection.
+ *
+ * @param ctx the {@link WorkspaceContext}
+ */
+export async function selectToolchain(ctx: WorkspaceContext) {
+    const selected = await vscode.window.showQuickPick<SelectToolchainItem>(
+        getQuickPickItems(ctx),
+        {
+            title: "Select the swift toolchain",
+            placeHolder: "Pick a swift toolchain",
```
plemarquand, line 189:
```suggestion
            title: "Select the Swift toolchain",
            placeHolder: "Pick a Swift toolchain",
```

src/ui/ToolchainSelection.ts
```diff
@@ -0,0 +1,198 @@
...
+/**
+ * Prompt the user to select a folder where they have installed the swift toolchain.
+ * Updates the swift.path configuration with the selected folder.
+ */
+export async function selectToolchainFolder() {
+    const selected = await vscode.window.showOpenDialog({
+        canSelectFiles: false,
+        canSelectFolders: true,
+        canSelectMany: false,
+        title: "Select the folder containing swift binaries",
```
plemarquand, line 46:
```suggestion
        title: "Select the folder containing Swift binaries",
```

src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -113,9 +113,9 @@ export class LanguageClientManager {
     public subFolderWorkspaces: vscode.Uri[];
 
     constructor(public workspaceContext: WorkspaceContext) {
-        this.singleServerSupport = workspaceContext.swiftVersion.isGreaterThanOrEqual(
-            new Version(5, 7, 0)
-        );
+        this.singleServerSupport = workspaceContext.toolchain
+            ? workspaceContext.toolchain.swiftVersion.isGreaterThanOrEqual(new Version(5, 7, 0))
+            : false;
```
plemarquand, line 118:
```suggestion
        this.singleServerSupport = workspaceContext.toolchain?.swiftVersion.isGreaterThanOrEqual(new Version(5, 7, 0)) ?? false;
```

src/toolchain/toolchain.ts
```diff
@@ -181,12 +182,84 @@ export class SwiftToolchain {
...
+            const swiftlyConfigRaw = await fs.readFile(
+                path.join(swiftlyHomeDir, "config.json"),
+                "utf-8"
+            );
+            const swiftlyConfig: unknown = JSON.parse(swiftlyConfigRaw);
+            if (
+                swiftlyConfig === undefined ||
+                swiftlyConfig === null ||
+                typeof swiftlyConfig !== "object" ||
+                !("installedToolchains" in swiftlyConfig)
```
plemarquand, line 214:
```suggestion
                !(swiftlyConfig instanceof Object) ||
                !("installedToolchains" in swiftlyConfig)
```

adam-fowler:
Haven't looked at code in any detail as I'm doing this on my phone. I assume you have included a valid state where there is no toolchain. 

Instead of checking do we have a toolchain everywhere why don't you create two stages to the extension initialisation one for the toolchain and one for the rest. At initialisation if you have no toolchain you are forced through the select toolchain UI and the rest of the extension doesn't initialize until a toolchain is either downloaded or selected.

src/toolchain/toolchain.ts
matthewbastien: Yup. The other platforms don't have common directories where toolchains get installed. I've added a short circuit at the beginning of this function to avoid checking on non-darwin platforms.

src/toolchain/toolchain.ts
matthewbastien: done.

src/ui/ToolchainSelection.ts
matthewbastien: done

src/ui/ToolchainSelection.ts
matthewbastien: done

src/ui/ToolchainSelection.ts
matthewbastien: done

matthewbastien:
> Haven't looked at code in any detail as I'm doing this on my phone. I assume you have included a valid state where there is no toolchain.
> 
> Instead of checking do we have a toolchain everywhere why don't you create two stages to the extension initialisation one for the toolchain and one for the rest. At initialisation if you have no toolchain you are forced through the select toolchain UI and the rest of the extension doesn't initialize until a toolchain is either downloaded or selected.

Fair point. I've made it so that the WorkspaceContext doesn't even get created if there's no available toolchain. This allows us to keep relatively the same behaviour as before this patch. However, the command registration now takes an optional WorkspaceContext and will display the toolchain error message if commands are invoked without a WorkspaceContext. I've also moved some of the functionality that doesn't require a toolchain outside of the WorkspaceContext (e.g. the reload extensions notification).

adam-fowler:
So far tested this on macOS (Although I didn't test with multiple Xcode's installed) and works quite well. One minor issue, if you have a local toolchain and then select a new toolchain but set it globally you are still using the original local toolchain. I think you should probably clear the local setting here, or add another dialog to ask do you want to clear it.

I'll do a code review later, but I did see a lot of the code is still dealing with possibility of an undefined toolchain, folder context etc, which should be unnecessary now given your description of the changes you made.

src/commands.ts
```diff
@@ -893,6 +939,10 @@ export function register(ctx: WorkspaceContext) {
         vscode.commands.registerCommand("swift.selectXcodeDeveloperDir", () =>
```
adam-fowler, line 939: I guess this and the referenced function can be removed now, as select toolchain does the equivalent and more

src/extension.ts
```diff
@@ -42,11 +46,57 @@ export interface Api {
...
+                outputChannel.log(error);
+                contextKeys.createNewProjectAvailable = true;
+                return undefined;
+            });
+        const workspaceContext = toolchain
+            ? await WorkspaceContext.create(outputChannel, toolchain)
+            : undefined;
+        context.subscriptions.push(...commands.register(workspaceContext));
+
+        if (!toolchain && vscode.workspace.workspaceFolders) {
```
adam-fowler, line 74: Why `&& vscode.workspace.workspaceFolders` someone can create a single swift file and still expect swift support.

adam-fowler (changes requested):
Here is an initial review of code. There is a load of code related to having an optional WorkspaceContext that can be reverted. 

We need to make sure we get the UI flow correct for toolchain installation if no toolchain is currently installed.

src/ui/ToolchainSelection.ts
```diff
@@ -0,0 +1,200 @@
...
+import { showReloadExtensionNotification } from "./ReloadExtension";
+import configuration from "../configuration";
+import { WorkspaceContext } from "../WorkspaceContext";
+import { SwiftToolchain } from "../toolchain/toolchain";
+
+/**
+ * Open the installation page on Swift.org
+ */
+export async function downloadToolchain() {
+    if (await vscode.env.openExternal(vscode.Uri.parse("https://www.swift.org/install/"))) {
```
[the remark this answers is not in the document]
matthewbastien: I'm thinking this will be used more by developers who are new to Swift. The install page is a little more friendly to that, imo.

src/ui/ToolchainSelection.ts
```diff
@@ -0,0 +1,200 @@
...
+        const selected = await showReloadExtensionNotification(
```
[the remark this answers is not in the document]
matthewbastien: We can't auto-detect when a new toolchain is installed unless we do some polling. The user needs to be notified that they need to reload once they're done installing.

src/WorkspaceContext.ts
```diff
@@ -37,23 +36,22 @@ import { TestCoverageReportProvider } from "./coverage/TestCoverageReport";
...
  * Context for whole workspace. Holds array of contexts for each workspace folder
  * and the ExtensionContext
  */
 export class WorkspaceContext implements vscode.Disposable {
     public folders: FolderContext[] = [];
-    public currentFolder: FolderContext | null | undefined;
-    public currentDocument: vscode.Uri | null;
-    public outputChannel: SwiftOutputChannel;
+    public currentFolder: FolderContext | undefined;
+    public currentDocument: vscode.Uri | undefined;
```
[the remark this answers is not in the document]
matthewbastien: I can't even remember why I changed it tbh... I can revert if you want.

src/commands.ts
matthewbastien: Actually, I believe the DEVELOPER_DIR setting is still useful. There are cases where you would want to set it even though you've selected a public toolchain, for example.

src/configuration.ts
```diff
@@ -159,6 +159,58 @@ const configuration = {
     get path(): string {
         return vscode.workspace.getConfiguration("swift").get<string>("path", "");
     },
+    async setPath(
```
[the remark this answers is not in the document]
matthewbastien: Done.

src/configuration.ts
```diff
@@ -159,6 +159,58 @@ const configuration = {
...
+            ];
+            if (vscode.workspace.workspaceFolders) {
+                items.push({
+                    label: "Workspace",
+                    detail: "Add to Workspace configuration",
+                    target: vscode.ConfigurationTarget.Workspace,
+                });
+                if (vscode.workspace.workspaceFolders.length > 1) {
+                    items.push({
+                        label: "workspace folders",
```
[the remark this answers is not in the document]
matthewbastien: I'm leaning towards keeping this lower case. Microsoft seems to prefer lower case labels for their separators.

src/extension.ts
```diff
@@ -42,11 +46,57 @@ export interface Api {
...
+            .catch(error => {
+                outputChannel.log("Failed to discover Swift toolchain");
+                outputChannel.log(error);
+                contextKeys.createNewProjectAvailable = true;
+                return undefined;
+            });
+        const workspaceContext = toolchain
+            ? await WorkspaceContext.create(outputChannel, toolchain)
+            : undefined;
+        context.subscriptions.push(...commands.register(workspaceContext));
```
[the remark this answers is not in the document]
matthewbastien: The problem is that the commands will still appear in the command palette and running them yields an awful error message that just says something along the lines of "could not find the command". The way I have it set up now is that running the commands with an invalid toolchain pops up the toolchain error notification which they can use to resolve the issue.

src/extension.ts
```diff
@@ -42,11 +46,57 @@ export interface Api {
...
+                toolchain.logDiagnostics(outputChannel);
+                contextKeys.createNewProjectAvailable = toolchain.swiftVersion.isGreaterThanOrEqual(
+                    new Version(5, 8, 0)
+                );
+                return toolchain;
+            })
+            .catch(error => {
+                outputChannel.log("Failed to discover Swift toolchain");
+                outputChannel.log(error);
+                contextKeys.createNewProjectAvailable = true;
```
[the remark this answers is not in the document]
matthewbastien: See comment below about only registering certain commands. I'd like to leave the option for them to run the command and get a nice error message.

src/WorkspaceContext.ts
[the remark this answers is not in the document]
adam-fowler: yes please

src/WorkspaceContext.ts
```diff
@@ -414,14 +380,17 @@ export class WorkspaceContext implements vscode.Disposable {
...
-    ): Promise<FolderContext> {
+    ): Promise<FolderContext | undefined> {
         // find context with root folder
         const index = this.folders.findIndex(context => context.folder.fsPath === folder.fsPath);
         if (index !== -1) {
             console.error(`Adding package folder ${folder} twice`);
             return this.folders[index];
         }
         const folderContext = await FolderContext.create(folder, workspaceFolder, this);
+        if (!folderContext) {
```
adam-fowler, line 391: This should be necessary anymore

src/WorkspaceContext.ts
```diff
@@ -476,8 +445,8 @@ export class WorkspaceContext implements vscode.Disposable {
 
     /** find LLDB version and setup path in CodeLLDB */
     async setLLDBVersion() {
-        // check we are using CodeLLDB
-        if (DebugAdapter.adapterName !== "lldb") {
+        // check we are using CodeLLDB and have a valid toolchain
+        if (!this.toolchain || DebugAdapter.adapterName !== "lldb") {
```
adam-fowler, line 449: Don't need to check toolchain anymore

src/commands.ts
adam-fowler: Possibly but I was using it as a method to select the toolchain from an Xcode install. I think including this and `select toolchain...` is confusing to the user.

src/toolchain/toolchain.ts
```diff
@@ -242,12 +314,37 @@ export class SwiftToolchain {
...
     /**
      * Cannot use `getToolchainExecutable` to get the LLDB executable as LLDB
      * is not in macOS toolchain path
      */
-    public getLLDB(): string {
-        return path.join(this.swiftFolderPath, "lldb");
+    public async getLLDB(): Promise<string> {
+        let lldbPath = path.join(this.swiftFolderPath, "lldb");
+        if (!(await pathExists(lldbPath))) {
+            if (process.platform !== "darwin") {
```
adam-fowler, line 335: In what situation were you not finding LLDB?

src/ui/ToolchainSelection.ts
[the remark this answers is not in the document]
adam-fowler: Yeah it makes the flow a little more awkward, maybe the messages should be different for when someone selects download

adam-fowler (commented):
A few more comments, nothing major.

src/extension.ts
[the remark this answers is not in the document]
adam-fowler: The alternative is to use the `isActivated` context key to make the commands only available once the extension is fully activated with toolchain.

src/ui/ToolchainSelection.ts
```diff
@@ -0,0 +1,200 @@
...
+            });
+        }
+    }
+    return [
+        ...(xcodes.length > 0 ? [new SeparatorItem("Xcode"), ...xcodes] : []),
+        ...(toolchains.length > 0 ? [new SeparatorItem("toolchains"), ...toolchains] : []),
+        ...(swiftlyToolchains.length > 0
+            ? [new SeparatorItem("swiftly"), ...swiftlyToolchains]
+            : []),
+        new SeparatorItem("actions"),
```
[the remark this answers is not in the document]
matthewbastien: I've added an action that will bring the user to the Swiftly install page.

src/commands.ts
matthewbastien: Fair enough. I've removed it, but a sufficiently advanced user can still use the configuration to set DEVELOPER_DIR if they need it.

src/extension.ts
```diff
@@ -42,11 +46,57 @@ export interface Api {
...
+                    // FIXME: There is a bug stopping us from restarting SourceKit-LSP directly.
+                    // As long as it's fixed we won't need to reload on newer versions.
+                    showReloadExtensionNotification(
+                        "Changing the Swift SDK path requires the project be reloaded."
+                    );
+                }
+            })
+        );
+
+        if (!workspaceContext) {
```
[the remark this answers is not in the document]
matthewbastien: done

src/toolchain/toolchain.ts
adam-fowler: ok

src/extension.ts
```diff
@@ -42,11 +46,53 @@ export interface Api {
...
+                return undefined;
+            });
+
+        context.subscriptions.push(...commands.registerToolchainCommands(toolchain));
+        context.subscriptions.push(
+            vscode.workspace.onDidChangeConfiguration(event => {
+                // on toolchain config change, reload window
+                if (event.affectsConfiguration("swift.path")) {
+                    showReloadExtensionNotification(
+                        "Changing the Swift path requires the project be reloaded."
```
adam-fowler, line 76: You may not have a project loaded, you may have multiple projects loaded...
```suggestion
                        "Changing the Swift path requires Visual Studio Code be reloaded."
```

src/ui/ToolchainSelection.ts
```diff
@@ -0,0 +1,200 @@
...
+        {
+            type: "action",
+            label: "$(folder-opened) Select toolchain directory...",
+            detail: "Select a folder on your machine where the Swift toolchain is installed",
+            run: selectToolchainFolder,
+        },
+    ];
+}
+
+/**
```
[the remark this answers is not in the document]
adam-fowler: Ok yeah I mixed those two up. Yeah the two function names are very close `selectToolchain` could become `showSelectToolchainQuickPick` so something similar

adam-fowler (commented):
Two minor nits, apart from that I think we are good

src/extension.ts
matthewbastien: done

src/ui/ToolchainSelection.ts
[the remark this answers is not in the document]
matthewbastien: done

matthewbastien:
> So far tested this on macOS (Although I didn't test with multiple Xcode's installed) and works quite well. One minor issue, if you have a local toolchain and then select a new toolchain but set it globally you are still using the original local toolchain. I think you should probably clear the local setting here, or add another dialog to ask do you want to clear it.
> 
> I'll do a code review later, but I did see a lot of the code is still dealing with possibility of an undefined toolchain, folder context etc, which should be unnecessary now given your description of the changes you made.

So, I did some more investigation into this. It appears that settings for individual workspace folders are actually ignored for `swift.path`. I've removed the option to configure workspace folder settings and added a warning about workspace settings taking precedence over user settings.

@adam-fowler could you take one more look at the code and see if we're good?

src/configuration.ts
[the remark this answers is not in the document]
matthewbastien: My latest patch removes this code.

adam-fowler (approved):
This all looks good

I have thought of one issue though, perhaps we can resolve this in another PR

If I select an Xcode install that isn't the Xcode install returned by `xcode-select -p` (default). The calls to `SwiftToolchain.getLLDB` and `SwiftToolchain.getSDKPath` return the wrong paths ie from the default install. 

Should we set DEVELOPER_DIR environment var if we are setting the path to an Xcode install to ensure we get the correct LLDB and SDK path

PR #818, opened 2024-05-22
Allow hiding the create swift project button from the welcome page

matthewbastien:
Adding our own button to the welcome page by default is a tragedy of the commons issue. I'd like to keep the button available, but allow it to be hidden with a configuration option.

### DIFF
--- a/package.json
+++ b/package.json
@@ -293,7 +293,7 @@
           },
           "swift.showCreateSwiftProjectInWelcomePage": {
             "type": "boolean",
-            "default": false,
+            "default": true,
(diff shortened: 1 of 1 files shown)


plemarquand (approved):

matthewbastien:
> Other language extensions add buttons to the welcome page eg Java. Can we keep the option but default it to true.

Yeah, we can keep it true for now. I'm just concerned that the more extensions that add to this the more bloated it's going to become. At least people can disable it if they don't want it this way.

PR #821, opened 2024-05-28
Deduplicate swift diagnostics

### LINKED ISSUE #653: De-duplicate errors from sourcekitd and swiftc in the Problems view
In the problems pane, it would be nice if we could de-duplicate errors reported by `sourcekitd` and `swiftc`. My suggestion for this would be: If `sourcekitd` reports an error at the same location with the same message as `swiftc`, hide it.

award999:
Fixes #653

* Keeping "$swiftc" problem matcher around in case it is still being used
* Add "diagnosticsCollection" setting to give more control on which diagnostics are kept around and provided to the user
* Setup diagnostics from task output and hooking into sourcekit-lsp
* Handle merging based on "diagnosticsCollection" setting

This does not include serialized diagnostics which will involve changes in swift itself

### DIFF
--- a/src/DiagnosticsManager.ts
+++ b/src/DiagnosticsManager.ts
@@ -1,0 +1,276 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VSCode Swift open source project
+//
+// Copyright (c) 2024 the VSCode Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of VSCode Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import * as vscode from "vscode";
+import stripAnsi = require("strip-ansi");
+import configuration from "./configuration";
+import { SwiftExecution } from "./tasks/SwiftExecution";
+import { WorkspaceContext } from "./WorkspaceContext";
+
+interface ParsedDiagnostic {
+    uri: string;
+    diagnostic: vscode.Diagnostic;
+}
+
+type DiagnosticsMap = Map<string, vscode.Diagnostic[]>;
+
+/**
+ * Handles the collection and deduplication of diagnostics from
+ * various {@link vscode.Diagnostic.source | Diagnostic sources}.
+ *
+ * Listens for running {@link SwiftExecution} tasks and allows
+ * external clients to call {@link handleDiagnostics} to provide
+ * thier own diagnostics.
+ */
+export class DiagnosticsManager implements vscode.Disposable {
+    // Prior to Swift 6 "sourcekitd" was the source
+    static sourcekit: string[] = ["SourceKit", "sourcekitd"];
+    static swiftc: string[] = ["swiftc"];
+
+    private diagnosticCollection: vscode.DiagnosticCollection =
+        vscode.languages.createDiagnosticCollection("swift");
+
+    constructor(context: WorkspaceContext) {
+        this.onDidStartTaskDisposible = vscode.tasks.onDidStartTask(event => {
+            if (!this.includeSwiftcDiagnostics()) {
+                return;
+            }
+
+            // Will only try to provide diagnostics for `swift` tasks
+            const execution = event.execution.task.execution;
+            if (!(execution && execution instanceof SwiftExecution)) {
+                return;
+            }
+            const swiftExecution = execution as SwiftExecution;
+
+            // Clean up old "swiftc" diagnostics
+            this.removeBuildDiagnostics();
+
+            const provideDiagnostics: Promise<DiagnosticsMap> =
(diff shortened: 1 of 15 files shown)


award999:

award999:

src/DiagnosticsManager.ts
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    private mergeDiagnostics(
+        currentDiagnostics: vscode.Diagnostic[],
+        newDiagnostics: vscode.Diagnostic[],
+        precedence: string[]
+    ): vscode.Diagnostic[] {
+        const merged: vscode.Diagnostic[] = [];
+        const messages: string[] = [];
+        for (const diagnostic of newDiagnostics) {
```
adam-fowler, line 141: Is it not possible to do this just by merging straight  into `currentDiagnostics` instead 
- Creating a new list of unique build diagnostics
- Then creating a list from that of the ones that need to be merged into `currentDiagnostics`
- Then returning the union of that list with `currentDiagnostics`

src/DiagnosticsManager.ts
```diff
@@ -0,0 +1,281 @@
...
+            if (messages.includes(diagnostic.message)) {
+                continue;
+            }
+            messages.push(diagnostic.message);
+            // See if a duplicate diagnostic exists
+            const currentDiagnostic = currentDiagnostics.find(
+                d =>
+                    d.range.start.isEqual(diagnostic.range.start) &&
+                    d.message === diagnostic.message
+            );
```
adam-fowler, line 153: As I mentioned in Slack. Swift 5.10 compiler produces diagnostics with the first character capitalised and SourceKit-LSP produces diagnostics with everything lowercase. Thus nothing is merged

adam-fowler (commented):
In general pretty happy with this, Just need to be sure we are going to merge diagnostics correctly.

One issue that might come up as well, the compiler sometimes generates info diagnostics (related to an error) you don't get from SKLSP. Will be a bit weird if we remove an error because SKLSP says it is fixed, but the related info diagnostic is not removed because it came from the build.

src/DiagnosticsManager.ts
```diff
@@ -0,0 +1,291 @@
...
+
+    constructor(context: WorkspaceContext) {
+        this.onDidStartTaskDisposible = vscode.tasks.onDidStartTask(event => {
+            // Will only try to provide diagnostics for `swift` tasks
+            const execution = event.execution.task.execution;
+            if (!(execution && execution instanceof SwiftExecution)) {
+                return;
+            }
+            // Clean up old "swiftc" diagnostics
+            this.removeBuildDiagnostics();
```
adam-fowler, line 52: Prior to this PR build errors were cleared after a build completed and before the new build's errors were added to the problems view. This meant you had the list of build errors from the previous build available while you were in the process of doing a new build.

adam-fowler:
> One issue that might come up as well, the compiler sometimes generates info diagnostics (related to an error) you don't get from SKLSP. Will be a bit weird if we remove an error because SKLSP says it is fixed, but the related info diagnostic is not removed because it came from the build.

An easy way to replicate this is have a protocol and a struct that conforms to this protocol. Add a new requirement to the protocol and compile your project. You get an info diagnostic `Protocol requires ...`. When you add the requirement to your struct this info diagnostic is not removed.

The info diagnostics are already problematic as they are not associated with the error that created them. Ideally we would be able to create a hierarchy of diagnostics where the info diagnostic was a child of the actual error diagnostic instead of a flat list. Don't think that is possible in VSCode though

award999:
> > One issue that might come up as well, the compiler sometimes generates info diagnostics (related to an error) you don't get from SKLSP. Will be a bit weird if we remove an error because SKLSP says it is fixed, but the related info diagnostic is not removed because it came from the build.
> 
> An easy way to replicate this is have a protocol and a struct that conforms to this protocol. Add a new requirement to the protocol and compile your project. You get an info diagnostic `Protocol requires ...`. When you add the requirement to your struct this info diagnostic is not removed.
> 
> The info diagnostics are already problematic as they are not associated with the error that created them. Ideally we would be able to create a hierarchy of diagnostics where the info diagnostic was a child of the actual error diagnostic instead of a flat list. Don't think that is possible in VSCode though
> 
> EDIT: Just noticed `Diagnostic` has a `relatedInformation` field. Is this something we could use for informational diagnostics, if we can associate them with an error.

Ya I stumbled across yesterday that a "note" always is attached to the preceding warning or error. Will put that patch up soon, and ya will be adding it into the relatedInformation :)

src/DiagnosticsManager.ts
award999: done

src/DiagnosticsManager.ts
award999: Done

award999:

award999:

adam-fowler (approved):
Approved on the assumption we look at merging the `relatedInformation` arrays in a future PR

PR #830, opened 2024-05-30
Add Swift: Capture VSCode Swift Diagnostic Logs command

### LINKED ISSUE #842: Add command to the command palette to run `sourcekit-lsp diagnose`
Starting with Swift 6, the `sourcekit-lsp diagnose` terminal command generates a bundle that helps SourceKit-LSP developers diagnose issues. I think it would be great if we could make that more accessible in VS Code by adding a command to the command palette that runs `sourcekit-lsp diagnose --toolchain /path/to/toolchain` in the integrated terminal. In particular, this could ensure that you run the command with the toolchain that is selected in VS Code, which yields more actionable bugs.

As an added bonus, when sourcekitd crashes and SourceKit-LSP shows the `SourceKit-LSP: Restoring functionality: Please run 'sourcekit-lsp diagnose' to file an issue` status (you can reproduce this by running `killall SourceKitService` when having a project open), it would be great if there was a button to run `sourcekit-lsp diagnose`. I can add some indication to the work done progress that SourceKit-LSP sends for this.

plemarquand:
Adds a new command that users can use to help generate bug reports for the extension itself. The Capture VSCode Swift Diagnostic Logs command will create a new folder in a temporary directory that contains:

- Swift version and path information
- Extension logs
- Any diagnostics in the Problems pane

Users can copy the folder path or, if they're on macOS, open it in Finder. From there they can zip and attach these files to a GitHub issue. We'll want to update the New Issue Template with instructions on how to capture and attach these logs.

Fixes #842

### DIFF
--- a/package.json
+++ b/package.json
@@ -172,7 +172,7 @@
       },
       {
         "command": "swift.captureDiagnostics",
-        "title": "Capture VSCode Swift Diagnostic Logs",
+        "title": "Capture VS Code Swift Diagnostic Logs",
         "category": "Swift"
       },
       {
--- a/package.json
+++ b/package.json
@@ -172,7 +172,7 @@
       },
       {
         "command": "swift.captureDiagnostics",
-        "title": "Capture VS Code Swift Diagnostic Logs",
+        "title": "Capture VS Code Swift Diagnostic Bundle",
         "category": "Swift"
       },
       {
--- a/package.json
+++ b/package.json
@@ -171,6 +171,11 @@
         "category": "Swift"
       },
       {
+        "command": "swift.captureDiagnostics",
+        "title": "Capture VSCode Swift Diagnostic Logs",
+        "category": "Swift"
+      },
+      {
         "command": "swift.clearDiagnosticsCollection",
         "title": "Clear Diagnostics Collection",
         "category": "Swift"
--- a/src/commands/captureDiagnostics.ts
+++ b/src/commands/captureDiagnostics.ts
@@ -166,9 +166,9 @@ async function sourcekitDiagnose(ctx: WorkspaceContext, dir: string) {
             location: vscode.ProgressLocation.Notification,
         },
         async progress => {
-            progress.report({ message: "Generating Diagnostic Bundle..." });
+            progress.report({ message: "Diagnosing SourceKit-LSP..." });
             const writableStream = progressUpdatingWritable(percent =>
-                progress.report({ message: `Generating Diagnostic Bundle: ${percent}%` })
+                progress.report({ message: `Diagnosing SourceKit-LSP: ${percent}%` })
             );
 
             await execFileStreamOutput(
--- a/src/commands/captureDiagnostics.ts
+++ b/src/commands/captureDiagnostics.ts
@@ -26,17 +26,20 @@ export async function captureDiagnostics(ctx: WorkspaceContext) {
         `vscode-diagnostics-${formatDateString(new Date())}`
     );
 
-    const versionOutputChannel = new SwiftOutputChannel();
-    ctx.toolchain.logDiagnostics(versionOutputChannel);
+    const environmentOutputChannel = new SwiftOutputChannel();
+    ctx.toolchain.logDiagnostics(environmentOutputChannel);
+    environmentOutputChannel.log(
+        JSON.stringify(vscode.workspace.getConfiguration("swift"), null, 2)
+    );
 
     const logs = ctx.outputChannel.logs.join("\n");
-    const versionLogs = versionOutputChannel.logs.join("\n");
+    const environmentLogs = environmentOutputChannel.logs.join("\n");
     const diagnosticLogs = buildDiagnostics();
 
     try {
         await fs.mkdir(diagnosticsDir);
         await fs.writeFile(path.join(diagnosticsDir, "logs.txt"), logs);
-        await fs.writeFile(path.join(diagnosticsDir, "version.txt"), versionLogs);
+        await fs.writeFile(path.join(diagnosticsDir, "environment.txt"), environmentLogs);
         await fs.writeFile(path.join(diagnosticsDir, "diagnostics.txt"), diagnosticLogs);
 
         ctx.outputChannel.log(`Saved diagnostics to ${diagnosticsDir}`);
@@ -113,11 +116,9 @@ function severityToString(severity: vscode.DiagnosticSeverity): string {
     }
 }
 
-function padZero(num: number, length: number = 2): string {
-    return num.toString().padStart(length, "0");
-}
-
 function formatDateString(date: Date): string {
+    const padZero = (num: number, length: number = 2) => num.toString().padStart(length, "0");
+
     const year = date.getFullYear();
     const month = padZero(date.getMonth() + 1);
     const day = padZero(date.getDate());
--- a/src/commands/captureDiagnostics.ts
+++ b/src/commands/captureDiagnostics.ts
@@ -26,21 +26,12 @@ export async function captureDiagnostics(ctx: WorkspaceContext) {
         `vscode-diagnostics-${formatDateString(new Date())}`
     );
 
-    const environmentOutputChannel = new SwiftOutputChannel();
-    ctx.toolchain.logDiagnostics(environmentOutputChannel);
-    environmentOutputChannel.log(
-        JSON.stringify(vscode.workspace.getConfiguration("swift"), null, 2)
-    );
-
-    const logs = ctx.outputChannel.logs.join("\n");
-    const environmentLogs = environmentOutputChannel.logs.join("\n");
-    const diagnosticLogs = buildDiagnostics();
-
     try {
         await fs.mkdir(diagnosticsDir);
-        await fs.writeFile(path.join(diagnosticsDir, "logs.txt"), logs);
-        await fs.writeFile(path.join(diagnosticsDir, "environment.txt"), environmentLogs);
-        await fs.writeFile(path.join(diagnosticsDir, "diagnostics.txt"), diagnosticLogs);
+        await writeLogFile(diagnosticsDir, "logs.txt", extensionLogs(ctx));
+        await writeLogFile(diagnosticsDir, "environment.txt", environmentLogs(ctx));
+        await writeLogFile(diagnosticsDir, "sourcekit-lsp.txt", sourceKitLogs(ctx));
+        await writeLogFile(diagnosticsDir, "diagnostics.txt", diagnosticLogs());
 
         ctx.outputChannel.log(`Saved diagnostics to ${diagnosticsDir}`);
 
@@ -68,6 +59,44 @@ export async function captureDiagnostics(ctx: WorkspaceContext) {
     }
 }
 
+async function writeLogFile(dir: string, name: string, logs: string) {
+    if (logs.length === 0) {
+        return;
+    }
+    await fs.writeFile(path.join(dir, name), logs);
+}
+
+function extensionLogs(ctx: WorkspaceContext): string {
+    return ctx.outputChannel.logs.join("\n");
+}
+
+function environmentLogs(ctx: WorkspaceContext): string {
+    const environmentOutputChannel = new SwiftOutputChannel("Swift");
+    ctx.toolchain.logDiagnostics(environmentOutputChannel);
+    environmentOutputChannel.log("Extension Settings:");
+    environmentOutputChannel.log(
+        JSON.stringify(vscode.workspace.getConfiguration("swift"), null, 2)
+    );
+    return environmentOutputChannel.logs.join("\n");
+}
+
+function diagnosticLogs(): string {
+    const diagnosticToString = (diagnostic: vscode.Diagnostic) => {
+        return `${severityToString(diagnostic.severity)} - ${diagnostic.message} [Ln ${diagnostic.range.start.line}, Col ${diagnostic.range.start.character}]`;
+    };
+
+    return vscode.languages
+        .getDiagnostics()
+        .map(
+            ([uri, diagnostics]) => `${uri}\n\t${diagnostics.map(diagnosticToString).join("\n\t")}`
+        )
+        .join("\n");
+}
+
+function sourceKitLogs(ctx: WorkspaceContext) {
+    return (ctx.languageClientManager.languageClientOutputChannel?.logs ?? []).join("\n");
+}
+
 function showDirectoryCommand(dir: string): string {
     switch (process.platform) {
         case "win32":
@@ -88,19 +117,6 @@ function showCommandType(): string {
         default:
             return "File Manager";
     }
-}
-
-function buildDiagnostics(): string {
-    const diagnosticToString = (diagnostic: vscode.Diagnostic) => {
-        return `${severityToString(diagnostic.severity)} - ${diagnostic.message} [Ln ${diagnostic.range.start.line}, Col ${diagnostic.range.start.character}]`;
-    };
-
-    return vscode.languages
-        .getDiagnostics()
-        .map(
-            ([uri, diagnostics]) => `${uri}\n\t${diagnostics.map(diagnosticToString).join("\n\t")}`
-        )
-        .join("\n");
 }
 
 function severityToString(severity: vscode.DiagnosticSeverity): string {
--- a/src/commands/captureDiagnostics.ts
+++ b/src/commands/captureDiagnostics.ts
@@ -1,12 +1,12 @@
 //===----------------------------------------------------------------------===//
 //
-// This source file is part of the VSCode Swift open source project
+// This source file is part of the VS Code Swift open source project
 //
-// Copyright (c) 2021-2024 the VSCode Swift project authors
+// Copyright (c) 2021-2024 the VS Code Swift project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
-// See CONTRIBUTORS.txt for the list of VSCode Swift project authors
+// See CONTRIBUTORS.txt for the list of VS Code Swift project authors
 //
 // SPDX-License-Identifier: Apache-2.0
 //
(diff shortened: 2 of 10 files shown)


src/commands/captureDiagnostics.ts
```diff
@@ -0,0 +1,111 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VSCode Swift open source project
+//
+// Copyright (c) 2021-2024 the VSCode Swift project authors
```
award999, line 5: just 2024?

src/commands/captureDiagnostics.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    try {
+        await fs.mkdir(diagnosticsDir);
+        await fs.writeFile(path.join(diagnosticsDir, "logs.txt"), logs);
+        await fs.writeFile(path.join(diagnosticsDir, "version.txt"), versionLogs);
+        await fs.writeFile(path.join(diagnosticsDir, "diagnostics.txt"), diagnosticLogs);
+
+        ctx.outputChannel.log(`Saved diagnostics to ${diagnosticsDir}`);
+
+        const showInFinderButton = "Show In Finder";
```
award999, line 44: question: What about Windows / Linux?

src/commands/captureDiagnostics.ts
```diff
@@ -0,0 +1,111 @@
...
+        } else if (result === showInFinderButton) {
+            exec(`open ${diagnosticsDir}`, error => {
+                if (error) {
+                    vscode.window.showErrorMessage(`Failed to open Finder: ${error.message}`);
+                    return;
+                }
+            });
+        }
+    } catch (error) {
+        vscode.window.showErrorMessage(`Unable to captrure diagnostics logs: ${error}`);
```
award999, line 65: spelling: capture

src/ui/SwiftOutputChannel.ts
```diff
@@ -65,6 +66,7 @@ export class SwiftOutputChannel {
 
     private sendLog(line: string) {
         this.channel.append(line);
+        this.logStore.append(line);
```
award999, line 69: suggestion: what I've seen before is the file log have all diagnostics, including the verbose ones. i.e. the "diagnostics" setting would only affect output channel verbosity

adam-fowler:
I have virtually never needed the diagnostics in the problems page when someone is submitting an issue. Things that are useful are swift log (which already includes swift version and paths), sourcekit-lsp log, user settings.

src/ui/SwiftOutputChannel.ts
plemarquand: There isn't really a verbose/non-verbose concept in the extension at the moment; everything just goes to the output channel

plemarquand:
@adam-fowler I've added sourcekit-lsp logs and user settings to the diagnostics. Sourcekit-lsp logs will only be captured if the user has the `trace` level turned on in their settings.

package.json
```diff
@@ -164,6 +164,11 @@
         "command": "swift.attachDebugger",
         "title": "Attach to Process...",
         "category": "Swift"
+      },
+      {
+        "command": "swift.captureDiagnostics",
+        "title": "Capture VSCode Swift Diagnostic Logs",
```
ahoppen, line 170: Drive-by nitpick: As as far as I know the recommended spelling is `VS Code`, https://code.visualstudio.com/brand

```suggestion
        "title": "Capture VS Code Swift Diagnostic Logs",
```

plemarquand: Good call, I did a separate pass and made all usages consistent in #846

src/commands/captureDiagnostics.ts
```diff
@@ -0,0 +1,209 @@
...
+    );
+
+    try {
+        await fs.mkdir(diagnosticsDir);
+        await writeLogFile(diagnosticsDir, "logs.txt", extensionLogs(ctx));
+        await writeLogFile(diagnosticsDir, "environment.txt", environmentLogs(ctx));
+
+        if (ctx.swiftVersion.isGreaterThanOrEqual(new Version(6, 0, 0))) {
+            // sourcekit-lsp diagnose command is only available in 6.0 and higher.
+            // await writeLogFile(diagnosticsDir, "sourcekit-lsp.txt", );
```
ahoppen, line 40: Leftover code?

src/commands/captureDiagnostics.ts
```diff
@@ -0,0 +1,209 @@
...
+}
+
+async function sourcekitDiagnose(ctx: WorkspaceContext, dir: string) {
+    const sourcekitDiagnosticDir = path.join(dir, "sourcekit-lsp");
+    await fs.mkdir(sourcekitDiagnosticDir);
+
+    const toolchainSourceKitLSP = ctx.toolchain.getToolchainExecutable("sourcekit-lsp");
+    const lspConfig = configuration.lsp;
+    const serverPathConfig = lspConfig.serverPath;
+    const serverPath = serverPathConfig.length > 0 ? serverPathConfig : toolchainSourceKitLSP;
```
ahoppen, line 119: Could you also pass the toolchain that the user has selected via `swift.path` to `sourcekit-lsp diagnose`? That way we will diagnose using the same toolchain that the user is building with if they have a custom SourceKit-LSP server set.

src/commands/captureDiagnostics.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    await vscode.window.withProgress(
+        {
+            location: vscode.ProgressLocation.Notification,
+        },
+        async progress => {
+            progress.report({ message: "Capturing Diagnostics..." });
```
ahoppen, line 126: Should this be something like _Generating SourceKit-LSP Diagnose Bundle_? _Capturing Diagnostics_ sounds to much like the the diagnostics aka live issues to me.

ahoppen (commented):
Great stuff! I expect this will really help us diagnose issues more quickly. Do you want to add instructions to run the diagnose command to the GitHub issue template as well? We should probably only do that once a version of the extension with the diagnose command has been released.

src/commands/captureDiagnostics.ts
plemarquand: I want to keep this message a bit more ambiguous, because if you're not using a Swift 6 toolchain then `sourcekit-lsp diagnose` isn't available. We fall back to recording the captured logs from the sourcekit-lsp output channel, but this isn't really the same thing as the full diagnostics bundle.

src/commands/captureDiagnostics.ts
```diff
@@ -0,0 +1,209 @@
...
+
+        if (ctx.swiftVersion.isGreaterThanOrEqual(new Version(6, 0, 0))) {
+            // sourcekit-lsp diagnose command is only available in 6.0 and higher.
+            // await writeLogFile(diagnosticsDir, "sourcekit-lsp.txt", );
+            await sourcekitDiagnose(ctx, diagnosticsDir);
+        } else {
+            await writeLogFile(diagnosticsDir, "sourcekit-lsp.txt", sourceKitLogs(ctx));
+        }
+
+        await writeLogFile(diagnosticsDir, "diagnostics.txt", diagnosticLogs());
```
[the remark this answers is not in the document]
plemarquand: Yes, I agree this is redundant if a `sourcekit-lsp diagnose` bundle is included, but if it isn't available it could maybe be useful.

src/ui/SwiftOutputChannel.ts
```diff
@@ -65,4 +102,47 @@ export class SwiftOutputChannel {
...
+        return [...this._logs];
+    }
+
+    appendLine(log: string) {
+        // It can be costly to calculate the actual memory size of a string in Node so just
+        // use the total number of characters in the logs as a huristic for total size.
+        const logSize = log.length;
+
+        while (this.currentLogLength + logSize > this.maxSizeCharacters && this.logs.length > 0) {
+            const oldestLog = this.logs.shift();
```
adam-fowler, line 127: I don't know how `shift` works internally but this is going to be either re-allocating large buffers of memory, or not freeing up any memory (because the array is equivalent to an array slice). Perhaps we should use a circular buffer. It might be harder to keep a consistent maximum memory usage but wouldn't require creating new arrays of strings repeatedly.

src/ui/SwiftOutputChannel.ts
```diff
@@ -15,11 +15,56 @@
...
 
-    constructor() {
-        this.channel = vscode.window.createOutputChannel("Swift");
+    public name: string;
+
+    /**
+     * Creates a vscode.OutputChannel that allows for later retrival of logs.
+     * @param name
+     */
+    constructor(name: string, logToConsole: boolean = true) {
```
adam-fowler, line 29:
```suggestion
    constructor(
        public name: string, 
        private logToConsole: boolean = true
    ) {
```
And remove them from above

src/commands/captureDiagnostics.ts
```diff
@@ -0,0 +1,209 @@
...
+
+        ctx.outputChannel.log(`Saved diagnostics to ${diagnosticsDir}`);
```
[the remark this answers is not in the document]
plemarquand: I planned on adding this to the `Saved diagnostic logs` information message but it looks like VS Code forbids newlines in these messages, making it a bit awkward. https://github.com/Microsoft/vscode/blob/8616dbae8bc2abf7972a45449b0fb6b2b2d0f429/src/vs/workbench/common/notifications.ts#L412-L413

src/commands/captureDiagnostics.ts
ahoppen: What do you think about _Diagnosing SourceKit-LSP…_? That should resolve the ambiguity with diagnostics aka. live issues.

src/commands/captureDiagnostics.ts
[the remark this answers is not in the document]
ahoppen: OK, I think we’re talking about different things here. What does `diagnostics.txt` include?

src/commands/captureDiagnostics.ts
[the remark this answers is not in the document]
ahoppen: Ah, I forgot that notifications can’t contain newlines. I hit this as well for indexing progress…

What do you think about adding the following sentence (or something of the like): 
> This folder contains information that helps developers of Swift for VS Code diagnose issues and may contain information about your source code and VS Code setup.

ahoppen (commented):
One more question. Do you think the following from https://github.com/swift-server/vscode-swift/issues/842 would be possible to implement?

> As an added bonus, when sourcekitd crashes and SourceKit-LSP shows the `SourceKit-LSP: Restoring functionality: Please run 'sourcekit-lsp diagnose' to file an issue` status (you can reproduce this by running `killall SourceKitService` when having a project open), it would be great if there was a button to run `sourcekit-lsp diagnose`. I can add some indication to the work done progress that SourceKit-LSP sends for this.

src/commands/captureDiagnostics.ts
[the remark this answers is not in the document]
plemarquand: Its essentially the contents of the Problems panel

src/commands/captureDiagnostics.ts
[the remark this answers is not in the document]
plemarquand: I reworked the initial dialog to be a modal instead of an information panel. It gives a full accounting of what is captured, and offers two options, Full or Minimal diagnostic bundle. If the user chooses Minimal then sourcekit-lsp logs are omitted. I put the call to action in this dialog.

Screenshot 2024-06-11 at 12 09 58 PM

src/ui/SwiftOutputChannel.ts
plemarquand: I refactored the RollingLog to use a ring buffer style algorithm.

src/commands/captureDiagnostics.ts
```diff
@@ -0,0 +1,209 @@
...
+    }
+    await fs.writeFile(path.join(dir, name), logs);
+}
+
+function extensionLogs(ctx: WorkspaceContext): string {
+    return ctx.outputChannel.logs.join("\n");
+}
+
+function environmentLogs(ctx: WorkspaceContext): string {
+    const environmentOutputChannel = new SwiftOutputChannel("Swift", false);
```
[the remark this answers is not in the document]
plemarquand: I've refactored this to no longer need the output channel.

plemarquand:
@ahoppen it looks like it should be pretty straightforward to watch the work done progress events. However, there is only `title` and `message` on the work progress messages, meaning I'd have to check if `title` === `'SourceKit-LSP: Restoring functionality'`, which seems brittle. I guess we can use it if we can guarantee that string wont change

src/commands/captureDiagnostics.ts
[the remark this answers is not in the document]
ahoppen: OK, then it shouldn’t really be redundant if `sourcekit-lsp diagnose` is included because it will also include eg. build issues. That’s what threw me off for a short bit. 

That’s why I was suggesting to rename it to `source-code-diagnostics.txt` to avoid any ambiguities with the diagnose command added by this PR and the `sourcekit-lsp diagnose` bundle.

src/commands/captureDiagnostics.ts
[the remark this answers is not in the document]
ahoppen: But even the minimal diagnostic bundle could include sensitive information, right? Error messages from the Problems pane alone might contain eg. strings from the user’s code base and those might be sensitive.

ahoppen:
I can totally add an ID for this to the work done progress that we can check for. Just wanted to verify that it’s possible on the VS Code side before I extend LSP. Let’s do this in a follow-up PR.

src/commands/captureDiagnostics.ts
```diff
@@ -0,0 +1,281 @@
...
+
+This information contains:
+- Extension logs
+- Versions of Swift installed on your system
+- Crash logs from SourceKit
+- Log messages emitted by SourceKit
+- If possible, a minimized project that caused SourceKit to crash
+- If possible, a minimized project that caused the Swift compiler to crash
+
+Please attach this bundle to GitHub issues.`,
```
ahoppen, line 103: Should we have a call for action here to file an issue? I think this would be especially valuable when a user captures a diagnostic bundle after sourcekitd has crashed. Something like

```suggestion
Please file an issue with a description of the problem you are seeing and this diagnose bundle at https://github.com/swiftlang/vscode-swift.`,
```

Or use `apple/sourcekit-lsp`, I don’t have strong opinions here. It’ s pretty easy to route between the two.

src/commands/captureDiagnostics.ts
```diff
@@ -0,0 +1,281 @@
...
+    if (
+        ctx.swiftVersion.isGreaterThanOrEqual(new Version(6, 0, 0)) ||
+        vscode.workspace.getConfiguration("sourcekit-lsp").get<string>("trace.server", "off") !==
+            "off"
+    ) {
+        const fullButton = allowMinimalCapture ? "Capture Full Diagnostics" : "Capture Diagnostics";
+        const minimalButton = "Capture Minimal Diagnostics";
+        const buttons = allowMinimalCapture ? [fullButton, minimalButton] : [fullButton];
+        const fullCaptureResult = await vscode.window.showInformationMessage(
+            `A Diagnostic Bundle collects information that helps the developers of the VS Code Swift extension diagnose and fix issues.
```
ahoppen, line 93: I think according to https://code.visualstudio.com/brand we should refer to it as *Swift for VS Code*

```suggestion
            `A Diagnostic Bundle collects information that helps the developers of the Swift for VS Code extension diagnose and fix issues.
```

src/commands/captureDiagnostics.ts
```diff
@@ -0,0 +1,283 @@
...
+
+        const diagnosticsDir = path.join(
+            tmpdir(),
+            `vscode-diagnostics-${formatDateString(new Date())}`
+        );
+
+        await fs.mkdir(diagnosticsDir);
+        await writeLogFile(diagnosticsDir, "extension-logs.txt", extensionLogs(ctx));
+        await writeLogFile(diagnosticsDir, "settings.txt", settingsLogs(ctx));
+        await writeLogFile(diagnosticsDir, "source-code-diagnostics.txt", diagnosticLogs());
```
ahoppen, line 46: I think we should only add the source code diagnostics in the full capture mode because it may contain information about the user’s source code.

ahoppen (approved):

src/commands/captureDiagnostics.ts
plemarquand: Done

plemarquand:

award999 (approved):

src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -534,6 +540,20 @@ export class LanguageClientManager {
...
+                        const result = await next(token, params);
+                        const title = (params as langclient.WorkDoneProgressBegin).title;
+                        if (title === "SourceKit-LSP: Restoring functionality") {
+                            restoringLSPStarted = true;
+                        } else if (params.kind === "end" && restoringLSPStarted) {
+                            restoringLSPStarted = false;
+                            await promptForDiagnostics(this.workspaceContext);
+                        }
+                        return result;
+                    };
```
ahoppen, line 555: With https://github.com/apple/sourcekit-lsp/pull/1493 you can check whether the token has the `sourcekitd-crashed` prefix. 

Also, two things I just noticed:
1. I think this triggers the prompt for diagnostics whenever *any* work done progress ends after the `Restoring functionality` has started
2. We can show the prompt for diagnostics when sourcekitd crashes. We don’t need to wait for the service to be restored.

ahoppen: I think there would be value in resetting the prompt status after 1 hour or maybe 1 day. It’s quite likely that if sourcekitd crashes again after a long time period, that it’s a different kind of issue that would be worth reporting.

plemarquand: Sounds good, I've capped it to show the dialog at most once an hour.

PR #831, opened 2024-05-31
Respect withKnownIssue in swift-testing tests

plemarquand:
If an expectation fails within a withKnownIssue closure the issue is marked with `isKnown` and should be excluded from the list of expectation failures. If all expectations in a test are marked `isKnown` then the test should be shown as skipped.

rdar://129024156


adam-fowler (approved):
Looks good, shame we can't add a message to indicate why something is skipped

PR #837, opened 2024-06-04
Fix flakey pipe removal

plemarquand:
On posix platforms if a named pipe hasn't been written to then occasionally calls to fs.lstat (which I believe fs.rm uses internally) will  hang indefinitely.

If a test run failed due to a compiler error in the build then the pipe never be written to. By repeatedly trying to perform a test run with code that fails to compile eventually this hang would appear.

Work around the issue by writing to the pipe if compilation fails, ensuring the connection opens/closes and the pipe can be removed


matthewbastien (approved):
Looks alright to me. Seems hacky, but I suppose there isn't a better solution. Is there a related issue that's been raised against NodeJS?

adam-fowler:
The other solution would be to write a dummy line to the named pipe.

plemarquand:
@adam-fowler You're right, I think this works better and is more close to the source of the issue, which is that if the other side of a pipe never opens then the hang is possible.

PR #841, opened 2024-06-05
Dispose LSP's output channel on restart

### LINKED ISSUE #839: Relaunching the LSP Server doesn't clean up the output sections
**Describe the bug**
I am working on a complex Swift package, and I sometimes need to relaunch the LSP in order to see changes I make propagate. Doing so, however, seems to add an entry to the Output section in VSCode every time.

**To Reproduce**
1. Go to the Output tab in the bottom panel in VSCode
2. Open the dropdown of output sections
3. Observe a single entry of "Swift" and a single entry of "SourceKit Language Server"
4. Press Command-Shift-P and find the "Swift: Restart LSP Server" option and select it
5. Go back to the dropdown and observe a second entry of "SourceKit Language Server"
6. Repeat step 4 and observe a new entry each time.

**Expected behavior**
Restarting the LSP server should not add an entry each time, or, at the very least, these entries should be distinguished from each other.

**Environment**
 - OS: macOS 14.5
 - Swift version (output of `swift --version`)
 ```
 swift-driver version: 1.90.11.1 Apple Swift version 5.10 (swiftlang-5.10.0.13 clang-1500.3.9.4)
Target: arm64-apple-macosx14.0
```
 - Visual Studio Code version: 1.89.1
 - vscode-swift version: v1.9.0

plemarquand:
When restarting SourceKit-LSP its output channel was not disposed, which caused a new entry to be added to the output channel list without removing the old one.

Fixes #839

### DIFF
--- a/src/sourcekit-lsp/LanguageClientManager.ts
+++ b/src/sourcekit-lsp/LanguageClientManager.ts
@@ -351,6 +351,9 @@ export class LanguageClientManager {
                 .stop()
                 .then(async () => {
                     await this.setupLanguageClient(workspaceFolder?.uri);
+
+                    // Now that the client has been replaced, dispose the old client's output channel.
+                    client.outputChannel.dispose();
                 })
                 .catch(async reason => {
                     // error message matches code here https://github.com/microsoft/vscode-languageserver-node/blob/2041784436fed53f4e77267a49396bca22a7aacf/client/src/common/client.ts#L1409C1-L1409C54


plemarquand:

adam-fowler:
This is kind of dodgy, in that we are assuming the VSCode language client never calls the `dispose` on the outputChannel. Looking through the language client code it should be calling it, so I am a little confused as to why this call is needed.
https://github.com/microsoft/vscode-languageserver-node/blob/0226f6cd925d279dda2b777d25c5b69518ac0035/client/src/common/client.ts#L1414

plemarquand:
After some digging I understand what is happening here, and it does seem to be a bug in vscode-languageclient. The flow is:

- The initial `client` is created, client's output channel ("OC1") is created.
- User runs "Restart LSP Server"
- `client.stop` is called, ["OC1" is disposed, client._outputChannel is set to `undefined`](https://github.com/microsoft/vscode-languageserver-node/blob/0226f6cd925d279dda2b777d25c5b69518ac0035/client/src/common/client.ts#L1451)
- Cleanup continues, and the spawned LSP process is terminated
- This [finally](https://github.com/microsoft/vscode-languageserver-node/blob/1ce790dd341a38da1fbf902cddc16ce456149de0/client/src/node/main.ts#L505) block executes, listening for the LSP process to exit.
- LSP exits with code `0`, _which is then logged_ [here](https://github.com/microsoft/vscode-languageserver-node/blob/1ce790dd341a38da1fbf902cddc16ce456149de0/client/src/node/main.ts#L509).
- The `outputChannel` getter, seeing that `this._outputChannel` is undefined due to previous disposal, [creates a new output channel](https://github.com/microsoft/vscode-languageserver-node/blob/1ce790dd341a38da1fbf902cddc16ce456149de0/client/src/common/client.ts#L656) ("OC2") that replaces the one that was disposed. 
  - This output channel is what I'm disposing in this patch. Without this code, OC2 never gets disposed.
- Finally, we create a new client, which creates a new output channel "OC3".

This leaves us with one more output channel than we started with.

adam-fowler:
> After some digging I understand what is happening here, and it does seem to be a bug in vscode-languageclient. The flow is:
> 
Do you want to add an issue for the vscode language client?

adam-fowler (approved):

plemarquand:
Sounds good, I'll create that issue once I have received internal approval and link it back to here.

plemarquand:
Created https://github.com/microsoft/vscode-languageserver-node/issues/1496 to track the root cause in vscode-language-server-node.

PR #845, opened 2024-06-05
Fix flakey pipe removal for debug runs

plemarquand:
#837 fixed this for regular runs, but more work was required for debug runs


award999 (approved):

src/TestExplorer/TestRunner.ts
```diff
@@ -785,11 +787,13 @@ export class TestRunner {
                                             });
                                         subscriptions.push(terminateSession);
                                     } else {
+                                        await this.swiftTestOutputParser.close();
```
adam-fowler, line 790: You could add a dispose function that called `this.swiftTestOutputParser.close();` and add it to the subscription list instead of adding the call everywhere. Might be slightly cleaner

adam-fowler:
I guess this works, but it seems more sensible to move the test building outside of the `TemporaryFolder.withNamedTemporaryFile` closure. Or at least create the fifo once we have built the tests and only if we have swift-testing tests.

plemarquand:
I agree this isn’t ideal. I’m planning on revisiting this next week with the approach you outlined.

PR #853, opened 2024-06-06
Provide error when debugger isn't started

plemarquand:
When debugging an XCTest that fails to build the debugger promise rejects but did not provide an error. The undefined error is caught and passed to `getErrorDescription` which fails to read `stderr` off of it. This caused another error to be thrown and caught at the top level, showing a dialog.

Throw an error string instead of undefined, and handle the undefined error case in getErrorDescription.

### DIFF
--- a/src/TestExplorer/TestRunner.ts
+++ b/src/TestExplorer/TestRunner.ts
@@ -19,7 +19,7 @@ import * as os from "os";
 import * as asyncfs from "fs/promises";
 import { FolderContext } from "../FolderContext";
 import { execFile, getErrorDescription } from "../utilities/utilities";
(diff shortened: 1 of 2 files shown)


award999 (approved):

award999:
> If a build fails and I click on debug anyway in the dialog, it doesn't run the debug anymore

So it works for me IFF it succeeded building before. So if delete .build and try to debug the test with build errors, even if I click Debug Anyway, the LLDB extension bails on the debug session. But if the build succeeds once and then introduce the error, then Debug Anyway runs

plemarquand:
I've broken out the build step when debugging and if there is a failure we exit before the debugger is started, sidestepping this issue.

PR #858, opened 2024-06-07
Suppress swift-testing output when debugging XCTests

plemarquand:
When debugging an XCTest in a project that contains both XCTests and swift-testing tests the `xctest` utility runs both test executables. This results in an empty swift testing test run output being printed to console.

Prevent `xctest` from running the swift-testing binary by setting the `SWIFT_TESTING_ENABLED` environment variable to 0.

### DIFF
--- a/src/debugger/buildConfig.ts
+++ b/src/debugger/buildConfig.ts
@@ -98,19 +98,14 @@ export class TestingDebugConfigurationFactory {
     /* eslint-disable no-case-declarations */
     private buildLinuxConfig(): vscode.DebugConfiguration | null {
         if (this.testKind === TestKind.debug && this.testLibrary === TestLibrary.xctest) {
-            const { folder } = getFolderAndNameSuffix(this.ctx, this.expandEnvVariables);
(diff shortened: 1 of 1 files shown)


src/debugger/buildConfig.ts
```diff
@@ -191,28 +186,15 @@ export class TestingDebugConfigurationFactory {
...
                             ...this.baseConfig,
                             program: path.join(xcTestPath, "xctest"),
-                            args: this.addXCTestExutableTestsToArgs([
-                                path.join(
-                                    buildDirectory,
-                                    "debug",
-                                    this.ctx.swiftPackage.name + "PackageTests.xctest"
-                                ),
-                            ]),
+                            args: this.addXCTestExutableTestsToArgs([this.xcTestOutputPath]),
```
adam-fowler, line 193: "addXCTestExutableTestsToArgs" [sic]

adam-fowler (approved):
I minor typo inherited from another PR

src/debugger/buildConfig.ts
plemarquand: Fixed in #866

award999:
Verified

PR #859, opened 2024-06-07
Update language file extensions

award999:
VSCode docs suggest .<extension> is expected. Also adding .swift as it seems like VSCode uses this list to decide which language extensions to recommend in extensions view

Adding "Testing" category


adam-fowler (approved):

plemarquand:
Verified

Issue #750, opened 2024-04-12
Mark compiler issues as stale if a file is edited


ahoppen:
When compilation of a file produces errors and those errors, these errors are displayed in the source editor. However, even when the source file has been modified, the issue will linger around until I build again. I’m not sure if it’s possible at all, but I think it would be great if build errors could be marked as stale (eg. gray instead of red) when the source file has been edited. That way, it’s easy to tell which errors are live errors from `sourcekit-lsp` that still need to be addressed and which ones are from a previous build.

adam-fowler:
I have no control over how the problems view is rendered.

matthewbastien:
I can investigate this one to figure out the feasibility.

adam-fowler:
See #759

award999:
I'd love to run with this issue

adam-fowler:
@award999 #759 gives a possible route to solving this.

award999:
Not working

PR #860, opened 2024-06-07
Remove stale swiftc diagnostics

award999:
Remove stale swiftc diagnostics

Issue: #750

If the user modifies source and SourceKit diagnostics differ, they may
want to cleanup all the old swiftc ones. This will be turned on by
default. The swiftc must match with a former SourceKit one, and once
SourceKit returns a new list of diagnostics, if that matched diagnostic
is no longer there, assume it is fix and remove the matching swiftc error

Tie this to a setting so the user can turn off.

Keep track of all current diagnostics regardless of source to allow for
this cleanup, and has the added benefit that we can update the list
when the diagnosticsCollection setting changes.

### DIFF
--- a/src/DiagnosticsManager.ts
+++ b/src/DiagnosticsManager.ts
@@ -48,14 +48,6 @@ export class DiagnosticsManager implements vscode.Disposable {
     constructor(context: WorkspaceContext) {
         this.onDidChangeConfigurationDisposible = vscode.workspace.onDidChangeConfiguration(e => {
             if (e.affectsConfiguration("swift.diagnosticsCollection")) {
-                // if (!this.includeSwiftcDiagnostics()) {
-                //     // Clean up "swiftc" diagnostics
-                //     this.removeSwiftcDiagnostics();
-                // }
-                // if (!this.includeSourceKitDiagnostics()) {
-                //     // Clean up SourceKit diagnostics
-                //     this.removeSourceKitDiagnostics();
-                // }
                 this.diagnosticCollection.clear();
                 this.allDiagnostics.forEach((_, uri) =>
                     this.updateDiagnosticsCollection(vscode.Uri.file(uri))
@@ -112,8 +104,7 @@ export class DiagnosticsManager implements vscode.Disposable {
         // of Swift as to whether the first letter is capitalized or not,
         // so we'll always display messages capitalized to user and this
         // also will allow comparing messages when merging
-        newDiagnostics.forEach(this.capitalizeMessage);
-        // const mergedDiagnostics = this.diagnosticCollection.get(uri)?.slice() || [];
+        newDiagnostics = newDiagnostics.map(this.capitalizeMessage);
         const allDiagnostics = this.allDiagnostics.get(uri.fsPath)?.slice() || [];
         // Remove the old set of diagnostics from this source
         const removedDiagnostics = this.removeDiagnostics(allDiagnostics, d =>
@@ -139,8 +130,8 @@ export class DiagnosticsManager implements vscode.Disposable {
 
     private updateDiagnosticsCollection(uri: vscode.Uri): void {
         const diagnostics = this.allDiagnostics.get(uri.fsPath) ?? [];
-        const swiftcDiagnostics = this.getDiagnostics(diagnostics, DiagnosticsManager.swiftc);
-        const sourceKitDiagnostics = this.getDiagnostics(diagnostics, DiagnosticsManager.sourcekit);
+        const swiftcDiagnostics = diagnostics.filter(d => this.isSwiftc(d));
+        const sourceKitDiagnostics = diagnostics.filter(d => this.isSourceKit(d));
         const mergedDiagnostics: vscode.Diagnostic[] = [];
         switch (configuration.diagnosticsCollection) {
             case "keepSourceKit":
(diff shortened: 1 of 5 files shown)


adam-fowler:
Is there any situation where someone would want to keep the stale diagnostics?

adam-fowler (approved):
I am concerned we are adding too many configuration settings. We now have 10 possible diagnostic lists. In what situation would someone want to not remove stale swifts diagnostics? Otherwise everything looks good

award999:
> Is there any situation where someone would want to keep the stale diagnostics?



> I am concerned we are adding too many configuration settings. We now have 10 possible diagnostic lists. In what situation would someone want to not remove stale swifts diagnostics? Otherwise everything looks good

You're right, really just using it as a "kill switch" but I cannot think of a user case. Will address that

src/DiagnosticsManager.ts
```diff
@@ -40,18 +43,23 @@ export class DiagnosticsManager implements vscode.Disposable {
...
-                    this.removeSourceKitDiagnostics();
-                }
+                // if (!this.includeSwiftcDiagnostics()) {
+                //     // Clean up "swiftc" diagnostics
+                //     this.removeSwiftcDiagnostics();
+                // }
+                // if (!this.includeSourceKitDiagnostics()) {
+                //     // Clean up SourceKit diagnostics
+                //     this.removeSourceKitDiagnostics();
+                // }
```
matthewbastien, line 58: nit: should just delete this instead of commenting it out

src/DiagnosticsManager.ts
```diff
@@ -40,18 +43,23 @@ export class DiagnosticsManager implements vscode.Disposable {
...
             if (e.affectsConfiguration("swift.diagnosticsCollection")) {
-                if (!this.includeSwiftcDiagnostics()) {
-                    // Clean up "swiftc" diagnostics
-                    this.removeSwiftcDiagnostics();
-                }
-                if (!this.includeSourceKitDiagnostics()) {
-                    // Clean up SourceKit diagnostics
-                    this.removeSourceKitDiagnostics();
-                }
+                // if (!this.includeSwiftcDiagnostics()) {
```
plemarquand, line 51: Leftover code?

src/DiagnosticsManager.ts
```diff
@@ -92,108 +100,161 @@ export class DiagnosticsManager implements vscode.Disposable {
...
         }
-        this.diagnosticCollection.set(uri, newDiagnostics);
+        this.diagnosticCollection.set(uri, mergedDiagnostics);
     }
 
     private mergeDiagnostics(
-        combinedDiagnostics: vscode.Diagnostic[],
-        incomingDiagnostics: vscode.Diagnostic[],
+        mergedDiagnostics: vscode.Diagnostic[],
+        // removedDiagnostics: vscode.Diagnostic[],
```
plemarquand, line 178: nit: remove

src/DiagnosticsManager.ts
```diff
@@ -92,108 +100,161 @@ export class DiagnosticsManager implements vscode.Disposable {
...
+        newDiagnostics: vscode.Diagnostic[]
+    ): void {
+        const isFromSourceKit = !!DiagnosticsManager.sourcekit.find(s => sources.includes(s));
         // Is a descrepency between SourceKit-LSP and older versions
         // of Swift as to whether the first letter is capitalized or not,
         // so we'll always display messages capitalized to user and this
         // also will allow comparing messages when merging
-        diagnostics.forEach(this.capitalizeMessage);
-        const newDiagnostics = this.diagnosticCollection.get(uri)?.slice() || [];
+        newDiagnostics.forEach(this.capitalizeMessage);
```
plemarquand, line 115: nit: Swift has taught me to prefer the functional style where possible: `newDiagnostics = newDiagnostics.map(this.capitalizeMessage)` where `capitalizeMessage` returns a copy of the object with the message field updated.

src/DiagnosticsManager.ts
```diff
@@ -92,108 +100,161 @@ export class DiagnosticsManager implements vscode.Disposable {
...
-        });
+        }
+        return removed;
     }
 
-    private removeDiagnostics(diagnostics: vscode.Diagnostic[], sources: string[]): void {
+    private getDiagnostics(
+        diagnostics: vscode.Diagnostic[],
+        sources: string[]
+    ): vscode.Diagnostic[] {
```
plemarquand, line 241: This function modifies its input as well as returns output. Would be more clear if it returned an object of `{ removed: Diagnostic[], diagnostics: Diagnostic[] }`

src/DiagnosticsManager.ts
award999: done

src/DiagnosticsManager.ts
award999: done

src/DiagnosticsManager.ts
award999: Ya really not needed, I should have just .filter(...) the array

src/DiagnosticsManager.ts
award999: done

plemarquand (approved):

tristanlabelle:
Hi @award999 ! I'm just catching up on this work and excited for the improvements. I thought VSCode did not have enough extensibility hooks to implement it. Do you know if this will work with non-SPM builds? We use CMake and sourcekit-lsp, and we use the `$swiftc` problem matcher on a CMake invocation.

award999:
@tristanlabelle this applies to `swift` tasks. The `$swiftc` problem matcher can still be used with any tasks you define, but no vscode does not allow for cleaning up diagnostics produced from a problem matcher, i.e. you need to run the task again

PR #861, opened 2024-06-07
Disable swiftly toolchain discovery on macOS and Windows

matthewbastien:
Swiftly is only available on Linux. Short circuit the toolchain discovery if the platform is macOS or Windows as it will find nothing of interest.


award999 (approved):

adam-fowler (approved):

award999:
Verified

PR #869, opened 2024-06-09
Do not append diagnostics style when debugging XCTests

plemarquand:
`xctest` doesn't recognize it and fails to run.


adam-fowler (approved):
I assume this is only adding build args to the calls to `swift test` now

plemarquand:
@adam-fowler correct

award999:
Verified with 205c628e9007f5ee26c16352ee27f3883916e4b2

PR #870, opened 2024-06-09
Don't match similarly named suites

plemarquand:
Add a trailing slash to suite test runs so that running a suite like `TestTarget.MySuite` doesn't also run `TestTarget.MySuite2`.

Test runs now register whenever a test starts/completes/records issue that isn't in the expected list of tests to be run. This is used by tests to verify no unknown tests ran.


award999 (approved):

award999:
Verified with 205c628

PR #874, opened 2024-06-10
Only add trailing suite slash to XCTests when not debugging

plemarquand:
The xctest executable accepts exact matches, not a regex.


award999 (approved):
Pulled patch, looks good to me

adam-fowler (approved):

PR #875, opened 2024-06-10
Show diff for failed XCTAssertEqual and XCTAssertIdentical calls

### LINKED ISSUE #654: Show diff output for failed `XCTAssertEqual` calls
When an `XCTAssertEqual` fails, it would be great if we could see a diff of the actual vs expected output. That would make spotting the actual difference a lot easier.

plemarquand:
If a test fails with XCTAssertEqual or XCTAssertIdentical, parse the result to provide a diff by setting the expected and actual values on TestMessage.

If the string representation of the class used for XCTAssertIdentical is identical, the diff would be identical, so don't show the diff view in that case.

Fixes #654

### DIFF
--- a/src/TestExplorer/TestParsers/TestRunState.ts
+++ b/src/TestExplorer/TestParsers/TestRunState.ts
@@ -46,7 +46,8 @@ export interface ITestRunState {
         index: number,
         message: string | vscode.MarkdownString,
         isKnown: boolean,
-        location?: vscode.Location
+        location?: vscode.Location,
+        diff?: TestIssueDiff
     ): void;
 
     // set test index to have been skipped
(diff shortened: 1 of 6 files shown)


plemarquand:
Screenshot 2024-06-10 at 3 56 10 PM

adam-fowler (commented):
I never knew you could add actual and expected values to a test message. Looks really good. I guess you need to do this for swift-testing as well now.

Originally I was going to ask why didn't you extract the diff from the message at the point of TestMessage construction but then realised the diff text format will be different based on which test framework you are using.

plemarquand:
@adam-fowler We'll want to add this for Swift Testing but need to wait for it to be included in the [JSON interface](https://github.com/apple/swift-testing/blob/main/Documentation/ABI/JSON.md).

adam-fowler:
> @adam-fowler We'll want to add this for Swift Testing but need to wait for it to be included in the [JSON interface](https://github.com/apple/swift-testing/blob/main/Documentation/ABI/JSON.md).

Ok we can treat that separately. I'm happy for this to be merged.

adam-fowler (approved):

PR #877, opened 2024-06-11
Rename `swift-latest` to `Latest Installed Toolchain` in the toolchain selection dialog

### LINKED ISSUE #865: Don't include swift-latest symbolic link in available toolchains
When you install a swift snapshot toolchain on macOS it setups up a `swift-latest.xctoolchain` symbolic link to point to it. We shouldn't include this symbolic link in the list of toolchains available when displaying the `Select toolchain...` menu

matthewbastien:
Issue: #865 

Make it clear that `swift-latest` actually points to the latest toolchain that the user has manually installed on their machine. Additionally, check to make sure that the Swift binary actually exists inside each found toolchain folder.


award999:

src/ui/ToolchainSelection.ts
```diff
@@ -150,13 +150,19 @@ async function getQuickPickItems(
...
-            type: "toolchain",
-            label: path.basename(toolchainPath, ".xctoolchain"),
-            detail: toolchainPath,
-            toolchainPath: path.join(toolchainPath, "usr"),
-            swiftFolderPath: path.join(toolchainPath, "usr", "bin"),
-        }));
+        .map<SwiftToolchainItem>(toolchainPath => {
+            let label = path.basename(toolchainPath, ".xctoolchain");
+            if (label === "swift-latest") {
+                label = "Latest Installed Toolchain";
```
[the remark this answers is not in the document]
matthewbastien: It points to the most recently installed toolchain.

src/ui/ToolchainSelection.ts
```diff
@@ -150,13 +150,19 @@ async function getQuickPickItems(
...
+            }
+            return {
```
adam-fowler, line 158: When someone selects the latest toolchain, do they expect to be selecting the latest toolchain installed at that point, or the latest installed toolchain ad infinitum. ie If they install another toolchain later on it automatically gets used.
 
Should we point to the linked toolchain or the symbolic link?

matthewbastien: That's a good question. My intention with this option was the latter: to always point at the most recently installed toolchain via the symbolic link. Otherwise, why have the option in the first place when you can just select the toolchain below it? However, the current wording is potentially confusing.

How does `Always Use Most Recently Installed Toolchain` sound? Seems a bit wordy to me, but it does get the point across.

adam-fowler: Yeah that is a bit wordy, maybe we can keep as is, but display a info window about it always being the most recently installed toolchain.

award999 (approved):
LGTM

src/ui/ToolchainSelection.ts
matthewbastien: I've added a notification that pops up after the toolchain has been selected. Let me know if you want to change the wording.

award999 (approved):

PR #883, opened 2024-06-11
Allow writing to different sourcekit output channels

### LINKED ISSUE #833: Write log messages from SourceKit-LSP’s index log to a separate output view
https://github.com/apple/sourcekit-lsp/pull/1398 adds an LSP extension to set a `logName` in the `window/logMessage` notifications. It would be great if the VS Code extension could write log messages that have `logName` set to an output view with that name. This would ensure that the index log does not get interleaved with other logs from SourceKit-LSP such as stderr messages or the message trace if verbose logging is enabled in the Swift extension.

award999:
Issue: #833

IFF sourcekit logMessage notification includes a logName property, we'll
use a different output channel to not polute the langauge server's main
log. For a known list of logs we'll set them up on startup so we can
control if the log level and/or timestamp is included in the log message.
For the current known indexing log, for example, all messages will be
"Info" so we'll just add timestamp. If new logs are introduced, we'll
show log level and timestamp by default

### DIFF
--- a/src/sourcekit-lsp/LanguageClientManager.ts
+++ b/src/sourcekit-lsp/LanguageClientManager.ts
@@ -24,11 +24,19 @@ import { FolderContext } from "../FolderContext";
 import { LanguageClient } from "vscode-languageclient/node";
 import { ArgumentFilter, BuildFlags } from "../toolchain/BuildFlags";
 import { DiagnosticsManager } from "../DiagnosticsManager";
+import { LSPLogger, LSPOutputChannel } from "./LSPOutputChannel";
+
+interface SourceKitLogMessageParams extends langclient.LogMessageParams {
+    logName?: string;
+}
 
 /** Manages the creation and destruction of Language clients as we move between
  * workspace folders
  */
 export class LanguageClientManager {
+    // known log names
+    static indexingLogName = "SourceKit-LSP: Indexing";
+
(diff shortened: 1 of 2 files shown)


award999:

src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -527,6 +533,35 @@ export class LanguageClientManager {
             this.workspaceContext.outputChannel.log(`SourceKit-LSP setup`);
         }
 
+        client.onNotification(langclient.LogMessageNotification.type, params => {
+            const sourceKitParams = params as SourceKitLogMessageParams;
+            if (sourceKitParams.logName) {
+                const outputChannel =
```
adam-fowler, line 539: Can we create the log channels in advance. We should know what logs SourceKit-LSP is going to want to open.

award999: After a chat will Alex, he asked if we could show timestamp but not log level for the Indexing log because it will always be info. So, for known logs like that, I'm letting set that configuration on startup. My idea behind allowing unknown logs to be created is if sourcekit-LSP changes, for development or as a user, they don't need to wait for an extension release to see the new log. Let me know if you still have concerns

plemarquand (approved):
LGTM

PR #886, opened 2024-06-12
Always register the swift-lldb debug adapter provider

### LINKED ISSUE #848: Need to reload vscode to start swift-lldb debug session
**Describe the bug**
Need to reload vscode to start swift-lldb debug session when enabling swift.debugger.useDebugAdapterFromToolchain setting

Screenshot 2024-06-05 at 4 05 03 PM

**To Reproduce**
1. Set swift.debugger.path to lldb-dap path
2. Enable the  swift.debugger.useDebugAdapterFromToolchain 
3. When prompted select to "Update" generated launch configurations

**Expected behavior**
Should be able to run this debug session with restarting vscode

**Environment**
 - OS: [e.g. Ubuntu 20.04] macOS 
 - Swift version (output of `swift --version`) 6.0 snapshot
 - Visual Studio Code version: 
 - vscode-swift version:

**Additional context**

award999:
Issue: #848

We can always register this and that way the user does not need to reload the window when use change to using lldb-dap. Now they can use the updated debug configurations right away

### DIFF
--- a/src/debugger/debugAdapter.ts
+++ b/src/debugger/debugAdapter.ts
@@ -58,7 +58,7 @@ export class DebugAdapter {
             : workspace.toolchain.getToolchainExecutable(debugAdapter);
         if (!(await this.doesFileExist(lldbDebugAdapterPath))) {
             if (!quiet) {
-                vscode.window.showInformationMessage(
+                vscode.window.showErrorMessage(
                     useCustom
                         ? `Cannot find ${debugAdapter} debug adapter specified in setting Swift.Debugger.Path.`
                         : `Cannot find ${debugAdapter} debug adapter in your Swift toolchain.`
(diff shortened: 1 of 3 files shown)


adam-fowler (commented):
Perhaps we shouldn't register it if it doesn't exist?

award999:
@adam-fowler the debug session does nothing if the path does not exist or cannot find in toolchain, so added a check to verify that the the debug adapter does exist so the user gets an error message instead of silently failing. I see that we are already listening to changes to `swift.debugger.useDebugAdapterFromToolchain` and `swift.debugger.path` settings, else where so seemed a little more complex than necessary to listen to this again to register the debug provider

adam-fowler (approved):

PR #888, opened 2024-06-12
Warn users about lack of symlink privileges on Windows

### LINKED ISSUE #805: Verify Developer mode is set on Windows before running extension
@compnerd, @tristanlabelle Is there an easy way to find this out?

matthewbastien:
Issue: #805 

Our available solutions are:

1. Check for Developer Mode being enabled in the registry using an existing npm package like `regedit`
2. Check for the `Create Symbolic Links` privilege. However, there are no good npm packages for this which means we would have to create a native module and compile it for all Windows architectures that we support.
3. Check that `fs.symlink()` works using a temporary file as the target.

I've opted for the 3rd approach since it is the simplest and it seems like all we really care about is that we can create symlinks on Windows. If `fs.symlink()` fails then the Swift extension will pop up a warning notification pointing the user at Microsoft's Developer Mode documentation. They can also choose to silence the warning in their user settings if they do not wish to enable Developer Mode or grant the `Create Symbolic Link` permission.


src/extension.ts
```diff
@@ -49,8 +50,30 @@ export interface Api {
...
-
         const outputChannel = new SwiftOutputChannel();
+
+        if (process.platform === "win32" && configuration.warnAboutSymlinkCreation) {
+            isSymlinkAllowed(outputChannel).then(async canCreateSymlink => {
+                if (canCreateSymlink) {
+                    return;
+                }
+                const selected = await vscode.window.showWarningMessage(
+                    "The Swift extension is unable to create symbolic links on your system and some features may not work correctly. Please either enable Developer Mode or allow symlink creation via Windows priviledges.",
```
award999, line 61:
```suggestion
                    "The Swift extension is unable to create symbolic links on your system and some features may not work correctly. Please either enable Developer Mode or allow symlink creation via Windows privileges.",
```

compnerd:
Could the `fs.symlink` fail for other reasons other than lack of privileges? Developer Mode being disabled is not a great heuristic either - I have a group policy applied to grant myself the privilege and do not have developer mode enabled.

matthewbastien:
> Could the `fs.symlink` fail for other reasons other than lack of privileges?

It would also fail if the user didn't have permissions to write to the directory, but presumably it would fail at the creation of the temporary directory instead. The only other thing I can think of (and I'm just speculating here - I have not tested) is if the workspace were on a remote file system or something that didn't support symlinks at all. Perhaps we could write to a hidden file/folder in the workspace as a better check instead?

I did test this out with a Windows VM and it works as expected at least with my setup. I was able to set the privilege, restart Windows, and the symlink creation succeeded.

> Developer Mode being disabled is not a great heuristic either

Agreed. That's why I'm not too keen on that approach. The only thing it has going for it is ease of implementation since there are already npm modules out there for reading from the registry. Checking the privilege directly, unfortunately, requires us to build native modules which is just a pain for one small warning.

package.json
```diff
@@ -357,6 +357,13 @@
               "Show the Swift build status as a progress notification."
             ],
             "order": 17
+          },
+          "swift.warnAboutSymlinkCreation": {
+            "type": "boolean",
+            "default": true,
+            "description": "Controls whether or not the extension will warn about being unable to create symlinks. (Windows only)",
+            "scope": "application",
+            "order": 18
```
adam-fowler, line 366: Why is this an option? The extension will not work correctly if symlink creation is disabled? So we should warn about it regardless

matthewbastien: In my testing the extension mostly works even with symbolic links disabled. I'm able to build a simple application and use SourceKit-LSP for language features. The only thing I noticed is that the following warning is displayed by swift while building:
```
warning: unable to create symbolic link at C:\Users\matthew_bastien\Documents\swift-executable\.build\debug: Error Domain=NSCocoaErrorDomain Code=256 "(null)"
```

Of course, this means that debugging won't work, but you still get other features and can `swift run` from the command line. I can see a circumstance in which a user might not care (or be unable to modify the policy for whatever reason) and just want the warning to not show up every time they open a swift project.

src/extension.ts
```diff
@@ -49,8 +50,30 @@ export interface Api {
 export async function activate(context: vscode.ExtensionContext): Promise<Api | undefined> {
     try {
         console.debug("Activating Swift for Visual Studio Code...");
-
         const outputChannel = new SwiftOutputChannel();
+
+        if (process.platform === "win32" && configuration.warnAboutSymlinkCreation) {
```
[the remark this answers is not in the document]
matthewbastien: I've moved it into src/ui/win32.ts since it shows UI now.

src/extension.ts
```diff
@@ -49,8 +50,30 @@ export interface Api {
...
+            isSymlinkAllowed(outputChannel).then(async canCreateSymlink => {
```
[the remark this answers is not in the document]
matthewbastien: Shouldn't affect startup times at all. I purposefully didn't `await` the symlink check to have it happen in the background.

adam-fowler (approved):

adam-fowler:

Issue #887, opened 2024-06-12
Failed to setup CodeLLDB for debugging of Swift code


DeltriDev:
**Describe the bug**
When I try to run a swift program through vscode (so like, F5) it doesnt work. When I launch vs code, i get this popup message:
`Failed to setup CodeLLDB for debugging of Swift code. Debugging may produce unexpected results. Error: Failed to find LLDB in swift toolchain. (source:Swift)`
For now I just use `swift run` on the terminal to run prokects

**To Reproduce**
1. Install swift
2. Install the swift vscode plugin
3. this happens

**Expected behavior**
I should be able to just run the swift project i guess

**Environment**
 - OS: Windows 10
 - Swift version: 5.10
 - Visual Studio Code version: 1.90.0
 - vscode-swift version: v1.10.1

DeltriDev:
@matthewbastien does that mean it's fixed? How can I test it?

adam-fowler:
@DeltriDev you can either build the extension for yourself or wait for the next release, hopefully not too long. We have build instructions in the CONTRIBUTING.md guide

plemarquand:
I can't verify this because I only have an ARM64 Windows VM and the CodeLLDB extension doesn't support that architecture.

@DeltriDev I updated the CONTRIBUTING.md with a section on [building and installing the extension from the repository](https://github.com/swiftlang/vscode-swift/blob/main/CONTRIBUTING.md#building). Could you try it out to see if this issue is fixed for you?

DeltriDev:
I'll try at some point tomorrow, I'll have to reinstall swift because it caused me some problems and I had to uninstall it

DeltriDev:
@adam-fowler I cannot build so I'll wait for the next update to let you know if it works!

award999:
Verified b461f8f

PR #890, opened 2024-06-12
Use proper lldb path on Windows

matthewbastien:
Issue: #887 

On Windows, the LLDB executable is named `lldb.exe` and wasn't being located correctly.

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -345,10 +345,10 @@ export class SwiftToolchain {
      * is not in macOS toolchain path
      */
     public async getLLDB(): Promise<string> {
-        let lldbPath = path.join(this.swiftFolderPath, "lldb");
-        if (process.platform === "win32") {
-            lldbPath = path.join(this.swiftFolderPath, "lldb.exe");
-        }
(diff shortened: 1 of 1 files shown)


award999 (approved):

award999:
Is there existing tests we can add an extra case?

matthewbastien:
> Is there existing tests we can add an extra case?

Not until we have a Windows testing platform on Jenkins.

src/toolchain/toolchain.ts
```diff
@@ -346,6 +346,9 @@ export class SwiftToolchain {
      */
     public async getLLDB(): Promise<string> {
         let lldbPath = path.join(this.swiftFolderPath, "lldb");
```
compnerd, line 348: How about sinking this into an `else` condition after the `win32` check? Alternatively, you could write this as a ternary:

```javascript
  let lldbPath = path.join(this.swiftFolderPath, process.platform === 'win32' ? 'lldb.exe' : 'lldb')
```

matthewbastien: That works, too. Done!

adam-fowler (approved):

PR #897, opened 2024-06-14
The _testCase field is no longer provided for non paramtereized tests

plemarquand:
https://github.com/apple/swift-testing/pull/482 removes the `_testCase` field if the test isn't parameterized. The code that recorded issues relied on this being present. If the issue for a parameterized test case then just extract the test from the testID.


award999 (approved):
LGTM

Issue #902, opened 2024-06-14
Have separate debug consoles for xctest and swift-testing


award999:
Right now if you run debug on all tests the swift-testing tests run first then the xctests. Because the 2 debug sessions seem to have the same name, the swift-testing debug console gets cleared when the xctests start to run. In the past I've noticed if 2 debug sessions have different names, they both stick around and you can use the dropdown selector to switch between them.

Since the test run output is empty for debug sessions, it becomes more crucial to be able to see the output for both test runs, and so you can see any print statements for debugging

plemarquand:
This looks implementable, with the caveat that VS Code doesn't seem to have an API to remove an old session with a different name from the dropdown. 

Specifically, if we do a debugging run of all tests in a XCTest+Swift Testing project this creates two sessions with, for instance, the names "XCTest" and "Swift Testing". However if then we run a single XCTest the "XCTest" debug session is updated but the old "Swift Testing" session sticks around in the dropdown with the output from the last run.

award999:
Verified with b461f8f

PR #904, opened 2024-06-14
Give debugging sessions unique names

plemarquand:
If we're testing in both frameworks we're going to start more than one debugging session. If both build configurations have the same name LLDB will replace the output of the first one in the Debug Console with the output of the second one. This means the user loses the debug output from the first session.

If they each have a unique name the Debug Console gets a nice dropdown the user can switch between to see the output for both sessions.

This approach has the caveat where VS Code doesn't seem to have an API to remove an old session with a different name from the dropdown.

Specifically, if we do a debugging run of all tests in a XCTest+Swift Testing project this creates two sessions with, for instance, the names "XCTest" and "Swift Testing". However if then we run a single XCTest the "XCTest" debug session is updated but the old "Swift Testing" session sticks around in the dropdown with the output from the last run.

Issue: #902

### DIFF
--- a/src/TestExplorer/TestRunner.ts
+++ b/src/TestExplorer/TestRunner.ts
@@ -725,6 +725,13 @@ export class TestRunner {
 
                     swiftTestBuildConfig.testType = TestLibrary.swiftTesting;
                     swiftTestBuildConfig.preLaunchTask = null;
+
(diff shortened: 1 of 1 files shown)


plemarquand:

award999 (approved):

adam-fowler (approved):

Issue #901, opened 2024-06-14
Add new UI section to settings


adam-fowler:
Include 
 - Exclude Paths From Package Dependencies
 - Show Build Status
 - Show Create Swift Project In Welcome Page
 - Open After Create New Project

PR #913, opened 2024-06-19
Add UI Settings Group

plemarquand:
Move the following settings in to their own group. Their ID does not change so there is no need to deprecate them or make code changes.

- `swift.excludePathsFromPackageDependencies`
- `swift.showBuildStatus`
- `swift.showCreateSwiftProjectInWelcomePage`
- `swift.openAfterCreateNewProject`

Issue #901

### DIFF
--- a/package.json
+++ b/package.json
@@ -276,23 +276,11 @@
             ],
             "order": 9
           },
-          "swift.excludePathsFromPackageDependencies": {
-            "description": "A list of paths to exclude from the Package Dependencies view.",
-            "type": "array",
-            "items": {
-              "type": "string"
-            },
-            "default": [
-              ".git",
-              ".github"
-            ],
-            "order": 10
-          },
           "swift.backgroundCompilation": {
             "type": "boolean",
             "default": false,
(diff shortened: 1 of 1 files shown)


award999:
Debugger and SourceKit-LSP settings put something in the setting ID. So we could add `.userInterface` to these settings (ex. `swift.userInterface.showBuildStatus`) to follow that theme and would show that category in the Settings editor (ex. `Swift › User Interface: Show Build Status`). Now 3 of these settings are fairly new so not quite as concerned about changing them as not too many people should have set, but `swift.excludePathsFromPackageDependencies` has been around a little longer so not sure how concerned we should be

plemarquand:
Yeah I was hesitant to change the actual setting IDs as these settings have been released and so we'd need to maintain/migrate them. Does adding a segment to the ID have any benefit other than code clarity?

award999:
Shows up marginally nicer in the Setting editor, ex. `Swift › User Interface: Show Build Status`

adam-fowler:
This would mean we would have to deprecate the old settings and point people to the new ones. Unless we are making a significant gain I wouldn't bother changing them.

adam-fowler (approved):

PR #917, opened 2024-06-19
Fix src/*.ts source mapping when debugging extension tests

award999:
Otherwise need to set a breakpoint in .js file under out


plemarquand (approved):
🎉 🎉 🎉

PR #924, opened 2024-06-21
Update title of command to select toolchain (or Xcode directory)

aoberoi:
Just fixing an outdated part of the README that probably should have landed with #817.

I am not sure if the Select Toolchain command is only available under macOS. That might be an outdated concept since the previous command was about selecting Xcode versions, which obviously are only on macOS. Toolchains are available on other platforms. I'm happy to update this PR if someone wants to confirm this one way or another.


award999 (changes requested):
Thanks for the PR 👍 but this is not just a macOS concept anymore. @matthewbastien can you advise on this one?

aoberoi:
Request change integrated. Ready for re-review/merge 😄

adam-fowler (approved):

award999 (approved):

0xTim (approved):

matthewbastien (approved):

PR #929, opened 2024-06-24
Parameterized test results should inherit tags

plemarquand:
Parameterized test results should inherit the tags of their parent. This keeps them visible in the Test Explorer when tests are filtered by tag.


adam-fowler (approved):
LGTM

award999 (approved):

Issue #911, opened 2024-06-18
Confusing which Xcode I'm using with a nightly toolchain


dschaefer2:
Now that Xcode 16 Beta is out I have two Xcodes installed on my machine. I have also downloaded the nightly build of the 6.0 Swift toolchain. Sometimes I want to use it with Xcode 15, sometimes with Xcode 16 Beta. When I use the command to select a Swift toolchain it's confusing which Xcode it's actually using. It would be nice if I could select that separately from the toolchain selection. At the end of they day, they are different environment variables on Mac, DEVELOPER_DIR and TOOLCHAINS.

adam-fowler:
When you select an Xcode installation you are selecting the toolchain from that Xcode install. If you select a toolchain install it isn't using your Xcode install at all, outside of one thing, the xctest library. I'm not certain offhand which Xcode version of xctest it would be using. Maybe @matthewbastien who wrote the toolchain select can answer that. 

If you are using swift-testing selecting a toolchain install should mean you aren't using an Xcode install at all. In theory you shouldn't even need Xcode installed. I haven't tested that though. 

We are not using the DEVELOPER_DIR or TOOLCHAINS environment variables.

plemarquand:
Just for clarity, swift-testing is included in the Xcode 16 Beta. This means that if you are using Xcode and then switch to a toolchain install you may need to explicitly specify swift-testing as a dependency.

adam-fowler:
> Just for clarity, swift-testing is included in the Xcode 16 Beta. This means that if you are using Xcode and then switch to a toolchain install you may need to explicitly specify swift-testing as a dependency.

You'd need to do the same if your package is compiled on Linux. I am assuming this will not be necessary once Swift 6 has a proper release though.

dschaefer2:
> We are not using the DEVELOPER_DIR or TOOLCHAINS environment variables.

I wonder if that's what we should be doing on macOS. That would allow us to be consistent with the command line experience.

Though as it is, if you swift build from the terminal inside VSCode you actually get different settings right now.

adam-fowler:
> Though as it is, if you swift build from the terminal inside VSCode you actually get different settings right now.

Yeah not ideal

[...]

PR #938, opened 2024-06-26
Prompt to update `DEVELOPER_DIR` for macOS toolchain selection

matthewbastien:
This updates the toolchain selection workflow to also set the `DEVELOPER_DIR` environment variable on macOS. This will be done automatically if the user selected an Xcode toolchain. However, they will be prompted to select a developer directory if they select a public Swift toolchain and there is more than one Xcode application installed on their system.

Linux and Windows toolchain selection is unaffected by this change.

Issue: #911

### DIFF
--- a/src/ui/ToolchainSelection.ts
+++ b/src/ui/ToolchainSelection.ts
@@ -63,7 +63,7 @@ export async function selectToolchainFolder() {
     if (!selected || selected.length !== 1) {
         return;
     }
-    await setToolchainPath(selected[0].fsPath, "prompt");
+    await setToolchainPath(selected[0].fsPath, "public");
 }
 
 /**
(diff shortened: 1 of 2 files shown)


adam-fowler:
We need to make it clear why you need the DEVELOPER_DIR, as it is not clear to me what the swift downloaded toolchains use from Xcode. @dschaefer2 perhaps you can answer that

dschaefer2:
> We need to make it clear why you need the DEVELOPER_DIR, as it is not clear to me what the swift downloaded toolchains use from Xcode. @dschaefer2 perhaps you can answer that

The macOS SDK? Lots of good stuff in there :).

src/ui/ToolchainSelection.ts
```diff
@@ -232,69 +256,170 @@ async function getQuickPickItems(
...
+                            return 1;
+                        }
+                        // Otherwise sort by name
+                        return a.label.localeCompare(b.label);
+                    });
+            }
+        ),
+        {
+            title: "Select a developer directory",
+            placeHolder: "Pick an Xcode installation to use as the developer directory",
```
adam-fowler, line 333: Indicate that this is for the macOS SDK

adam-fowler (changes requested):
Looks good, just one minor text update to look into. We should make it obvious what setting the DEVELOPER_DIR means when choosing a installed swift toolchain

src/ui/ToolchainSelection.ts
adam-fowler:
```suggestion
            placeHolder: "Pick an Xcode installation to use as the developer directory and for the macOS SDK",
```

adam-fowler (approved):

matthewbastien:

PR #941, opened 2024-06-27
Don't prepend target id to test ids if it exists

### LINKED ISSUE #936: Issues with tests in classes with the same name but in different targets
**Describe the bug**
If I have two test targets and both of them have a XCTest test class with the same name in a file with the same name then the Test Explorer struggles to create the correct TestItem hierarchy for these two classes. It tends to place all the tests from both classes under one TestItem, instead of under the correct TestItem associated with their class

**Environment**
 - OS: macOS 14.2
 - Swift version: Apple Swift version 6.0-dev (LLVM de395d39a90ed7a, Swift 490cf64aee23f13)
 - vscode-swift version: 1.10.3

plemarquand:
With https://github.com/swiftlang/sourcekit-lsp/pull/1530, `sourcekit-lsp` will start prepending a test target's name to the test id. This helps disambiguate when there are two identically named suites in two different test targets.

To maintain the current behaviour with older sourcekit-lsp versions only prefix the test target if it isn't already there.

- Depends on https://github.com/swiftlang/sourcekit-lsp/pull/1530
- Fixes #936

### DIFF
--- a/src/TestExplorer/LSPTestDiscovery.ts
+++ b/src/TestExplorer/LSPTestDiscovery.ts
@@ -148,7 +148,12 @@ export class LSPTestDiscovery {
             .getTargets(TargetType.test)
             .find(target => swiftPackage.getTarget(location.uri.fsPath) === target);
 
-        const id = target !== undefined ? `${target.c99name}.${item.id}` : item.id;
+        // If we're using an older sourcekit-lsp it doesn't prepend the target name
+        // to the test item id.
+        const id =
+            target !== undefined && !item.id.startsWith(`${target.c99name}.`)
+                ? `${target.c99name}.${item.id}`
+                : item.id;
         return item.style === "XCTest" ? id.replace(/\(\)$/, "") : id;
     }
 }


award999 (approved):

src/TestExplorer/LSPTestDiscovery.ts
```diff
@@ -148,7 +148,12 @@ export class LSPTestDiscovery {
             .getTargets(TargetType.test)
             .find(target => swiftPackage.getTarget(location.uri.fsPath) === target);
 
-        const id = target !== undefined ? `${target.c99name}.${item.id}` : item.id;
+        // If we're using an older sourcekit-lsp it doesn't prepend the target name
+        // to the test item id.
+        const id =
+            target !== undefined && !item.id.startsWith(`${target.c99name}.`)
+                ? `${target.c99name}.${item.id}`
+                : item.id;
```
adam-fowler, line 156: Should the test not be is id of form `target.class/method` return id else find the target and prefix to id.

plemarquand: This check applies to both XCTest and swift-testing tests. Swift-testing can have tests in the form `target.method()`, `target.struct/method()`, `target.struct/anotherstruct/method()`, etc... so it seems safer to check if the test has already been prefixed with the target instead of trying to match the test id form exactly.

adam-fowler (approved):

PR #942, opened 2024-06-27
Fix linker error when building Windows tests w/ coverage

plemarquand:
This fixes the fatalError seen on Windows when attempting to build tests with code coverage enabled. The linker doesn't allow duplicate symbols by default on Windows, while this is OK on macOS/Linux.

Use the `-lldmingw` linker flag to work around this issue.

### DIFF
--- a/src/debugger/buildConfig.ts
+++ b/src/debugger/buildConfig.ts
@@ -336,6 +336,20 @@ export class TestingDebugConfigurationFactory {
         if (isRelease(this.testKind)) {
             result = [...result, "-c", "release", "-Xswiftc", "-enable-testing"];
         }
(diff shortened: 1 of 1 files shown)


adam-fowler (approved):
I guess, I don't know the background to this though.

plemarquand:
@adam-fowler Basically the background is the --enable-code-coverage flag is currently causing a fatalError on Windows because during instrumentation the same symbol is inserted into multiple places, which the linker then fails to link since it doesn't support duplicate symbols on Windows. 

This seems to be desired behaviour to match the existing windows linker (link.exe), but the output of --enable-code-coverage wasn't designed for this. In the future there will be something like `-lld-allow-duplicate-weak` which on Windows SPM will pass automatically when it sees --enable-code-coverage.

Issue #914, opened 2024-06-19
Swift terminal profile


award999:
**Is your feature request related to a problem? Please describe.**
I've made the mistake a few times where I have select a toolchain in vscode and it has been built for via tasks. I then go to the integrated vscode terminal to issue to run some swift command that is a different version of swift.

**Describe the solution you'd like**
You can provide different terminal profiles https://code.visualstudio.com/api/references/vscode-api#TerminalProfileProvider so we could provide a Swift terminal profile with selected toolchain in the PATH so that user can select this swift profile in settings, for example `terminal.integrated.defaultProfile.osx`

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

matthewbastien:
There's also the option of just adding to the terminal environment directly using the `ExtensionContext`:

```typescript
context.environmentVariableCollection.clear();
context.environmentVariableCollection.prepend(
    "PATH",
    "/path/to/usr/bin:",
    { applyAtShellIntegration: true }
);
for (const variable in configuration.swiftEnvironmentVariables) {
    context.environmentVariableCollection.replace(
        variable,
        configuration.swiftEnvironmentVariables[variable],
        { applyAtShellIntegration: true }
    );
}
```

Using this approach the user will have the environment automatically populated in their VS Code terminal without having to launch a special profile. The only minor annoyance being that modifying certain variables (e.g. `PATH`) triggers a warning before the terminal will be updated:

<img width="522" alt="Screenshot 2024-06-27 at 2 30 02 PM" src="https://github.com/swiftlang/vscode-swift/assets/65295938/1bd08676-6cd5-413e-a334-28321f1c8bcc">

adam-fowler:
@matthewbastien so that appears every time you start up a terminal window or only for existing terminal windows?

matthewbastien:
@adam-fowler the warning will only appear for existing terminal windows that need to be reloaded to capture new environment variables.

award999:
@matthewbastien if I disable `swift.enableTerminalEnvironment` and then create a new terminal with the swift profile, `swift.swiftEnvironmentVariables` values are added but `which swift` still indicates `/usr/bin/swift` is being used not the selected toolchain path.

PR #944, opened 2024-06-27
Set Swift environment variables in the integrated terminal

matthewbastien:
This PR adds two new features to the extension:

1. The `swift.path` and `swift.swiftEnvironmentVariables` settings will automatically be applied to the integrated terminal's environment variables. This can be disabled using the setting `swift.enableTerminalEnvironment` if the user wants.
2. A new Terminal Profile has been added primarily for users that wish to have a special terminal for Swift related commands and will disable the automatic terminal environment.

I've also added an icon font with the Swift icon that can be used in various places. At the moment it is being used for the Terminal Profile and the Swiftly installation from the toolchain selection workflow.

Issue: #914


adam-fowler (commented):
This looks good, but what is the point of a Swift terminal which doesn't pass on the extension environment variables?

Issue #564, opened 2023-07-04
Showing Macro Expansion


filip-sakel:
**Is your feature request related to a problem? Please describe.**
Without seeing what code macros generate, it's hard to understand macro-related errors. For example, a macro might try to access a variable "_count" but forget the underscore prefix: "count". Being able to see the macro expansion would make it easy to identify the problem, which is critical for developing macros.

**Describe the solution you'd like**
An Xcode-like action could be shown when opening the context menu (when right clicking) on a macro. Then, after finding the expansion, the extension could be shown below the code using the macro (similar to inlay hints). Otherwise, a pop up showing expanded code could appear (similar to when inspecting documentation).

**Describe alternatives you've considered**
The alternative is to use a compiler flag to see macro expansions. That lacks specificity, as all macro expansions are shown. Even if the compiler gains the ability to only show the expansion of one macro, typing that macro's name or source location into the command line can hamper productivity. Instead, it's much more convenient to click on the macro, right in the editor, and be shown its expansion.

**Additional context**
Related SourceKit LSP issue https://github.com/apple/sourcekit-lsp/issues/755

adam-fowler:
As you have seen I've already added an issue for sourcekit-lsp. I've had a look through the VSCode APIs and so far haven't found an easy way to display extra content inside an TextEditor window. I'll need to investigate further.

filip-sakel:
I found [this](https://github.com/microsoft/vscode-cpptools/issues/1734) discussion in the cpp-tools repo, where macro expansions look like so:

![image](https://github.com/swift-server/vscode-swift/assets/51793967/fa8e14b8-407e-4149-bf7a-789f05cc04d6)

adam-fowler:
As an aside there is already a code action "Inline Macro" which will expand the macro into your file. You can use that plus Ctrl+Z to view the macro and revert

filip-sakel:
Thanks! I tried that but no code action showed up :( Are there specific version requirements (for Swift or even VSCode) for this to work?

adam-fowler:
Probably best to install Xcode beta 3. Beta 2 version of sourcekit-lsp had issues.
You click on the macro and a little lightbulb should appear next to it, click on the lightbulb and a series of options are shown including "Inline Macro"

PR #945, opened 2024-06-28
Handle `PeekDocumentsRequest` to show documents from sourcekit-lsp (like, Macro Expansions) in a "peeked" editor 

lokesh-tr:
This PR utilises the following PR in sourcekit-lsp: https://github.com/swiftlang/sourcekit-lsp/pull/1479
This takes a custom `PeekDocumentsRequest` and opens a peek window to display each given document. In the current state, this request is being used by `ExpandMacroCommand` in sourcekit-lsp to display Macro Expansions.

To check out this functionality, build the `main` branch of `sourcekit-lsp`. Then, configure your "swift.sourcekit-lsp.serverPath" to the built sourcekit-lsp. (This step will not be necessary once this build of sourcekit-lsp is packaged with Swift itself given that you use the latest Swift Toolchain)

*Since the macro expansions feature is behind an experimental feature flag*, configure your swift extension settings as follows:
```
"swift.sourcekit-lsp.serverArguments": [
  "--experimental-feature", "show-macro-expansions"
],
```
(The above step will not be necessary when showing macro expansions comes to a mature stage in the sourcekit-lsp side and thus will be out of its experimental feature flag)

The "Expand Macro" will be available as a Code Action when the cursor sits on a Macro annotation, and selecting the code action will trigger the `ExpandMacroCommand` -> `PeekDocumentsRequest`, and thus shows the Macro Expansions inside a "peeked" editor.

**Note for extension / plugin developers of other editor ecosystems:**
If you are an extension developer wanting to implement this custom functionality in your own editor, set the experimental client capability `workspace/peekDocuments` to `true`, in order to make `sourcekit-lsp` to send the `PeekDocumentsRequest`. From there, you can handle the request and display the documents in any way as you wish, as long as it is "inline" to the editor (or) "peeks" the documents inside the editor.

Fixes https://github.com/swiftlang/vscode-swift/issues/564

-------

### DIFF
--- a/src/sourcekit-lsp/LanguageClientManager.ts
+++ b/src/sourcekit-lsp/LanguageClientManager.ts
@@ -20,6 +20,7 @@ import { isPathInsidePath, swiftRuntimeEnv } from "../utilities/utilities";
 import { Version } from "../utilities/version";
 import { FolderEvent, WorkspaceContext } from "../WorkspaceContext";
 import { activateLegacyInlayHints } from "./inlayHints";
+import { activatePeekDocuments } from "./peekDocuments";
 import { FolderContext } from "../FolderContext";
 import { LanguageClient } from "vscode-languageclient/node";
 import { ArgumentFilter, BuildFlags } from "../toolchain/BuildFlags";
@@ -108,6 +109,7 @@ export class LanguageClientManager {
     private languageClient: langclient.LanguageClient | null | undefined;
     private cancellationToken?: vscode.CancellationTokenSource;
     private legacyInlayHints?: vscode.Disposable;
+    private peekDocuments?: vscode.Disposable;
     private restartedPromise?: Promise<void>;
     private currentWorkspaceFolder?: vscode.Uri;
     private waitingOnRestartCount: number;
@@ -244,6 +246,7 @@ export class LanguageClientManager {
         this.cancellationToken?.cancel();
         this.cancellationToken?.dispose();
         this.legacyInlayHints?.dispose();
+        this.peekDocuments?.dispose();
         this.subscriptions.forEach(item => item.dispose());
         this.languageClient?.stop();
         this.namedOutputChannels.forEach(channel => channel.dispose());
@@ -392,6 +395,8 @@ export class LanguageClientManager {
         this.currentWorkspaceFolder = workspaceFolder?.uri;
         this.legacyInlayHints?.dispose();
         this.legacyInlayHints = undefined;
+        this.peekDocuments?.dispose();
+        this.peekDocuments = undefined;
         if (client) {
             this.cancellationToken?.cancel();
             this.cancellationToken?.dispose();
@@ -559,6 +564,9 @@ export class LanguageClientManager {
                 })(),
             },
             errorHandler: new SourceKitLSPErrorHandler(5),
+            initializationOptions: {
+                "workspace/peekDocuments": true, // workaround for client capability to handle `PeekDocumentsRequest`
+            },
         };
 
         return new langclient.LanguageClient(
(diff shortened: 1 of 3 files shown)


plemarquand:
This looks great! I tried it out and expanded swift-testing's `#expect` macro. Then I ran the test and it behaved exactly the same, as expected.

This fixes #564

adam-fowler (commented):
@lokesh-tr this all looks good. Shocking how little code it is on the VSCode side. As Alex said in the call let's not make this macro specific and call it PeekDocument or something similar.

If we could somehow work out how to display multiple peek views this would be even better.

lokesh-tr:
> This looks great! I tried it out and expanded swift-testing's `#expect` macro. Then I ran the test and it behaved exactly the same, as expected.
> 
> This fixes #564

Thanks @plemarquand for testing this out. 😄

lokesh-tr:
@adam-fowler 
> @lokesh-tr this all looks good. Shocking how little code it is on the VSCode side

This is due to the fact that we didn't try to create a custom scheme and handle the fetching of documents through the scheme, etc. Other than that,  I had finished the major heavy lifting works in sourcekit-lsp side already.

> As Alex said in the call let's not make this macro specific and call it PeekDocument or something similar.

I have just now pushed the code to the PR (https://github.com/swiftlang/sourcekit-lsp/pull/1479) with all the necessary updates as we discussed in the call. Let's see how the review goes.

> If we could somehow work out how to display multiple peek views this would be even better.

I wish that's possible. I am still trying to research on this and find a solution, but so far, no clue.

lokesh-tr:
@adam-fowler I guess this PR is now ready for review. I will squash the commits, once we pass the review stage.

plemarquand:

lokesh-tr:
@adam-fowler The PR in `sourcekit-lsp` has been merged: https://github.com/swiftlang/sourcekit-lsp/pull/1479. I would like to have your comments if there are any. I had also squashed the commits. Let's get this shipped. 🚀

src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -616,6 +620,36 @@ export class LanguageClientManager {
         this.languageClient = client;
         this.cancellationToken = new vscode.CancellationTokenSource();
 
+        this.languageClient.onRequest(
```
adam-fowler, line 623: Could we stick this in a separate function and also call it once the LSP server has started up next to where `activateLegacyInlayHints` is called.

src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -559,6 +560,9 @@ export class LanguageClientManager {
                 })(),
             },
             errorHandler: new SourceKitLSPErrorHandler(5),
+            initializationOptions: {
+                "workspace/peekDocuments": true, // workaround for client capability to handle `PeekDocumentsRequest`
```
adam-fowler, line 564: Is it worthwhile sticking the capability inside another block eg "capabilities", just to separate it from any future initialization options. @ahoppen what do you think? We've discussed in the past using the `initializationOptions` as a way to pass commands for codeActions. 
```suggestion
                "capabilities": {
                    "workspace/peekDocuments": true, // workaround for client capability to handle `PeekDocumentsRequest`
                }
```

adam-fowler (changes requested):
A couple of minor changes, but everything else looks good

src/sourcekit-lsp/LanguageClientManager.ts
ahoppen: Because the key of the initialization option is the request method name, there shouldn’t be any conflicts, so I’m fine with the current design.

lokesh-tr: @adam-fowler Yeah, I too believe there won't be any conflicts.

adam-fowler (approved):
we are good

PR #960, opened 2024-07-08
Test strategy proposal

award999:
* Propose levels of testing and the test matrix
* Propose a single mocking library to introduce unit testing
* Add test strategy documentation


adam-fowler (approved):
I'm happy to go ahead with this

Issue #939, opened 2024-06-27
Add action to the command palette to re-index a project


ahoppen:
https://github.com/swiftlang/sourcekit-lsp/pull/1507 added a `workspace/triggerReindex` request to SourceKit-LSP, which re-indexes a project in SourceKit-LSP if background indexing is enabled.

The request is not intended to be something that users invoke regularly but it is useful as a workaround should SourceKit-LSP’s index up-to-date tracker get out of sync with the project state (which we know can happen in some situations like https://github.com/apple/sourcekit-lsp/issues/1264). The workflow should thus be as follow:
- The command palette has a `SourceKit-LSP: Re-index project` item
- When it is selected, the `workspace/triggerReindex` is sent to the SourceKit-LSP server. 
- If a `methodNotFound` error (error code -32601) is returned, a custom error message should be displayed in VS Code that says that says: *The installed version of SourceKit-LSP does not support background indexing*. If any other error is returned, that error should be displayed as a popup.
- If the request succeeds, a popup should be displayed saying something like: *Re-indexing a project should never be necessary and indicates a bug in SourceKit-LSP. Please file an issue at https://github.com/swiftlang/sourcekit-lsp/issues/new describing which symbol was out-of-date and how you got into the state*. If the user chooses to file an issue, we should enter the workflow to create a diagnose bundle.

ahoppen:
Implemented by https://github.com/swiftlang/vscode-swift/pull/964

PR #964, opened 2024-07-10
Add Re-Index Project Command

plemarquand:
Adds a command to trigger a re-indexing of the open project. Should only be needed when background indexing is enabled and it gets out of sync. While that is a bug, this can act as a temporary workaround.

Blocked By: https://github.com/swiftlang/sourcekit-lsp/pull/1561, https://github.com/swiftlang/sourcekit-lsp/pull/1563
Issue: #939

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -32,6 +32,7 @@ import { SwiftExecOperation, TaskOperation } from "./tasks/TaskQueue";
 import { SwiftProjectTemplate } from "./toolchain/toolchain";
 import { showToolchainSelectionQuickPick, showToolchainError } from "./ui/ToolchainSelection";
 import { captureDiagnostics } from "./commands/captureDiagnostics";
+import { reindexProjectRequest } from "./sourcekit-lsp/lspExtensions";
 
 /**
  * References:
@@ -656,6 +657,25 @@ function restartLSPServer(workspaceContext: WorkspaceContext): Promise<void> {
     return workspaceContext.languageClientManager.restart();
 }
 
+/** Request that the SourceKit-LSP server reindexes the workspace */
+function reindexProject(workspaceContext: WorkspaceContext): Promise<unknown> {
+    return workspaceContext.languageClientManager.useLanguageClient(async (client, token) => {
+        try {
+            return await client.sendRequest(reindexProjectRequest, {}, token);
+        } catch (err) {
+            const error = err as { code: number; message: string };
+            // methodNotFound, version of sourcekit-lsp is likely too old.
+            if (error.code === -32601) {
+                vscode.window.showWarningMessage(
+                    "The installed version of SourceKit-LSP does not support background indexing."
+                );
+            } else {
+                vscode.window.showWarningMessage(error.message);
+            }
+        }
+    });
+}
+
 /** Execute task and show UI while running */
 async function executeTaskWithUI(
     task: vscode.Task,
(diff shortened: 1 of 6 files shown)


plemarquand:

src/commands.ts
```diff
@@ -656,6 +657,25 @@ function restartLSPServer(workspaceContext: WorkspaceContext): Promise<void> {
...
+/** Request that the SourceKit-LSP server reindexes the workspace */
+function reindexProject(workspaceContext: WorkspaceContext): Promise<unknown> {
+    return workspaceContext.languageClientManager.useLanguageClient(async (client, token) => {
+        try {
+            return await client.sendRequest(reindexProjectRequest, {}, token);
+        } catch (err) {
+            const error = err as { code: number; message: string };
+            // methodNotFound, version of sourcekit-lsp is likely too old.
+            if (error.code === -32601) {
+                vscode.window.showWarningMessage(
```
matthewbastien, line 669: Is there a way to check this in advance so that we can only show the command in the command palette if background indexing is enabled?

src/commands.ts
```diff
@@ -817,6 +837,9 @@ export function register(ctx: WorkspaceContext): vscode.Disposable[] {
         vscode.commands.registerCommand("swift.debugSnippet", () => debugSnippet(ctx)),
         vscode.commands.registerCommand("swift.runPluginTask", () => runPluginTask()),
         vscode.commands.registerCommand("swift.restartLSPServer", () => restartLSPServer(ctx)),
+        ...(ctx.swiftVersion.isGreaterThanOrEqual(new Version(6, 0, 0))
+            ? [vscode.commands.registerCommand("swift.reindexProject", () => reindexProject(ctx))]
```
matthewbastien, line 841: You can actually add a `"when"` clause to the package.json to hide the command under certain circumstances. Otherwise, the command will still show up in the command palette and give an ugly error message if the user tries to run it. Take a look at `contextKeys.ts` to see how to add a new context key. There are a bunch of examples for the `"when"` clause in the package.json already.

matthewbastien (changes requested):

src/commands.ts
plemarquand: Ideally we'd solve both these comments with a context key called `supportsReindexing`. It would be set by `languageClient.initializeResult.capatiblities.experimental['workspace/triggerReindex']` which would be returned when you initialize the language server. However this capability doesn't exist. 

@ahoppen Is this possible to add? If you agree I can work up a PR.

src/commands.ts
adam-fowler: SourceKit-LSP should report this capability

plemarquand:
I've updated the PR to check if the `workspace/triggerReindex` capability is reported by sourcekit-lsp. There is a corresponding PR here: https://github.com/swiftlang/sourcekit-lsp/pull/1563

matthewbastien (approved):

adam-fowler (approved):
👍

ahoppen:
Do you  think it would also be possible to address this from https://github.com/swiftlang/vscode-swift/issues/939?

> If the request succeeds, a popup should be displayed saying something like: Re-indexing a project should never be necessary and indicates a bug in SourceKit-LSP. Please file an issue at https://github.com/swiftlang/sourcekit-lsp/issues/new describing which symbol was out-of-date and how you got into the state. If the user chooses to file an issue, we should enter the workflow to create a diagnose bundle.

Issue #963, opened 2024-07-10
Specify list of files/directories to omit from code coverage


plemarquand:
Currently the extension generates a coverage report from all .swift files in the project except for those in Tests. It would be nice to be able to add a list of files/directories to a denylist so they are omitted from the coverage report. This could be defined in a new setting that is a list of paths/filenames qualified from the workspace root.

PR #965, opened 2024-07-10
Add setting to exclude files/directories from code coverage reports

plemarquand:
Adds `swift.excludeFromCodeCoverage` which provides a list of absolute or relative paths to exclude from the code coverage reports. Relative paths are specified from the workspace root.

Issue: #963

### DIFF
--- a/src/coverage/LcovResults.ts
+++ b/src/coverage/LcovResults.ts
@@ -198,6 +198,8 @@ export class TestCoverage {
     private ignoredFilenamesRegex(): string {
         const basePath = this.folderContext.folder.path;
         const buildFolder = path.join(basePath, ".build");
+        const snippetsFolder = path.join(basePath, "Snippets");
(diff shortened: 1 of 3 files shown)


src/coverage/LcovResults.ts
```diff
@@ -202,7 +202,11 @@ export class TestCoverage {
             .getTargets(TargetType.test)
             .map(target => path.join(basePath, target.path));
 
-        return [buildFolder, ...testTargets].join("|");
+        const excluded = configuration.excludeFromCodeCoverage.map(target =>
+            path.isAbsolute(target) ? target : path.join(basePath, target)
+        );
+
+        return [buildFolder, ...testTargets, ...excluded].join("|");
```
adam-fowler, line 209: Should we be ignoring the snippets folder as well

adam-fowler (approved):
👍

PR #967, opened 2024-07-12
Add Throws to doc comment template

plemarquand:
Makes the behaviour consistent with Xcode, where if a function throws there is a templated field to describe the error.

Adds tests for the existing behaviour, as well as this new behaviour.


adam-fowler (approved):
Looks good. I guess this will become more important when typed throws come in.

PR #971, opened 2024-07-17
Retrieve macro expansions for `PeekDocumentsRequest` through `GetReferenceDocumentRequest` instead of showing temporary files

lokesh-tr:
This PR utilises the newly introduced `GetReferenceDocumentRequest` to resolve the contents of macro expansion obtained from `PeekDocumentRequest` through a new scheme.

This change doesn't cause any behavioural changes facing the user.
The `GetReferenceDocumentRequest` implemented in sourcekit-lsp allows it to not create temporary files whenever it makes a `PeekDocumentRequest`.

#### References:
Accompanying PR in sourcekit-lsp repository: https://github.com/swiftlang/sourcekit-lsp/pull/1567

------

### DIFF
--- a/src/sourcekit-lsp/getReferenceDocument.ts
+++ b/src/sourcekit-lsp/getReferenceDocument.ts
@@ -1,0 +1,40 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2024 the VS Code Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of VS Code Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import * as vscode from "vscode";
+import * as langclient from "vscode-languageclient/node";
+import { GetReferenceDocumentParams, GetReferenceDocumentRequest } from "./lspExtensions";
+
+export function activateGetReferenceDocument(client: langclient.LanguageClient): vscode.Disposable {
+    const getReferenceDocument = vscode.workspace.registerTextDocumentContentProvider(
+        "sourcekit-lsp",
+        {
+            provideTextDocumentContent: async (uri, token) => {
+                const params: GetReferenceDocumentParams = {
+                    uri: uri.toString(true),
+                };
+
+                const result = await client.sendRequest(GetReferenceDocumentRequest, params, token);
+
+                if (result) {
+                    return result.content;
+                } else {
+                    return "Unable to retrieve reference document";
+                }
+            },
+        }
+    );
+
+    return getReferenceDocument;
+}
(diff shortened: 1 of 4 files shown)


src/sourcekit-lsp/getReferenceDocument.ts
```diff
@@ -0,0 +1,45 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import * as vscode from "vscode";
+import * as langclient from "vscode-languageclient/node";
+import { GetReferenceDocumentParams, GetReferenceDocumentRequest } from "./lspExtensions";
+
+export function activateGetReferenceDocument(client: langclient.LanguageClient): vscode.Disposable {
+    const getReferenceDocument = vscode.workspace.registerTextDocumentContentProvider(
+        "swift-macro-expansion",
```
adam-fowler, line 21: As with the peek document can we keep this generic so it is possible to use in other situations. Maybe `reference-document`. 

@ahoppen thoughts?

src/sourcekit-lsp/getReferenceDocument.ts
```diff
@@ -0,0 +1,45 @@
...
+                const expansion = await client.sendRequest(
+                    GetReferenceDocumentRequest,
+                    params,
+                    token
+                );
+
+                if (expansion) {
+                    return expansion.content;
+                } else {
+                    vscode.window.showWarningMessage("Unable to retrieve Macro Expansion");
```
adam-fowler, line 37: Instead of showing a warning message, perhaps just return the contents of the document as `Unable to retrieve reference document ...`

adam-fowler (changes requested):
Looks good, mainly looking to not make this macro specific because it could be used in other situations. @ahoppen perhaps this could be used for generated swiftinterface files

src/sourcekit-lsp/getReferenceDocument.ts
ahoppen: Interesting thought to just use one URL scheme. 

My thought was that we would add another URL scheme for Swift interfaces in the future and add that to the VS Code extension as well. The benefit of having different URL schemes would be that we can use the path of the URL to actually carry data. If we use one scheme, we need to use the path portion to disambiguate which kind of document we want. 

But given that we already use the path for a dual purpose (the last component is the display name to be used in VS Code), I agree that it would be good to change the URL scheme to 

```swift
sourcekit-lsp://<document-type>/<display-name>?parameters=values
```

For macro expansions, `document-type` would be `swift-macro-expansion` and the full URL would be adjusted from the proposal in https://github.com/swiftlang/sourcekit-lsp/pull/1567 to the following (`sourcekit-lsp://macro-expansion` is the only fixed portion here).

```
sourcekit-lsp://macro-expansion/LaCb-LcCd.swift?sourceFile=path_to_source_file&lowerL=&lowerC=&upperL=&upperC=&bufferName=`
```

lokesh-tr:
@adam-fowler The sourcekit-lsp PR got merged just a few hours back. Can you review this PR now?

adam-fowler (approved):
Looks good

adam-fowler:

Issue #970, opened 2024-07-17
unable to debug the tests in a swift package (library not executable) on Windows


dave256:
The discussion of this issue started here:
https://forums.swift.org/t/vs-code-wont-run-swift-on-windows/72721/22

If you create a package on Windows 11 with
```
swift package init
```

and then use VSCode to open the directory containing the package, you can run the package, but you can't debug the package.

I've tried this with the Swift 6 Toolchain (from the July 2nd build available on swift.org) but I also believe it doesn't work with Swift 5.10 release on swift.org. 

If I right click in the gutter and choose to Debug the test, a Windows dialog pops up complaining "unable to find executable in the Swift\Platforms\...\xctest.exe"

I turned on the Swift extension setting to "Output additional diagnostics to the Swift Output View" and this is what appears there when I right click in the gutter for the XCTestCase subclass and choose to debug.

```
09:51:56: SourceKit-LSP setup
09:51:57: Failed to find "C:\Users\roger\AppData\Local\Programs\Swift\Toolchains\6.0.0+Asserts\usr\bin\lldb-dap.exe"
09:52:18: NewPlayingCards: add: c:\Users\roger\Desktop\NewPlayingCards
09:52:18: NewPlayingCards: Resolving Dependencies: starting ... 
09:52:18: Exec Task: (NewPlayingCards) swift package resolve
09:52:18: NewPlayingCards: focus: c:\Users\roger\Desktop\NewPlayingCards
09:52:23: NewPlayingCards: Resolving Dependencies: ... done.
10:06:39: Exec Task: swift build --build-tests -Xlinker -debug:dwarf -Xswiftc -diagnostic-style=llvm
[...]

PR #972, opened 2024-07-18
Use correct xctest path when debugging on Windows

plemarquand:
The windows debug config was delegating to the macOS debugging config, which runs xctests through a prebuilt executable included in the toolchain. On Windows, like Linux, an xctest executable is built during `swift test`/`swift build --build-tests`. This should be used instead.

This patch brings the Windows config in line with what we do on Linux.

Issue: #970


adam-fowler (approved):
Damn I should have caught that.

PR #977, opened 2024-07-19
Add XCTest.dll path to Path when testing on Windows

adam-fowler:


src/debugger/buildConfig.ts
```diff
@@ -89,14 +89,42 @@ export class TestingDebugConfigurationFactory {
...
+    private buildWindowsConfig(): vscode.DebugConfiguration | null {
+        if (isDebugging(this.testKind) && this.testLibrary === TestLibrary.xctest) {
+            const testEnv = {
+                ...swiftRuntimeEnv(),
+                ...configuration.folder(this.ctx.workspaceFolder).testEnvironmentVariables,
+            };
+            // On Windows, add XCTest.dll to the Path
+            // and run the .xctest executable from the .build directory.
+            const runtimePath = this.ctx.workspaceContext.toolchain.runtimePath;
+            const xcTestPath = this.ctx.workspaceContext.toolchain.xcTestPath;
```
plemarquand, line 110: It is possible xcTestPath  could be undefined, check that it isn't undefined as well before adding it to the Path

PR #978, opened 2024-07-19
Add XCTest.dll path to Path when testing on Windows

plemarquand:
Add XCTest.dll to PATH and supply arguments when running swift-testing tests. Also fixes extra whitespace in output when debugging XCTests on Windows.


adam-fowler (approved):
looks good

PR #982, opened 2024-07-22
Improve error reporting when SourceKit-LSP fails to start

plemarquand:
If SourceKit-LSP fails to initialize, typically because of unrecognized launch arguments, don't attempt to restart it and limit the number of erorr dialogs to just one descriptive one. Clicking the "Go to Output" button on this dialog takes you to the SourceKit-LSP Output panel which now shows the SourceKit-LSP launch error.

### DIFF
--- a/src/sourcekit-lsp/LanguageClientManager.ts
+++ b/src/sourcekit-lsp/LanguageClientManager.ts
@@ -434,12 +434,14 @@ export class LanguageClientManager {
             this.languageClient = undefined;
             return;
         }
-        return this.startClient(...this.createLSPClient(folder));
+        const { client, errorHandler } = this.createLSPClient(folder);
+        return this.startClient(client, errorHandler);
     }
 
-    private createLSPClient(
-        folder?: vscode.Uri
-    ): [langclient.LanguageClient, SourceKitLSPErrorHandler] {
+    private createLSPClient(folder?: vscode.Uri): {
+        client: langclient.LanguageClient;
+        errorHandler: SourceKitLSPErrorHandler;
+    } {
         const toolchainSourceKitLSP =
             this.workspaceContext.toolchain.getToolchainExecutable("sourcekit-lsp");
         const lspConfig = configuration.lsp;
@@ -574,15 +576,15 @@ export class LanguageClientManager {
             },
         };
 
-        return [
-            new langclient.LanguageClient(
+        return {
+            client: new langclient.LanguageClient(
                 "swift.sourcekit-lsp",
                 "SourceKit Language Server",
                 serverOptions,
                 clientOptions
             ),
             errorHandler,
-        ];
+        };
     }
 
     private async startClient(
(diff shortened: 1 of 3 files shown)


plemarquand:
Here is a screen recording of the current behaviour when SourceKit-LSP fails to start due to a bad argument.
https://github.com/user-attachments/assets/ec00dd8e-08f1-4929-bc5c-e5a4a4b8bb1e

plemarquand:
Here is the new dialog (well, same as the old dialog, except it sticks around and you can read it).

NewSourceKitErrorMessage

Here is the output you see when you click "Go to Output". The error is not automatically highlighted, that is just to show that the error is now properly reported.

NewSourceKitErrorLogs

plemarquand:

src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -609,15 +619,27 @@ export class LanguageClientManager {
...
                 if (this.workspaceContext.swiftVersion.isLessThan(new Version(5, 7, 0))) {
                     this.legacyInlayHints = activateLegacyInlayHints(client);
                 }
 
                 this.peekDocuments = activatePeekDocuments(client);
             })
             .catch(reason => {
-                this.workspaceContext.outputChannel.log(`${reason}`);
-                // if language client failed to initialise then shutdown and set to undefined
+                const lspOutputChannel = client.outputChannel as SwiftOutputChannel;
```
adam-fowler, line 633: There seems to be some weird timing error where this code doesn't seem to be called when you pass in a bad command line parameter to SKLSP, unless you stick breakpoints in here.

src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -434,11 +434,12 @@ export class LanguageClientManager {
...
         }
-        const client = this.createLSPClient(folder);
-        return this.startClient(client);
+        return this.startClient(...this.createLSPClient(folder));
     }
 
-    private createLSPClient(folder?: vscode.Uri): langclient.LanguageClient {
+    private createLSPClient(
+        folder?: vscode.Uri
+    ): [langclient.LanguageClient, SourceKitLSPErrorHandler] {
```
adam-fowler, line 442: Would it not be cleaner to return 
```
{client: langclient.LanguageClient, errorHandler: SourceKitLSPErrorHandler}
```

adam-fowler (commented):
The bits of this PR that work seem to be about ensuring the error handler doesn't try to restart the server if initialisation failed. Which already makes things clearer as you get a dialog saying there was an issue go to the LSP output to find out what it was.

The parts which attempt to parse the log don't seem to get called. Maybe you could move them to the error handler instead of in the catch block.

src/sourcekit-lsp/LanguageClientManager.ts
plemarquand: I just reverted this block, it doesn't actually get called when sourcekit-lsp fails to launch. I think its only if there is some error on the VS Code side.

award999 (approved):

award999:
Verified

Issue #981, opened 2024-07-21
In Test Results pane display text associated with individual test


adam-fowler:
In the Test Results screen you can expand a test run to see all the test cases that were run. When you click on one of these test cases the Test Results screen reports `The test case did not report any output.` Can we associated the output from individual test cases with their test case so we get some output here.

plemarquand:
Fixed in #985

award999:
Verified with d6c91dd379abd6a402dddb3503e35fd328c6ce21

PR #985, opened 2024-07-23
Associate test run output with individual tests

plemarquand:
For both XCTests and swift-testing associate each line of the output with the appropriate test. This lets users view just the output for a given test via the test history pane, and also shows the output for the given test when navigating through the test run issues via prev/next issue buttons. Clicking a test issue in the test history pane highlights the issue inline and then clicking the "Go to Previous Test Failure" button will show complete output captured for the test.

A current limitation is this does not capture output from user `print()` calls in the same order they'd appear in stdout. We do the best we can to capture prints from stdout and inject them in to the test output, but because we're mixing this with output generated from the JSON event stream the exact position of print calls in the output is not guaranteed.

Hopefully this can be improved in the future and stdout output from tests can be emitted on the JSON event stream, where we could then associate it with a specific test.

The existing output parsing tests are expanded and reworked to more completely assert against the state of a parsed test.

Issue: #981

### DIFF
--- a/assets/test/defaultPackage/Tests/PackageTests/PackageTests.swift
+++ b/assets/test/defaultPackage/Tests/PackageTests/PackageTests.swift
@@ -41,7 +41,9 @@ final class DebugReleaseTestSuite: XCTestCase {
 #if swift(>=6.0)
 import Testing
 
-@Test func topLevelTestPassing() {}
+@Test func topLevelTestPassing() {
+  print("A print statement in a test.")
+}
 @Test func topLevelTestFailing() {
   #expect(1 == 2)
 }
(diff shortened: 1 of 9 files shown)


award999 (commented):
Code LGTM. Pull and tried out the patch and seems to be working well

adam-fowler:
Shame you can't associate stdout output with the test that generated it. I had assumed that swift-testing would push all the output include stdout/err through its json stream.

adam-fowler (approved):
👍

plemarquand:
@adam-fowler I believe that's the long term goal, but the machinery at the language level isn't there yet. Hopefully one day we can revisit and tighten this up.

PR #986, opened 2024-07-25
Support a Unified Test Binary

plemarquand:
With the merge of https://github.com/swiftlang/swift-package-manager/pull/7766 a `swift build --build-tests` produces a single binary that contains both XCTests and swift-testing tests.

Update the build configs when debugging tests to use this new unified test binary if it exists.

To determine if we're running a unified binary on macOS we can check for the presence of the swiftpm-testing-helper, which is used to run the swift-testing tests in the unified binary.

This utility isn't required on Windows and Linux as the test binary can be invoked directly. To determine if we're running a unified binary we check if the latest build produced a .swift-testing binary in the build folder. If it doesn't exist, or if it does exist but its `mtime` is older than the last build start time, we know its a unified binary. The `mtime` check is required in case the user updates their toolchain to a version that produces a unified binary but still has a .swift-testing binary in their .build folder left over from an old build.

In the future we can remove these checks once we're confident users on 6.0 and main toolchains have moved on to versions that generate a unified test binary. Tentatively I propose this to be around when Swift 6.0 is finalized and released.

This change only affects debugging tests. All other testing code paths use `swift test` which lets SwiftPM automatically handle the switch to the new unified binary.

### DIFF
--- a/src/TestExplorer/TestParsers/TestEventStreamReader.ts
+++ b/src/TestExplorer/TestParsers/TestEventStreamReader.ts
@@ -74,8 +74,7 @@ export class UnixNamedPipeReader implements INamedPipeReader {
                     });
 
                     readable.on("drain", () => pipe.resume());
-
-                    pipe.on("error", () => fs.close(fd));
+                    pipe.on("error", () => pipe.close());
                     pipe.on("end", () => {
                         readable.push(null);
                         fs.close(fd);
--- a/src/debugger/buildConfig.ts
+++ b/src/debugger/buildConfig.ts
@@ -15,6 +15,7 @@
 import * as os from "os";
 import * as path from "path";
 import * as vscode from "vscode";
+import * as fs from "fs/promises";
 import configuration from "../configuration";
 import { FolderContext } from "../FolderContext";
 import { BuildFlags } from "../toolchain/BuildFlags";
@@ -32,35 +33,38 @@ import { TestKind, isDebugging, isRelease } from "../TestExplorer/TestKind";
  * and `xcTestConfig` functions to create
  */
 export class TestingDebugConfigurationFactory {
-    public static swiftTestingConfig(
+    public static async swiftTestingConfig(
         ctx: FolderContext,
         fifoPipePath: string,
         testKind: TestKind,
         testList: string[],
+        buildStartTime: Date,
         expandEnvVariables = false
-    ): vscode.DebugConfiguration | null {
+    ): Promise<vscode.DebugConfiguration | null> {
         return new TestingDebugConfigurationFactory(
             ctx,
             fifoPipePath,
             testKind,
             TestLibrary.swiftTesting,
             testList,
+            buildStartTime,
             expandEnvVariables
         ).build();
     }
 
-    public static xcTestConfig(
+    public static async xcTestConfig(
         ctx: FolderContext,
         testKind: TestKind,
         testList: string[],
         expandEnvVariables = false
-    ): vscode.DebugConfiguration | null {
+    ): Promise<vscode.DebugConfiguration | null> {
         return new TestingDebugConfigurationFactory(
             ctx,
             "",
             testKind,
             TestLibrary.xctest,
             testList,
+            null,
             expandEnvVariables
         ).build();
     }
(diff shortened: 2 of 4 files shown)


src/toolchain/toolchain.ts
```diff
@@ -575,6 +579,42 @@ export class SwiftToolchain {
...
 
+    /**
+     * @returns path to the swiftpm-testing-helper binary, if it exists.
+     */
+    private static async getSwiftPMTestingHelperPath(): Promise<string | undefined> {
+        if (process.platform === "darwin") {
+            const developerDir = await this.getXcodeDeveloperDir(
+                configuration.swiftEnvironmentVariables
+            );
+            const swiftPMHelperPath = path.join(
```
award999, line 590: should this not be the selected toolchain?

adam-fowler:
Are we certain `swiftpm-testing-helper` is only used to run swift-testing tests. I'm sure I've seen that process name before the existence of swift-testing. I maybe wrong.

plemarquand:
I consulted @grynspan who mentioned looking for the `swiftpm-testing-helper` would be an appropriate way to check if SwiftPM had moved to the new unified test executable. Both that tool and the unified binaries landed in approximately the same nightly toolchain.

grynspan:
There's a window of maybe two nightly builds where the tool is available but the new build mechanism is not yet implemented. I think that's narrow enough we don't need to worry about it, and the solution for anybody affected is to slightly upgrade or downgrade their nightly toolchains. :)

grynspan:
> Are we certain `swiftpm-testing-helper` is only used to run swift-testing tests. I'm sure I've seen that process name before the existence of swift-testing. I maybe wrong.

There's a `swiftpm-xctest-helper` on macOS that's used by `swift test list`. Different tool. :)

src/TestExplorer/TestParsers/TestEventStreamReader.ts
```diff
@@ -74,8 +74,7 @@ export class UnixNamedPipeReader implements INamedPipeReader {
                     });
 
                     readable.on("drain", () => pipe.resume());
-
-                    pipe.on("error", () => fs.close(fd));
+                    pipe.on("error", () => pipe.close());
```
grynspan, line 77: Drive-by fix, or…?

src/debugger/buildConfig.ts
```diff
@@ -387,37 +422,82 @@ export class TestingDebugConfigurationFactory {
...
+
+        // TODO: When Swift 6 is released and enough time has passed that we're sure no one is building the .swift-testing
+        //       binary anymore this fs.stat workaround can be removed and `swiftTestingPath` can be returned. The
+        //       `buildStartTime` argument can be removed and the build config generation can be made synchronous again.
+
+        try {
+            const stat = await fs.stat(this.swiftTestingOutputPath());
+            return !this.buildStartTime || stat.mtime.getTime() < this.buildStartTime.getTime();
+        } catch {
+            // If the .swift-testing binary doesn't exist then swift-testing tests are in the unified binary.
```
grynspan, line 457: This may not be true if `swift test --disable-experimental-swift-testing` had previously been passed.

src/toolchain/toolchain.ts
grynspan: I believe you're right.

plemarquand: Updated to look in the `toolchainPath` which is either the users selected toolchain or their Xcode, whichever they have configured.

src/TestExplorer/TestParsers/TestEventStreamReader.ts
plemarquand: Yeah, the latest VS Code update means when developing the extension if an error is thrown a breakpoint is automatically hit, which would happen here if the `fd` was already closed. Double closing it wasn't really an issue, but now it makes development annoying. `pipe.close()` should do the same thing without throwing if its already closed.

src/debugger/buildConfig.ts
plemarquand: It shouldn't be possible to get in to this situation since the last run command will be one that we control that either does or does not build the unified test binary. This code path only happens immediately after a `swift build --build-tests` has run in VS Code, and we're doing a test run with swift-testing tests.

plemarquand:

adam-fowler:
> There's a `swiftpm-xctest-helper` on macOS that's used by `swift test list`. Different tool. :)

Yeah you’re probably right

award999 (approved):

grynspan:
Woohoo!

award999:
Verified

PR #993, opened 2024-08-06
Pass build arguments to some swiftExec calls

award999:
These particular commands (or subcommands) take the same build arguments as we pass to "build", "run", and "test" commands. To get our tests to run in a sandbox, we need to disable sandboxing because you cannot create a new sandbox when you're already running under a sandbox


award999:
Leaving in draft state until we have macOS builders to test this

src/SwiftPackage.ts
```diff
@@ -215,9 +216,15 @@ export class SwiftPackage implements PackageContents {
         toolchain: SwiftToolchain
     ): Promise<SwiftPackageState> {
         try {
-            let { stdout } = await execSwift(["package", "describe", "--type", "json"], toolchain, {
-                cwd: folder.fsPath,
-            });
+            let { stdout } = await execSwift(
+                ["package"]
+                    .concat(configuration.buildArguments)
+                    .concat(["describe", "--type", "json"]),
```
adam-fowler, line 222: I prefer the `["package", ...configuration.buildArguments, "describe", "--type", "json"] format, but I guess `concat` works

adam-fowler (commented):
Have you verified that all of these commands will work with all the build arguments.
I'm just wondering if it is safer to have a disable sandbox option which is used in these cases

src/SwiftPackage.ts
plemarquand: Same, we should be consistent about using the spread operator

award999: ACK, ya no problem spreading

award999:
> Have you verified that all of these commands will work with all the build arguments. I'm just wondering if it is safer to have a disable sandbox option which is used in these cases

Ya should have explained in commit message but if you look at:
```
swift help build > build.txt
swift help package > package.txt
diff build.txt package.txt
 ```
 
 Other than "package" having different subcommands, the options are the same other that build can specify target, product, and enable/disable swift-testing and XCTest. In my mind the user wouldn't put these in build arguments to use for everything but maybe you can think of a case I haven't.
 
 Yes the other option is a boolean setting to disable sandboxing but wasn't keen on introducing a new setting for our niche use case, but yes may be safer bet

PR #996, opened 2024-08-07
Remove dependency on swift-testing in defaultPackage test project

plemarquand:
Now that swift-testing is in the toolchain we should no longer need to specify it explicitly as a dependency.


plemarquand:
I jumped the gun, https://github.com/swiftlang/swift/pull/74582 is still in process.

adam-fowler:
This was the issue I wanted to ask whether a nightly had been released with the relevant changes.

PR #1000, opened 2024-08-07
Create separate Build All Release task

plemarquand:
We were naively transforming the arguments of the returned Build All task to create one that would run a release version of Build All. Unfortunately the `task.execution` remained the same and so while upon inspection the task looked like it would run a release build it was not.

Create two separate versions of the Build All task, one for Debug and one for Release builds, and use the release version when building tests in release mode.

### DIFF
--- a/src/TestExplorer/TestRunner.ts
+++ b/src/TestExplorer/TestRunner.ts
@@ -682,20 +682,9 @@ export class TestRunner {
 
     /** Run test session inside debugger */
     async debugSession(token: vscode.CancellationToken, runState: TestRunnerTestRunState) {
-        const buildAllTask = await getBuildAllTask(this.folderContext);
+        const buildAllTask = await getBuildAllTask(this.folderContext, isRelease(this.testKind));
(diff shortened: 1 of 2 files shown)


adam-fowler:
Are you saying even though it is printing `-c release` to the terminal it isn't running a release build

plemarquand:
@adam-fowler correct, we print the task’s description which isn’t necessarily what is executed. What runs, in this case, is determined by the tasks `execution` which is a `SwiftExecution` with its own set of `args`. If you check the terminal panel after running a test in release mode you see a build all has run, but without `-c release`.

award999:
Verified

adam-fowler:
I'm not a great fan of this. It creates a release build all task which virtually nobody needs and is only needed for running tests in release. Do we need to include the release task in the list of tasks returned by the task provider?

award999:
@adam-fowler https://github.com/swiftlang/vscode-swift/pull/1005 should address that

Issue #994, opened 2024-08-07
Remove support for legacy .swift-testing binary


plemarquand:
Once Swift 6.0 and Xcode 15 have been finalized and released, refactor the build config generation code to assume a unified test binary.

#986 introduced support for a unified binary but retained support for running the old `.swift-testing` binary if the user had a 6.0 toolchain that still generated it. Once we're confident users have moved on to generating a unified binary the build config generation can be simplified and the code path for the legacy swift-testing binary can be removed.

More context in the comment here:
https://github.com/swiftlang/vscode-swift/pull/986/files#diff-ed1fc3b638d3e40c8a0266ffea55087d5c16ef1bede54e612f8c1b3f6534288bR449

PR #1004, opened 2024-08-08
Use build description to check for unified testing binary

plemarquand:
We were checking the `mtime` of the `<ProductName>.swift-testing` binary to determine if it was produced by the latest build. The issue with this is that when you run tests over and over without making any code changes the binary is never rewritten because it is already up to date.

Instead of checking the `mtime` of the .swift-testing binary in the build folder to determine if it was produced, use the build `description.json` and check the builtTestProducts list to see if the legacy `.swift-testing` binary was produced by the latest build.

### DIFF
--- a/src/debugger/buildConfig.ts
+++ b/src/debugger/buildConfig.ts
@@ -38,7 +38,6 @@ export class TestingDebugConfigurationFactory {
         fifoPipePath: string,
         testKind: TestKind,
         testList: string[],
-        buildStartTime: Date,
         expandEnvVariables = false
     ): Promise<vscode.DebugConfiguration | null> {
         return new TestingDebugConfigurationFactory(
@@ -47,7 +46,6 @@ export class TestingDebugConfigurationFactory {
             testKind,
             TestLibrary.swiftTesting,
             testList,
-            buildStartTime,
             expandEnvVariables
         ).build();
     }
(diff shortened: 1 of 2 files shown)


src/debugger/buildConfig.ts
```diff
@@ -454,23 +449,33 @@ export class TestingDebugConfigurationFactory {
...
     private async isUnifiedTestingBinary(): Promise<boolean> {
         // Toolchains that contain https://github.com/swiftlang/swift-package-manager/commit/844bd137070dcd18d0f46dd95885ef7907ea0697
         // no longer produce a .swift-testing binary, instead we want to use `unifiedTestingOutputPath`.
         // In order to determine if we're working with a unified binary we need to check if the .swift-testing
-        // binary _doesn't_ exist, and if it does exist we need to check that it wasn't built by an old toolchain
-        // and is still in the .build directory. We do this by checking its mtime and seeing if it is after
-        // the `buildStartTime`.
+        // binary was produced by the latest build. If it was then we are not using a unified binary.
 
         // TODO: When Swift 6 is released and enough time has passed that we're sure no one is building the .swift-testing
```
adam-fowler, line 462: Maybe add an issue for this

adam-fowler (approved):
looks good

award999:
I tested this PR and confirm it fixes the issue I was seeing, and the unified binary approach still works too

award999 (approved):

src/debugger/buildConfig.ts
plemarquand: yep, I created https://github.com/swiftlang/vscode-swift/issues/994 to follow up on this

award999:
Verified

PR #1005, opened 2024-08-08
Don't add "Build All - Release" task to list of provided tasks

award999:
With #1000 merged in, we are now showing the Release variant of the build all task in the list of provided tasks when running the "Run Task" command. We don't want that since only needed for debugging a test in release mode


plemarquand (approved):

award999:

PR #1008, opened 2024-08-09
Fix lldb-dap path on Darwin

plemarquand:
The `lldb-dap` binary is not in the toolchain directory on Darwin like it is on Linux and Windows. Use `xcrun -find lldb-dap` on Darwin to find the correct `lldb-dap` location.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -50,7 +50,7 @@ export interface DebuggerConfiguration {
     /** Are we using debug adapter provided with Toolchain */
     readonly useDebugAdapterFromToolchain: boolean;
     /** Return path to debug adapter */
-    readonly debugAdapterPath: string;
+    readonly customDebugAdapterPath: string;
 }
 
 /** workspace folder configuration */
@@ -149,7 +149,7 @@ const configuration = {
                     .getConfiguration("swift.debugger")
                     .get<boolean>("useDebugAdapterFromToolchain", false);
             },
-            get debugAdapterPath(): string {
+            get customDebugAdapterPath(): string {
                 return vscode.workspace.getConfiguration("swift.debugger").get<string>("path", "");
             },
         };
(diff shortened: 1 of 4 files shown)


adam-fowler:
Has lldb-dap moved? It used to be in the same folder as `swift` etc. Using `xcrun` will mean we get the `lldb-dap` from the currently selected Xcode not the current toolchain in VSCode.

plemarquand:
I dont know if its moved, I don't have the Xcode 16 Beta 1 anymore to check, but its in `/Applications/Xcode.app/Contents/Developer/usr/bin/lldb-dap` not `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lldb-dap`, which is where we're looking today.

Good catch about `xcrun -find` returning the executable in the selected Xcode. I'll have to try another approach. We'll have to update the existing `getLLDB()` function as well, as it also uses `xcrun`.

adam-fowler:
The important thing is lldb and lldb-dap both come from the same source. If we are using xcrun for lldb then I guess we should do the same for lldb-dap. 

Ideally we would use the one that comes with the toolchain we are using though.

award999:
> Ideally we would use the one that comes with the toolchain we are using though.

The find command can specify toolchain, ex. `xcrun --toolchain OSX15_1 -f lldb-dap`

Seems like maybe only the default toolchain does not have `lldb-dap` because the other handful I try seem to, ex. `/Applications/Xcode.app/Contents/Developer/Toolchains/OSX15.1.xctoolchain/usr/bin/lldb-dap`

plemarquand:
@adam-fowler after some double checking, because we're specifying the `DEVELOPER_DIR` as an environment variable for `xcrun` it does search in the VS Code selected toolchain. 

I just pushed an update that searches the selected toolchain for lldb-dap before falling back to looking in Xcode on Darwin, same as how we were looking for `lldb`.

award999 (approved):

adam-fowler (approved):

matthewbastien:
Verified with d6c91dd

Issue #832, opened 2024-05-31
Add a way to run a test multiple times in a row


ahoppen:
Sometimes, when debugging a non-deterministic test, it is useful if you can run it multiple times in a row until it fails. It would be great if the Swift Extension offered that ability.

adam-fowler:
Using swift-testing you could do that.
```swift
@Test(arguments: 0..<100)
func testMultiple(attempt: Int){
    ...
}
```

ahoppen:
I was thinking XCTest ;-)

adam-fowler:
This seems quite niche. Is there any reason you can't wrap the test in a for loop that runs it multiple times.

ahoppen:
I could. I just found the Xcode feature to run a test repeatedly useful quite regularly because it also gives you a few things out of the box like logging which iteration it’s currently on.

plemarquand:
I've used this a few times in Xcode to diagnose flakey XCTests that have setUp/tearDown

adam-fowler:
I didn't know this was an option in Xcode.

award999:
Verified

PR #1009, opened 2024-08-12
Add "Run Test Multiple Times"

plemarquand:
Adds two new items to the context menu when you right click a test in the test explorer:

- Run Multiple Times
- Run Until Failure

Selecting either of these promts with a text input where the user can input how many times they want to run the test(s).

If the user selected Run Until Failure the tests will be run a maximum number of times, stopping at the first iteration that produces a failure.

A current limitation in VS Code is if you have multiple tests selected in the Test Explorer (by shift or ctrl clicking them) only the first is passed to the command.

Issue: #832

### DIFF
--- a/package.json
+++ b/package.json
@@ -201,6 +201,16 @@
         "command": "swift.clearDiagnosticsCollection",
         "title": "Clear Diagnostics Collection",
         "category": "Swift"
+      },
+      {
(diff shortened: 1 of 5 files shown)


award999 (approved):
LGTM

plemarquand:

PR #1012, opened 2024-08-15
Organize commands in to folders

plemarquand:
Break out individual commands from src/commands.ts in to individual files that export their main command function.


plemarquand:

award999 (approved):

Issue #1014, opened 2024-08-16
Tests in suites marked with swift-testing tag are filtered from Test Explorer


plemarquand:
If you add a tag to a suite and filter the test explorer by that tag, the suite appears in the list but its child tests/suites do not. If a suite has a tag, all children of that suite should inherit it as well.

The [swift-testing documentation on tags](https://swiftpackageindex.com/swiftlang/swift-testing/main/documentation/testing/addingtags#Add-a-tag) states: "If any tags are applied to a test suite, then all tests in that suite inherit those tags."

Running the suite while in the test explorer filtered view VS Code runs no tests, because no tests match that tag, only the suite.

See attached screenshot:

<img width="1805" alt="Screenshot 2024-08-16 at 2 56 01 PM" src="https://github.com/user-attachments/assets/0f0a9d04-15b6-41ca-9967-31f4e055093d">

award999:
Verified with d6c91dd379abd6a402dddb3503e35fd328c6ce21

PR #1015, opened 2024-08-16
Test tags from suites should be inherited

plemarquand:
Tags defined on swift-testing suites are meant to be inherited by child suites/tests. VS Code doesn't apply this inheritance automatically, so we must set a suite's tags on its children ourselves.

Issue: #1014


matthewbastien (approved):
lgtm

PR #1016, opened 2024-08-16
REPL: introduce a new jupyter notebook based REPL

compnerd:
VSCode has introduced a native jupyter notebook API. Wire up a "kernel" for Swift to provide a native REPL experience within VSCode. This is primarily a PoC implementation that demonstrates how such an implementation could work.

The primary limitation in the current implementation is the IO processing in the LLDB REPL. We can apply a heuristic to capture the error message (termination on an empty string), but do not have such a heuristic for the output and currently only capture the first line of the output, leaving the remaining output in the buffer. Improving the stream handling would significantly improve the experience.


compnerd:
@plemarquand anything to do here to move this along?

plemarquand:
@compnerd I did finally get a chance to try this out last week and found that the "Evaluate in REPL" functionality didn't work as I expected. Putting the cursor on `print("Hello")` and selecting `> Swift: Evaluate in REPL` brought the code in to a cell in a new notebook but didn't evaluate it, and there doesn't seem to be a way to evaluate or delete the cell. I also get a toolbar of buttons like Clear All, Variables, Export, etc but these aren't hooked up and so don't do anything. 

If I write the same code out in the input window at the bottom of the notebook and execute it, it does get evaluated as I expect.

I imagine there is work we could do here to hook everything up to the REPL hosted by the extension, but I think a better long term approach is to build a real Swift Jupyter kernel with something like [xeus](https://github.com/jupyter-xeus/xeus), likely as its own standalone repo, and then load in or depend on it in the VS Code extension. This would make it a "first class" experience that would be portable to other Jupyter environments. The Swift REPL could still power the kernel as you have it designed here.

kopyl:
How can I add this to my VSCode?

kopyl (approved):

kopyl:
Here is how I just installed it.
Not sure this is going to work for you on your system, but at least might be a good starting point and also if you google "How to install an extension for VSCode from GitHub", this comment probably might be of some use for you.

First, you have to pull this repository and switch to this PR's branch. [Here is how you can do it](https://stackoverflow.com/questions/14947789/github-clone-from-pull-request):
1. Clone the repo:
```
git clone https://github.com/swiftlang/vscode-swift.git
```
2. Feth the branch of this PR:
```
git fetch origin pull/1016/head:jupyter
```
3. Checkout to it:
```
git checkout jupyter
```

Then make sure you have NPM and Node installed on your computer. If not, Google how to do it. To verify it's installed, run `npm` command in your terminal.

Install vsce, which is a command line tool for working with VSCode:
```
npm install -g vsce
```

Then go to the root directory of this extension in the terminal and type
```
npm i
```
This is going to install all necessary packages for the extension build to be possible.

Then if I got it correctly, you have to build it. Here is how you do it:
```
npm run vscode:prepublish
```

Then you have to package it into a VSCode extension (finally):
```
vsce package --pre-release
```
After it's packaged, it will output a directory with the extension. Now all you need to install it in the editor is to run this command, replacing the path to the extension you got from the previous command (I'd make sure it's not installed in VSCode before running the following command):
```
code --install-extension vscode-swift/swift-lang-1.10.7.vsix
```

kopyl:
But the problem is that I don't see Swift as a language for the kernel :(

So this is pretty much useless.

image

compnerd:
> But the problem is that I don't see Swift as a language for the kernel :(
> 
> So this is pretty much useless.

Its been a while since I played with this, but I don't remember having to select a language. Can you refer to the Swift Forums post about this? This required experimental features in VSCode to enabled before it was usable.

kopyl:
> > But the problem is that I don't see Swift as a language for the kernel :(
> > So this is pretty much useless.
> 
> Its been a while since I played with this, but I don't remember having to select a language. Can you refer to the Swift Forums post about this? This required experimental features in VSCode to enabled before it was usable.

@compnerd could you please record a screen video of installing and using jupyter notebook with Swift in VSCode?

compnerd:
> @compnerd could you please record a screen video of installing and using jupyter notebook with Swift in VSCode?

This was already shared on the forums at https://forums.swift.org/t/vscode-jupyter-notebook-based-repl/73996. I don't have the development environment setup currently and have a number of changes that I am working on atm, so I cannot easily set this up again right now unless we have a guaranteed path to getting this merged.

PR #1019, opened 2024-08-21
Fix peeked editor closing without reopening with new contents when triggered again at the same position in the same file

lokesh-tr:
While expanding nested macros, the peeked window is supposed to appear at the same position as the previous one. The "editor.action.peekLocations" command causes the previous peeked editor to close without reopening with the new contents if invoked at the same position in the same file.

This fixes the issue by having a workaround in which we open a dummy peeked editor so that the previous peeked editor is closed, followed by opening the actual peeked editor so that, we have the peeked editor to always display at the same position in the same file while expanding a macro.

This issue is not macro expansion specific and exists in general.

Semantic Functionality and nested macro expansions PR in sourcekit-lsp where the issue was faced initially: https://github.com/swiftlang/sourcekit-lsp/pull/1634 + (https://github.com/swiftlang/vscode-swift/pull/1017 or https://github.com/swiftlang/sourcekit-lsp/pull/1636)

------

### DIFF
--- a/src/sourcekit-lsp/peekDocuments.ts
+++ b/src/sourcekit-lsp/peekDocuments.ts
@@ -16,26 +16,87 @@ import * as vscode from "vscode";
 import * as langclient from "vscode-languageclient/node";
 import { PeekDocumentsParams, PeekDocumentsRequest } from "./lspExtensions";
 
+/**
+ * Opens a peeked editor in `uri` at `position` having contents from `locations`.
+ *
+ * **NOTE**:
+ * - If the `uri` is not open in the editor, this opens the `uri` in the editor and then opens a peeked editor.
+ * - This closes any previously displayed peeked editor in `uri` and then, reopens a peeked editor in `uri` at
+ *   the given `position` with contents from the new `locations`.
+ *
+ * @param uri The uri of the file in which a peeked editor is to be opened
+ * @param position The position in the file in which a peeked editor is to be opened
+ * @param locations The locations of the contents which has to be displayed by the peeked editor
+ */
+async function openPeekedEditorIn(
+    uri: vscode.Uri,
+    position: vscode.Position,
+    locations: vscode.Location[]
+) {
+    // #### NOTE - Undocumented behaviour of invoking VS Code's built-in "editor.action.peekLocations" command:
+    // 1. If the `uri` is not open in the editor, it opens the `uri` in the editor and then opens a peeked editor.
+    // 2. It always closes the previous peeked editor (If any)
+    // 3. And after closing, It opens a new peeked editor having the contents of `locations` in `uri` **if and only
+    //    if** the previous peeked editor was displayed at a *different* `position` in `uri`.
+    // 4. If it happens to be that the previous peeked editor was displayed at the *same* `position` in `uri`, then it
+    //    doesn't open the peeked editor window having the contents of new `locations` at all.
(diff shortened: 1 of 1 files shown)


src/sourcekit-lsp/peekDocuments.ts
```diff
@@ -16,27 +16,82 @@ import * as vscode from "vscode";
...
+ * This function exists as a **workaround** for not having any API available to close the previously displayed peeked
+ * editor at `uri`
+ *
+ * **NOTE**: This function call must be immediately followed up by a call to `openPeekedEditorIn` at the same `uri`
+ * inorder to ensure that the dummy peeked editor isn't displayed to the user.
+ *
+ * @param uri The file in which the actual peeked editor is meant to be shown
+ */
+async function openDummyPeekedEditorIn(uri: vscode.Uri) {
+    // Since there isn't any API available to close the previous peeked editor, as a workaround, we open a dummy
```
plemarquand, line 58: This comment should expand on why its necessary to close the previous editor like the commit message does, explaining that this is to work around opening a nested peek editor at the same position as an old one.

src/sourcekit-lsp/peekDocuments.ts
```diff
@@ -16,27 +16,82 @@ import * as vscode from "vscode";
...
+    );
+}
+
+/**
+ * This function opens at dummy peeked editor in `uri` so that the previously displayed peeked editor gets closed.
+ *
+ * This function exists as a **workaround** for not having any API available to close the previously displayed peeked
+ * editor at `uri`
+ *
+ * **NOTE**: This function call must be immediately followed up by a call to `openPeekedEditorIn` at the same `uri`
```
plemarquand, line 52: This note indicates to me that this functionality should be moved in to `openPeekedEditorIn` so its impossible to do the wrong thing.

plemarquand (commented):
Nice fix! I only have a few comments about internal structure and documentation.

adam-fowler (approved):

lokesh-tr:
@ahoppen I don't see that flicker up since its fast, it doesn't scroll but jumps to 1:1 and then comes back to the macro expansion. People will notice something happened before the macro expansion is shown, but they won't notice what exactly happened since its fast.

And, good catch with the issue on 1:1, I am aware of it and I was writing a fix for that. I do agree that its impossible to happen in most cases, but a preventive measure is still good.

lokesh-tr:
@plemarquand I have updated it as you suggested. I am not sure if the explanation is too big, but I felt it was necessary. Please have a look and let me know if you have any other concerns.

ahoppen:
One thought that could allow us to avoid any scrolling: How about when we want to show a peek window at `line:column`, we open the fake peek on `line:(column-1)` or `line:1` if `column == 0`. That way we don’t actually jump to any different location.

lokesh-tr:
I got the same thought few hours back, got time now... will do that now.

ahoppen:
Changing the line might still result in a slight scroll if the expanded line is right at the start of the view port, I would imagine. If it’s possible to adjust the column, I would prefer that. If not, let’s go with your line proposal, it’s probably good enough.

lokesh-tr:
I went with column one as you suggested and it works. 👍🏻

lokesh-tr:
Ready to be merged, i guess?

PR #1020, opened 2024-08-21
Add swift.testBuildArguments Setting

plemarquand:
Adds a new setting that allows users to specify extra arguments when running `swift test` and `swift build --build-tests` during a test run from inside VS Code.

For example users could set `swift.testBuildArguments: ["-Xswiftc", "-DTEST"]` to enable code guarded by an `#if TEST` compiler directive.

This patch required reworking how we do a build all before starting a debug session since using the premade Build All tasks do not allow for dynamically customizing arguments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -226,8 +226,7 @@
           "swift.path": {
             "type": "string",
             "default": "",
-            "markdownDescription": "Override the default path of the folder containing the Swift executables. The default is to look in the `PATH` environment variable. This path is also used to search for other executables used by the extension like `sourcekit-lsp` and `lldb`.",
-            "order": 1
+            "markdownDescription": "Override the default path of the folder containing the Swift executables. The default is to look in the `PATH` environment variable. This path is also used to search for other executables used by the extension like `sourcekit-lsp` and `lldb`."
           },
           "swift.buildArguments": {
             "type": "array",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -220,17 +220,24 @@
...
             "type": "array",
             "default": [],
             "items": {
               "type": "string"
             },
-            "markdownDescription": "Additional arguments to pass to `swift build`. Keys and values should be provided as individual entries in the list. If you have created a copy of the build task in `tasks.json` then these build arguments will not be propogated to that task.",
-            "order": 2
+            "markdownDescription": "Additional arguments to pass to `swift build`. Keys and values should be provided as individual entries in the list. If you have created a copy of the build task in `tasks.json` then these build arguments will not be propogated to that task."
+          },
+          "swift.testBuildArguments": {
```
award999, line 233: So if `buildArguments` is always added maybe we call `testBuildArguments` something different. At first I thought it was in lieu of `buildArguments` which may cause some confusion. Maybe `additionalTestArguments`? It may then be worth cleaning up the markdown description for `buildArguments`, since it does come into play for more than just `swift build` commands, really it's any command that may build an artifact like `swift package describe`

plemarquand: Good call, I've changed this.

award999 (approved):

award999:
Verified with d6c91dd379abd6a402dddb3503e35fd328c6ce21

Issue #1013, opened 2024-08-16
Use lldb-dap by default when using Swift 6


plemarquand:
Currently vscode-swift uses the CodeLLDB extension to facilitate debugging with all swift versions. In Swift 6 `lldb-dap` started shipping in the toolchain. `lldb-dap` can be used as a replacement, removing the need to use a separate extension to debug.

There is currently a setting to start using `lldb-dap` available in the settings (`swift.debugger.useDebugAdapterFromToolchain`). I propose we use `lldb-dap` by default if the user is using Swift 6, deprecate the existing setting, and add a new one to explicitly enable using CodeLLDB in Swift 6 if users want the current behaviour.

If the user is using a Swift 5.x version the setting would have no effect. Ideally we could hide it if the user is using Swift 5.x but I dont know of a VS Code API to dynamically hide settings.

dschaefer2:
The defaults for these settings should just do the right thing: use lldb-dap if they have it or else rely on CodeLLDB.

Also note that CodeLLDB doesn't support Windows ARM so you have to use lldb-dap there.

matthewbastien:
This is going to have to wait until `lldb-dap` is close to feature parity with CodeLLDB.

dschaefer2:
Yes, boo, but I get it. Need to get there first.

felixf4xu:
> There is currently a setting to start using `lldb-dap` available in the settings

I would like to know which setting in `vscode-swift` we can set to enable `lldb-dap`?

I managed to use lldb-dap once, but in that case, I need to install another vscode addon named **LLDB DAP** from llvm.org, then changed the setting of the `lldb-dap` addon to point to swift's `lldb-dap.exe`. I also need to use `"type": "lldb-dap"` in `launch.json`. But in the whole process, it's like vscode -> lldb dap addon -> swift's lldb-dap, `vscode-swift` addon is not involved at all.

plemarquand:
@felixf4xu Its not the most clearly named, but its the VS Code setting `"swift.debugger.useDebugAdapterFromToolchain": true`.

With that setting you shouldn't need the CodeLLDB or LLDB DAP extensions to debug swift code; the extension will use the `lldb-dap` included in the Swift toolchain.

matthewbastien:
Fixed by #1406

Issue #958, opened 2024-07-08
launch configurations change / to \ on Windows


dschaefer2:
I created a project on my Mac and then checked it out on my Windows machine. VSCode-Swift changed the paths in the launch configurations to be Windows paths. Going back and forth between the platforms is going to be painful.

e.g.
```
    "program": "${workspaceFolder:doug-proj}\\.build\\debug\\doug-proj",
```

Windows does understand the forward slashes.

plemarquand:
Verified in be8372e8cd00a036bbaa5ee90c5b22c34ae37665

PR #1024, opened 2024-08-22
Hide underlying debug adapter from the user

matthewbastien:
Users should not have to know which debug adapter is being used (CodeLLDB vs `lldb-dap`). The Swift extension will now always create a `swift-lldb` launch config for each executable that it finds in the workspace. When launched these configurations will either delegate to CodeLLDB or launch `lldb-dap` depending on which Swift toolchain version is currently being used. I've added a setting to force the usage of CodeLLDB when desired.

Issue: #1013 

Additionally, the `program` path in the launch configuration will always use POSIX style paths regardless of the platform. This allows the launch.json to be shared between users of different platforms.

Issue: #958


adam-fowler:
Does this still give the user the option to use CodeLLDB when running Swift 6. 
What are the test plans for lldb-dap vs CodeLLDB?

matthewbastien:
> Does this still give the user the option to use CodeLLDB when running Swift 6.

Yup! I renamed the old setting that governed this to `swift.debugger.useCodeLLDB` which, when enabled, will force the extension to always use CodeLLDB regardless of the Swift toolchain version. The behaviour is largely the same as it used to be, but the generated launch config will always be of the type `swift-lldb`. This has the added benefit that we no longer need to watch for settings changes to update the generated launch configs.

> What are the test plans for lldb-dap vs CodeLLDB?

Thanks for reminding me: I'll create a new issue to add tests for the debug adapter. A simple UI test to ensure that the debugger launches and hits breakpoints for each supported swift version should suffice (I'm not sure there's a good integration test for this and anything more than a simple UI test would be overkill and brittle in my experience). I'd also like to add unit tests for the creation of the launch config, but mocking the WorkspaceContext/FolderContext is too difficult at the moment. There'll probably need to be some changes to how the code is structured (potentially by adding DI) to make this easier. The overall testing plan for the extension right now is to add builds for macOS and Windows to cover all platforms.

award999:
> I'm not sure there's a good integration test for this and anything more than a simple UI test would be overkill and brittle in my experience

https://www.npmjs.com/package/@vscode/debugadapter-testsupport may be an option if we connected on a port instead of starting it up https://github.com/microsoft/vscode-debugadapter-node/blob/main/testSupport/src/debugClient.ts#L88. IDK how feasible though

src/debugger/debugAdapter.ts
```diff
@@ -24,18 +24,18 @@ import { SwiftToolchain } from "../toolchain/toolchain";
...
-            : "lldb";
+        return "swift-lldb";
     }
 
     /** Return debug adapter for toolchain */
-    public static getDebugAdapter(swiftVersion: Version): "lldb-vscode" | "lldb-dap" {
-        return swiftVersion.isLessThan(new Version(6, 0, 0)) ? "lldb-vscode" : "lldb-dap";
+    public static getDebugAdapterType(swiftVersion: Version): "lldb-vscode" | "lldb-dap" {
+        return configuration.debugger.useCodeLLDB
+            ? "lldb-vscode"
```
award999, line 35: question: does CodeLLDB even allow specifying a debug adapter?

package.json
```diff
@@ -1276,4 +1276,4 @@
     "vscode-languageclient": "^9.0.1",
     "xml2js": "^0.6.2"
   }
-}
```
award999, line 1279: nit: each PR keeps changing this. Think there is an NPMRC setting to set consistent line ending

src/debugger/debugAdapter.ts
award999: suggestion: assuming it is to make other code work below then maybe we should just adjust that logic

award999 (commented):
looking great

package.json
```diff
@@ -550,10 +550,10 @@
       {
         "title": "Debugger",
         "properties": {
-          "swift.debugger.useDebugAdapterFromToolchain": {
```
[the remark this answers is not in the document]
award999: I was thinking the same thing too but this only applies to Swift 6 right? So maybe not widely used. I guess we could use in `configuration.debugger.useCodeLLDB` as a fallback

src/debugger/debugAdapter.ts
matthewbastien: No, CodeLLDB does not allow specifying a debug adapter. This is used in other places to determine which debug adapter we're using.

package.json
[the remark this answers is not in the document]
matthewbastien: I've reverted this back to use the old setting so that we don't have to deprecate anything.

package.json
matthewbastien: Will create a separate pull request to enforce this behaviour.

plemarquand:

matthewbastien:
I've reverted back to using CodeLLDB by default for this patch until we can verify that lldb-dap has feature parity with CodeLLDB.

PR #1025, opened 2024-08-22
Associate comment details on #expect w/ test issue

plemarquand:
If a swift-testing `#expect` macro has preceding comments they are sent along with the expectation failure text as entries in the EventMessage array. Pull them out of the array so they don't appear as their own separate issues and append them to the actual issue text.

Also fixes up some of the issue message colorization to be more consistent with how swift-testing behaves on the command line.


matthewbastien:
Verified with d6c91dd

PR #1026, opened 2024-08-22
Work around `Uri` round-tripping issue in VS Code for `sourcekit-lsp` scheme

ahoppen:
`vscode.uri` fails to round-trip URIs that have both a `=` and `%3D` (percent-encoded `=`) in the query component.

```ts
vscode.Uri.parse("scheme://host?outer=inner%3Dvalue").toString() -> 'scheme://host?outer%3Dinner%3Dvalue'
vscode.Uri.parse("scheme://host?outer=inner%3Dvalue").toString(/*skipEncoding*/ true) -> 'scheme://host?outer=inner=value'
```

The SourceKit-LSP scheme relies heavily on encoding options in the query parameters, eg. for Swift macro expansions and the values of those query parameters might contain percent-encoded `=` signs.

To work around the round-trip issue, don't percent-decode any components of the URI and then use `toString(/*skipEncoding*/ true)` to serialize it into a string again.

Technically, we should percent-decode authority and path but if we do that, we also need to percent-encode it in `code2Protocol`, which means that we can't use `vscode.Uri.toString` anymore and would need to write our own encoding code. Since SourceKit-LSP currently doesn't use any characters in authority and path that need percent encoding for the `sourcekit-lsp:` scheme, we can just not do any decoding here.

### DIFF
--- a/src/sourcekit-lsp/LanguageClientManager.ts
+++ b/src/sourcekit-lsp/LanguageClientManager.ts
@@ -30,6 +30,7 @@ import { LSPLogger, LSPOutputChannel } from "./LSPOutputChannel";
 import { SwiftOutputChannel } from "../ui/SwiftOutputChannel";
 import { promptForDiagnostics } from "../commands/captureDiagnostics";
 import { activateGetReferenceDocument } from "./getReferenceDocument";
+import { uriConverters } from "./uriConverters";
 
 interface SourceKitLogMessageParams extends langclient.LogMessageParams {
     logName?: string;
@@ -574,6 +575,7 @@ export class LanguageClientManager {
                     };
                 })(),
             },
+            uriConverters,
             errorHandler,
             // Avoid attempting to reinitialize multiple times. If we fail to initialize
             // we aren't doing anything different the second time and so will fail again.
--- a/src/sourcekit-lsp/uriConverters.ts
+++ b/src/sourcekit-lsp/uriConverters.ts
@@ -1,0 +1,125 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2024 the VS Code Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of VS Code Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import * as vscode from "vscode";
+
+export const uriConverters = {
+    protocol2Code: (value: string): vscode.Uri => {
+        if (!value.startsWith("sourcekit-lsp:")) {
+            // Use the default implementation for all schemes other than sourcekit-lsp, as defined here:
+            // https://github.com/microsoft/vscode-languageserver-node/blob/14ddabfc22187b698e83ecde072247aa40727308/client/src/common/protocolConverter.ts#L286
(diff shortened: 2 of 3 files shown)


test/unit-tests/sourcekit-lsp/uriConverters.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+/// Check that decoding the given URI string and re-encoding it results in the original string and that the decoded Uri
+/// does not cause any assertion failures in `verifyUri`.
+function checkUri(input: string, verifyUri: (uri: vscode.Uri) => void) {
+    const uri = uriConverters.protocol2Code(input);
+    verifyUri(uri);
+    assert.equal(uriConverters.code2Protocol(uri), input);
+}
+
+suite.only("uriConverters Suite", () => {
```
plemarquand, line 27: Need to remove the .only

src/sourcekit-lsp/uriConverters.ts
```diff
@@ -0,0 +1,63 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import * as vscode from "vscode";
+
+// Regular expression to decompose a URI, copied from
+// https://github.com/microsoft/vscode/blob/1647dde06cd9b610e2576f09f68101c31c00e574/src/vs/base/common/uri.ts#L80
+const _decodeUriRegexp = /^(([^:/?#]+?):)?(\/\/([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?/;
```
plemarquand, line 19: This works, but is it possible to use Node’s [URL](https://nodejs.org/api/url.html#new-urlinput-base) implementation here over the regex? All the sourcekit lsp Uris should be compatible with the initializer.

You’d have to build the authority out of the urls host and optional port.

plemarquand (changes requested):

lokesh-tr:
@ahoppen Thanks for providing this workaround. Did you test if this works without the sourcekit-lsp's workaround which you made earlier?

src/sourcekit-lsp/uriConverters.ts
adam-fowler: Internally VSCode uses decodeURIComponent

ahoppen: Good suggestion with using Node’s URL type to do the parsing. I updated the PR to use Node’s URL parser.

ahoppen:
> @ahoppen Thanks for providing this workaround. Did you test if this works without the sourcekit-lsp's workaround which you made earlier?

Yes, I did and nested macro expansions worked.

ahoppen:
@plemarquand Could you take another look at my new implementation that’s based on Node’s `URL` type?

src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -574,6 +575,7 @@ export class LanguageClientManager {
                     };
                 })(),
             },
+            uriConverters: uriConverters,
```
plemarquand, line 578: nit: can use property shorthand syntax, shortening to just `uriConverters`

plemarquand (approved):
@ahoppen Sorry I didn't come back to this one. LGTM, though I didn't realize quite how much vscode.URI and URL diverged.

ahoppen:
Yeah, I think the key takeaway here is that `vscode.Uri` is actually quite a bad URI type. But it’s the one we have to work with, so 🤷🏽

Issue #1021, opened 2024-08-21
Coverage report should include all files under src


michael-weng:
**Is your feature request related to a problem? Please describe.**
Right now files that are not executed by test will not be included in the coverage, inflating the actual coverage number, we should address this.

**Describe the solution you'd like**
Include files with 0% coverage in coverage report.

award999:
Done

PR #1032, opened 2024-08-27
Fix coverage report

award999:
Issue was the dist/extension.js file was what vscode-test was running as the extension as it was what was specified as the entry point in the package.json. So The globalWorkspaceContextPromise helper we use in a lot of tests to make sure the extension is activated, returns an instance of the WorkspaceContext. Any behaviour stimulated through that context is missed from the coverage report, and also any breakpoint hits are ignored. So for example the BackgroundCompilation instance that is created on extension activation was left out of the coverage report.

To fix, only bundle to a single extension.js file when packaging a vsix. When debugging or running tests we transpile all the files individually. The dist/extension.js file is still used when activating, but sourcemap now works so the coverage report is accurate.

Issue: #1021

### DIFF
--- a/.vscode-test.js
+++ b/.vscode-test.js
@@ -33,12 +33,11 @@ module.exports = defineConfig({
             mocha: {
                 ui: "tdd",
                 color: true,
-                // so doesn't timeout when breakpoint is hit
-                timeout: isDebugRun ? Number.MAX_SAFE_INTEGER : 3000,
                 timeout,
                 forbidOnly: isCIBuild,
                 grep: isFastTestRun ? "@slow" : undefined,
                 invert: isFastTestRun,
+                slow: 10000
             },
             installExtensions: ["vadimcn.vscode-lldb"],
             reuseMachineInstall: !isCIBuild,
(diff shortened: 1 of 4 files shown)


.vscode-test.js
```diff
@@ -50,17 +49,17 @@ module.exports = defineConfig({
...
             },
             reuseMachineInstall: !isCIBuild,
         },
         // you can specify additional test configurations, too
     ],
     coverage: {
         includeAll: true,
-        include: ["**/src/**"],
-        reporter: ["text", "html"],
+        exclude: ["**/test/**"],
```
michael-weng, line 62: This now pulls in all the node_modules files, intentional?

award999: maybe need to clear old "out" and "dist" directories? The vscode-test-cli tool automates adding node_modules

michael-weng: Yeah I think you mentioned this before, seems to be reported already:
https://github.com/microsoft/vscode-test-cli/issues/40
https://github.com/microsoft/vscode-test-cli/issues/37

award999: verified not in coverage report

plemarquand (approved):

PR #1034, opened 2024-08-27
Mark XCTests as passed without a second pass over the output

plemarquand:
The lastTestItem was not being set right away when a test succeeded because tests were being marked as passed in a second pass of the XCTest output after a run completed.

Going by the removed comment this second pass was added to disambiguate passing tests with the same suite/test name across multiple targets.

However, the test target is present in the `testName` used to look up a test's index. For instance, it may look like
`TestTargetA/Suite/testFoo` and so would not be identical to `TestTargetB/Suite/testFoo`, and so we get the correct `testIndex` for either test.

### DIFF
--- a/src/TestExplorer/TestParsers/XCTestOutputParser.ts
+++ b/src/TestExplorer/TestParsers/XCTestOutputParser.ts
@@ -195,7 +195,7 @@ export class XCTestOutputParser implements IXCTestOutputParser {
                         this.failTest(testIndex, { duration }, runState);
                         break;
                     case TestCompletionState.passed:
-                        // Tests are passed in the second pass below.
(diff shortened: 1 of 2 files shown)


plemarquand:
@adam-fowler The comment on the [removed code](https://github.com/swiftlang/vscode-swift/pull/1034/files#diff-ef7582afb64cc4118a63b15b369539b551a402b897443bb2466ef343e5407ebeL257) claims its to work around identically named suites/tests in different test targets (on Linux/Windows). However because the `testName` can be identical across targets I was still seeing incorrect results with the code in `main` today.

I created https://github.com/swiftlang/vscode-swift/issues/1036 to track this issue separately.

award999 (approved):

PR #1046, opened 2024-08-30
Route swift-testing output to /dev/stdout, CONOUT$

plemarquand:
Currently we are merging two streams of information to produce swift-testing test output, the JSON event stream written to a named pipe and the stream of data from stdout. This captures both testing events and user outputs produced by things like `print()` statements.

However, this approach interleaves the two sources in an arbitrary order. Print statements produced during a test run are typically read after events on the named pipe, which makes it difficult to tell what prints belong to what test.

In order to enforce the correct order we can configure swift-testing to write its events to `/dev/stdout` (or `CONOUT$` on Windows). Swift-testing aquires a lock to the `fd` it is writing events and output to, so the order of events and prints is correctly enforced.

This also simplifies a lot of the code that worked with the JSON event stream as we can now rely solely on the task's stdout/stderr output without the need for named pipes.

When parsing stdout we try and parse a JSON event that matches the swift-testing event schema, and if we fail we print the line directly. If it is parseable, we process it as a swift-testing JSON event and omit it from the test run output.


plemarquand:
Output on Windows will appear incorrectly ordered until https://github.com/swiftlang/swift-testing/pull/654 lands, which will adjust the locking behaviour of swift-testing w/ CONOUT$ on Windows to match how it works on Darwin/Linux.

plemarquand:
This addresses https://github.com/swiftlang/vscode-swift/issues/1054, however currently this approach doesn't account for the line wrapping that occurs when you write long JSON events to /dev/stdout.

plemarquand:
This is supplanted by https://github.com/swiftlang/vscode-swift/pull/1058 which is a simpler approach with less edge cases to contend with

Issue #1047, opened 2024-08-30
Run Multiple Times / Until Failure commands showing up for non-Swift tests


award999:
**Describe the bug**
You can see these new context menu commands for test cases that aren't swift tests

**To Reproduce**
1. Open a workspace that has no Swift source (I had a Go workspace open)
2. Go to the test explorer and right-click on a test case
3. See commands
![Screenshot 2024-08-29 at 3 27 23 PM](https://github.com/user-attachments/assets/54f80479-ce0a-464c-b521-94c222597f39)

NOTE When clicking the actually Swift extension activates and tries to run but fails obviously

**Expected behavior**
Hide these commands for non-Swift tests

**Environment**
All environments

**Additional context**
N/A

award999:
Verified with be8372e

PR #1048, opened 2024-08-30
Show Swift test explorer context menu items only for Swift tests

plemarquand:
Keep a flattened list of swift Test Explorer test IDs in the context and use a `when` clause on the context menu items to only show them on Swift tests in the Text Explorer.

Issue: #1047


award999 (approved):
LGTM

PR #1053, opened 2024-09-05
Create minimal list of test args for `swift test`

plemarquand:
When running tests in VS Code a list of arguments is produced that is passed to `swift test`. This list is determined by what the user has selected in the test explorer. If the user has several tests/suites selected the list was tests was paired down to a minimal list of tests/suites that covers the users selection. This paring down process stopped at the suite level, so if the user ran an entire test target then all suites in the target were included in the list. This works but only up to a point in larger projects. For instance running all the tests in swift-format generates a list of test suites so long that it breaks on Windows.

Improve the pairing logic to work down to the test target level. If a test target contains both XCTest and swift-testing tests it ends up in both the `xcTestArgs` and `swiftTestingArgs` and the two separate runs of `swift test` with the test library enablement/disablement flags takes care of excluding one type from the other.


plemarquand:

matthewbastien (approved):

Issue #1054, opened 2024-09-05
Print statements in swift-testing tests appear desynched from test result information


plemarquand:
**Describe the bug**
Currently the extension merges two streams of information to produce swift-testing test output: the JSON event stream written to a named pipe and the stream of data from stdout. This captures both testing events and user outputs produced by things like print() statements.

However, this approach interleaves the two sources in an arbitrary order. Print statements produced during a test run are typically read after events on the named pipe, which makes it difficult to tell what prints belong to what test. There is no ordering guarantee as there is latency in reading from the named pipe, and from stdout/err.

The ideal solution would be to have user prints emitted in the JSON event stream as their own event type, ensuring they appear between testCaseStart/End events. Barring that perhaps something is possible where both the events and stdout are piped to the same fd.

award999:
Verified with be8372e

PR #1055, opened 2024-09-05
Address flakey swift-testing test

plemarquand:
The `topLevelTestPassing()` test asserts that a print statement is captured during a test. However because of #1054 it is possible that this print is saved to the test run after the swift-testing test end JSON event has been recieved. Until this issue is addressed, add a 0.1 second delay to ensure the print is captured before the test end event.


award999 (approved):

award999:

Issue #1023, opened 2024-08-22
Add test suite for debug components


michael-weng:
**Is your feature request related to a problem? Please describe.**
This component is currently sitting at around 25% line coverage. We should bump that number up.

plemarquand:
Tests passing, verified

PR #1056, opened 2024-09-05
Add docstring and unit tests for attachDebugger

michael-weng:
We should hit 100% line coverage for this module now.

Also added example for same test written with sinon vs ts-mockito
for discussion, in this particular scenario ts-mockito was not able
to properly resolve to the fake object.

Issue: https://github.com/swiftlang/vscode-swift/issues/1023


test/unit-tests/debugger/attachDebugger.test.ts
```diff
@@ -0,0 +1,97 @@
...
+import { WorkspaceContext } from "../../../src/WorkspaceContext";
+
+suite("attachDebugger Unit Test Suite with Thenable Capture", () => {
+    const lldbSpy = spy(lldb);
+    const mockContext = mock(WorkspaceContext);
+    const windowMock = mockNamespace(vscode, "window");
+
+    setup(() => {});
+
+    test.skip("should call startDebugging with correct debugConfig and capture the return Thenable", async () => {
```
plemarquand, line 16: Deliberate `skip`?

test/unit-tests/debugger/attachDebugger.test.ts
```diff
@@ -0,0 +1,97 @@
...
+        // Verify startDebugging was called once
+        verify(debugSpy.startDebugging(undefined, anything())).once();
+
+        // Now, capture the returned Thenable and assert its behavior
+        const capturedThenable = capture(debugSpy.startDebugging).last()[0];
+        assert.strictEqual(capturedThenable, true, "Thenable should resolve to true");
+    });
+});
+
+import * as sinon from "sinon";
```
plemarquand, line 46: Can we accomplish what we need to with just ts-mockito? I'm hesitant to add multiple mocking libraries, especially when mocking should used as a be a last resort.

test/unit-tests/debugger/attachDebugger.test.ts
michael-weng: Yup, explained below.

test/unit-tests/debugger/attachDebugger.test.ts
michael-weng: Same wise, I was meant to have this patch up as a draft for discussion. For this test case, ts-mockito was not able to properly resolve the mock object.

award999: From what I have seen there should be a way to Proxy anything. It would be more of a discussion of we should use Sinon, getting rid of ts-mockito if necessary. This may fall into a case where we either test critical functionality at the integration level if there is no way to mock, or don't test at all if there is no way. We don't need to hit 100% coverage if there is too many hoops to jump through to get the source code under test

michael-weng: Figured out a way to do it, just have to work around the overloaded signature and make sure things match.

plemarquand (approved):

PR #1060, opened 2024-09-08
fix missing diagnostics from clangd

### LINKED ISSUE #1028: VS Code does not show diagnostics for C files
When opening a C file using VS Code with SourceKit-LSP as the language server, no diagnostics are shown. This seems to be because the diagnostic’s source is `clang` and that’s not in `DiagnosticsManager.sourcekit`. Since we forward the diagnostic’s source from clangd here, I wonder whether it would be possible to generalize the implementation of `DiagnosticsManager` to consider everything that’s not from `swiftc` as a SourceKit diagnostics.

I think build errors from C files in a SwiftPM package also don’t show up in the Problems view.

AppAppWorks:
fixes #1028

### DIFF
--- a/src/DiagnosticsManager.ts
+++ b/src/DiagnosticsManager.ts
@@ -26,8 +26,21 @@ interface ParsedDiagnostic {
 
 type DiagnosticsMap = Map<string, vscode.Diagnostic[]>;
 
+type SourcePredicate = (source: string) => boolean;
+
+type DiagnosticPredicate = (diagnostic: vscode.Diagnostic) => boolean;
+
 const isEqual = (d1: vscode.Diagnostic, d2: vscode.Diagnostic) =>
     d1.range.start.isEqual(d2.range.start) && d1.message === d2.message;
+
+const isSource = (diagnostic: vscode.Diagnostic, sourcesPredicate: SourcePredicate) =>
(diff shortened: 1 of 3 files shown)


award999:
@AppAppWorks thanks for the PR, I already had something on the go that I just put up which also fixes some of the issues related to the deduplication of diagnostics between the compiler and SourceKit-LSP. Feel free to leave feedback on #1062

AppAppWorks:
> @AppAppWorks to be clear though we'd love to see you code clean up patch on top :)

It's nice of you, but what should I do now? I'm not terribly familiar with GitHub workflow ;)

award999:
So I think the main difference in my solution is this section https://github.com/swiftlang/vscode-swift/pull/1062/files#diff-0e284d36e0b7ec176d59373496202e98eb0a2656e44ec437bd3f85a902e21bcdR108-R385. We use deduplication when merging in diagnostics to make sure we do not should duplicate issues from the compiler and the LSP. Additionally when the LSP indicates that an issue has been fixed we remove any duplicate compiler issue.

So we can work together on this one, if you want to apply that logic to your solution. I'm having some trouble with one of the integration tests in CI, so I can then put these test cases up after

PR #1067, opened 2024-09-12
Refresh the test explorer when the package changes

plemarquand:
Currently if the user adds or removes a test target from the package manifest this change is not reflected in the test list of the test explorer.

When the package manifest is updated, refresh the test explorer by making a workspace/test's LSP request.


matthewbastien (approved):

matthewbastien:
Verified with 737ad69

Issue #1078, opened 2024-09-18
`npm run package` is including unnecessary files


adam-fowler:
The package scripts now lists all the files being added to the vsix. Not all of these need adding to the extension eg the docs folder
```
swift-lang-1.11.1.vsix
├─ [Content_Types].xml
├─ extension.vsixmanifest
└─ extension/
   ├─ .mailmap
   ├─ .nvmrc
   ├─ .vscode-test.js
   ├─ CHANGELOG.md
   ├─ CONTRIBUTORS.txt
   ├─ LICENSE.txt
   ├─ NOTICE.txt
   ├─ README.md
   ├─ icon.png
   ├─ package.json
   ├─ .devcontainer/
   │  ├─ Dockerfile
   │  └─ devcontainer.json
   ├─ assets/
   │  └─ icon-font.woff
   ├─ dist/
   │  └─ extension.js
   ├─ docs/
   │  ├─ remote-dev.md
   │  ├─ settings.md
   │  ├─ test-coverage.md
   │  ├─ contributor/ (1 file)
   │  └─ images/ (3 files)
   ├─ images/
   │  ├─ install-extension.png
   │  ├─ package-dependencies.png
   │  └─ test-explorer.png
   └─ snippets/
      ├─ swift-testing.code-snippets
      ├─ swift.code-snippets
      └─ xctest.code-snippets
```

award999:
@michael-weng can you verify this?

michael-weng:
Verified:
commit 737ad6941ad7d147f2b3f2f05a5399c618bfa9ad

Files included in the VSIX properly reflects change of PR:
swift-lang-1.11.2.vsix
├─ [Content_Types].xml
├─ extension.vsixmanifest
└─ extension/
   ├─ CHANGELOG.md
   ├─ LICENSE.txt
   ├─ NOTICE.txt
   ├─ README.md
   ├─ icon.png
   ├─ package.json
   ├─ assets/
   │  └─ icon-font.woff
   ├─ dist/
   │  └─ extension.js
   ├─ images/
   │  ├─ install-extension.png
   │  ├─ package-dependencies.png
   │  └─ test-explorer.png
   └─ snippets/
      ├─ swift-testing.code-snippets
      ├─ swift.code-snippets
      └─ xctest.code-snippets
      
Tested with Apple Swift version 6.0 (swiftlang-6.0.0.4.52 clang-1600.0.21.1.3)
build and debug of test package is working, condition:
[...]

PR #1079, opened 2024-09-18
Ignore unnecessary files when packaging extension

plemarquand:
Slim down the extension by ignoring files that don't need to be included in the .vsix

Issue: #1078


adam-fowler:
Not sure about the CHANGELOG. Don't know where VSCode gets that from

award999 (changes requested):
So we don't have to update this every time, if you checkout https://code.visualstudio.com/api/working-with-extensions/publishing-extension#using-.vscodeignore, you can ! to include something so what I've seen work well before is ignore everything (**/*) by default and use ! to explicitly only include what we want

plemarquand:
@adam-fowler I think you're right, the tooltip on the Changelog tab in the extension's marketplace page implies it comes from the CHANGELOG.md. I've included that.

@award999 I've moved over to the recommended explicit include approach.

award999 (approved):
LGTM

adam-fowler:
Don't know enough about the `.vscodeignore` but what does `**/*` do? But otherwise looks good

award999:
> Don't know enough about the `.vscodeignore` but what does `**/*` do? But otherwise looks good

Basically this approach removes everything by default, and then preceding each item with ! will add those items in. That way we don't have to add new entries to this file every time we add a root level item. The developer has to explicitly add a new ! exception for it to end up in the vsix bundle

adam-fowler (approved):

Issue #1074, opened 2024-09-17
Warning and error messages are truncated in Problems view


HeMet:
**Describe the bug**
Warning and error messages are truncated in Problems view
![swift_messages_truncated](https://github.com/user-attachments/assets/0cb36d29-5246-4215-829a-0d95c9ccadcc)

**To Reproduce**
1. Build any project with long enough warnings

**Expected behavior**
Warning messages should not be truncated.

**Environment**
- Swift version 6.0 (swift-6.0-RELEASE)
  Target: x86_64-unknown-windows-msvc
 - Visual Studio Code version: Version: 1.93.1 (user setup)
Commit: 38c31bc77e0dd6ae88a4e9cc93428cc27a56ba40
Date: 2024-09-11T17:20:05.685Z
Electron: 30.4.0
ElectronBuildId: 10073054
Chromium: 124.0.6367.243
Node.js: 20.15.1
V8: 12.4.254.20-electron.0
OS: Windows_NT x64 10.0.22631
 - vscode-swift version: v1.11.0

**Additional context**

plemarquand:
Verified with 2dad6d7f37e5f9f9547979ee7c39a71c07c73e87

PR #1082, opened 2024-09-18
Fix truncated diagnostics on Windows

award999:
* We never want to resize the terminal regardless of if deployed or debug environment
* If we are using conpty don't put large cols values as it will throw an error when task terminal opens
* When we aren't using conpty (debugging) need to set a large cols value

Issue: #1074

### DIFF
--- a/src/tasks/SwiftProcess.ts
+++ b/src/tasks/SwiftProcess.ts
@@ -94,16 +94,17 @@ export class SwiftPtyProcess implements SwiftProcess {
 
     spawn(): void {
         try {
+            const isWindows = process.platform === "win32";
             // The pty process hangs on Windows when debugging the extension if we use conpty
             // See https://github.com/microsoft/node-pty/issues/640
-            const useConpty =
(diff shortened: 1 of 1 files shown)


src/tasks/SwiftProcess.ts
```diff
@@ -136,6 +137,11 @@ export class SwiftPtyProcess implements SwiftProcess {
     }
 
     setDimensions(dimensions: vscode.TerminalDimensions): void {
+        // https://github.com/swiftlang/vscode-swift/issues/1074
+        // Causing weird truncation issues
+        if (process.platform === "win32") {
+            return;
+        }
         this.spawnedProcess?.resize(dimensions.columns, dimensions.rows);
```
plemarquand, line 145: Does this cause the terminal output to be incorrectly sized in then Terminal tab of VS Code on mac/linux?

award999: Should only affect windows

plemarquand: setDimensions is only called on Windows?

plemarquand (approved):

PR #1083, opened 2024-09-19
Fix Several Windows Debugging Issues

plemarquand:
Fixes up some issues with Windows debugging found in the latest 6.0 toolchain:
- Add Testing.dll to the PATH same as XCTest.dll
- Don't append .exe when debugging testing binaries
- Don't follow incorrectly created symlink at `.\build\debug`, point to real folder at the build triple path

### DIFF
--- a/src/debugger/debugAdapterFactory.ts
+++ b/src/debugger/debugAdapterFactory.ts
@@ -77,9 +77,14 @@ export class LLDBDebugConfigurationProvider implements vscode.DebugConfiguration
     ): Promise<vscode.DebugConfiguration> {
         launchConfig.env = this.convertEnvironmentVariables(launchConfig.env);
         // Fix the program path on Windows to include the ".exe" extension
-        if (this.platform === "win32" && path.extname(launchConfig.program) !== ".exe") {
+        if (
+            this.platform === "win32" &&
+            launchConfig.testType === undefined &&
+            path.extname(launchConfig.program) !== ".exe"
+        ) {
             launchConfig.program += ".exe";
         }
+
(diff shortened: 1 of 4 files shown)


plemarquand:
This works around this issue in SwiftPM/Foundation: https://github.com/swiftlang/swift-package-manager/issues/7965

award999 (commented):
Looks good to me but can you review @matthewbastien since you're more familiar with this code

src/debugger/debugAdapterFactory.ts
```diff
@@ -77,9 +77,14 @@ export class LLDBDebugConfigurationProvider implements vscode.DebugConfiguration
     ): Promise<vscode.DebugConfiguration> {
         launchConfig.env = this.convertEnvironmentVariables(launchConfig.env);
         // Fix the program path on Windows to include the ".exe" extension
-        if (this.platform === "win32" && path.extname(launchConfig.program) !== ".exe") {
+        if (
+            this.platform === "win32" &&
+            launchConfig.testType === undefined &&
```
matthewbastien, line 82: This works for now, but it would be nice to have a more "universal" way to determine if the `.exe` suffix needs to be added. I didn't anticipate this use case when I added this code. It could definitely use some work.

Perhaps we can have the auto-generated launch configurations use a `target` parameter instead of `program`, for example. That way the launch config can be used across all platforms while retaining the ability to target a specific executable for the tests.

I'm going to open up a separate issue from this since this needs to go in to unblock test debugging on Windows.

matthewbastien (approved):
lgtm just one minor comment that probably needs a separate investigation

PR #1091, opened 2024-09-23
Release activities for 1.11.2

award999:


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## 1.11.2 - 2024-09-23
+
+### Fixed
+
+- Fix numerous debugging issues on Windows ([#1083](https://github.com/swiftlang/vscode-swift/pull/1083))
```
matthewbastien, line 7: I would use the same wording as the PR here `Fixed several debugging issues on Windows`

matthewbastien (changes requested):
I think it makes more sense to make the URLs point to the resolved issues rather than the PRs.

CHANGELOG.md
award999: done

matthewbastien (approved):

PR #1093, opened 2024-09-24
Proposal for test improvements

matthewbastien:
### Background

The current test strategy uses `ts-mockito` for mocking and the `assert` module from NodeJS for assertions. Unfortunately, `ts-mockito` is not very well maintained and the `assert` module is lacking in functionality when compared to other assertion libraries.

### Proposal

This PR updates the tests to use [Chai](https://www.chaijs.com/) for natural language assertions and [Sinon](https://sinonjs.org/) for stubbing/mocking. Both of these libraries are very well maintained and have decent documentation. They are also the de-facto standard when using the [Mocha](https://mochajs.org/) testing framework as this project does.

To fill in the gaps left by removing `ts-mockito`, I've updated MockUtils to use these new libraries and added documentation for how to write tests after this change is merged. All tests have been updated to use Sinon over `ts-mockito`. However, there are still tests that use the `assert` module which can be converted to Chai assertions at a later point.

Any comments, questions, or concerns are welcome!


docs/contributor/writing-effective-tests.md
```diff
@@ -0,0 +1,316 @@
+# Writing Effective Tests <!-- omit in toc -->
+
+This document provides guidance to contributors on how to create effective test cases for contributions to the [VSCode Swift extension](https://github.com/swiftlang/vscode-swift) using **Mocha**, **Chai**, and **Sinon**. These tools are widely used for testing JavaScript/TypeScript code and will help ensure that your code is reliable and maintainable.
+
+A brief description of each framework can be found below:
+
+-   [**Mocha**](https://mochajs.org/): A JavaScript test framework that runs in Node.js. It provides structure for writing test suites, hooks (e.g., `setup()`, `teardown()`), and test cases (`test()`).
+-   [**Chai**](https://www.chaijs.com/): An assertion library. It allows you to express expectations for your code’s behavior using natural language.
```
award999, line 8: awesome, this will make it easier to figure out test intent than "assert"

docs/contributor/writing-effective-tests.md
```diff
@@ -0,0 +1,316 @@
...
+    - [Setting global constants](#setting-global-constants)
+    - [Mocking an entire module](#mocking-an-entire-module)
+- [Conclusion](#conclusion)
+
+## Organizing Tests
+
+Tests are currently organized into one of two directories:
+
+1. **Unit Tests** - Tests that exercise smaller units of code go in `test/unit-tests`
+2. **Integration Tests** - Tests that exercise more complicated features that require the extension to be loaded in VSCode go in `test/integration-tests`
```
award999, line 31: good call moving these

docs/contributor/writing-effective-tests.md
```diff
@@ -0,0 +1,316 @@
...
+
+There are plans to add a third set of smoke tests that will have no mocking at all and actually click through the VS Code user interface to perform actions. For more information see the [Swift for Visual Studio Code test strategy](./test-strategy.md) document.
+
+## Running the Tests
```
award999, line 35: If the point of this guide is to highlight how to run the tests, maybe move this to the contributors guide

docs/contributor/writing-effective-tests.md
```diff
@@ -0,0 +1,316 @@
...
+    suite("ReloadExtension Unit Test Suite", () => {
+        test("displays a warning message asking the user to reload the window", async () => {
+            // Test code goes here
+        });
+    });
+    ```
+
+Now comes the fun part! For unit testing we use [Chai](https://chaijs.com) for assertions and [Sinon](https://sinonjs.org/) for stubbing/mocking user interaction and observing behaviour that is otherwise difficult to observe. Many helpful utility functions exist in [MockUtils.ts](../../test/MockUtils.ts). These utility methods take care of the setup and teardown of the mocks so the developer does not need to remember to do this for each suite/test.
+
+Our test is going to want to verify that a warning message is shown to the user. For this, we'll want to mock out VSCodes global `window` module. Thankfully, MockUtils contains a `mockGlobalObject` function that will replace a global object with a mocked version. Each property is replaced with a Sinon stub during setup and restored at teardown. Keep in mind that the `mockGlobal*` functions all need to be created at the **`suite()`** level. Unexpected behavior will occur if called within a `test()` function.
```
award999, line 75: maybe make clear the strength of this comes from the test utilities handling the setup and teardown

docs/contributor/writing-effective-tests.md
```diff
@@ -0,0 +1,316 @@
...
+
+#### Mocking global interfaces
+
+```typescript
+import { expect } from "chai";
+import { mockGlobalObject } from "../MockUtils";
+import * as vscode from "vscode";
+
+suite("Mocking a Global Interface", async function () {
+    const mockedVSCodeWindow = mockGlobalObject
```
award999, line 243: missing arguments?

package.json
```diff
@@ -1288,27 +1288,38 @@
...
     "@vscode/vsce": "^2.30.0",
+    "chai": "^4.5.0",
+    "chai-as-promised": "^7.1.2",
+    "chai-subset": "^1.6.0",
     "del-cli": "^5.1.0",
     "esbuild": "^0.23.0",
     "eslint": "^8.57.0",
     "eslint-config-prettier": "^9.1.0",
     "mocha": "^10.6.0",
+    "mock-fs": "^5.2.0",
```
award999, line 1317: Looks like there is only one example of this. So not convinced that adding is necessary, at the very least we need to add this to the writing effective unit tests docs and make clear why using the other utilities is not sufficient

src/FolderContext.ts
```diff
@@ -18,7 +18,7 @@ import { LinuxMain } from "./LinuxMain";
 import { PackageWatcher } from "./PackageWatcher";
 import { SwiftPackage, Target, TargetType } from "./SwiftPackage";
 import { TestExplorer } from "./TestExplorer/TestExplorer";
-import { WorkspaceContext, FolderEvent } from "./WorkspaceContext";
+import { WorkspaceContext, FolderOperation } from "./WorkspaceContext";
```
award999, line 21: renaming this enum is outside the scope of this PR and creates a lot of noise, especially under src/, let's tackle that as a separate review

test/unit-tests/utilities/version.test.ts
```diff
@@ -12,103 +12,103 @@
...
 
     test("isGreaterThanOrEqual", () => {
-        assert.equal(new Version(5, 10, 1).isGreaterThanOrEqual(new Version(6, 0, 0)), false);
-        assert.equal(new Version(5, 9, 0).isGreaterThanOrEqual(new Version(5, 10, 0)), false);
-        assert.equal(new Version(5, 10, 0).isGreaterThanOrEqual(new Version(5, 10, 1)), false);
-        assert.equal(new Version(5, 10, 1).isGreaterThanOrEqual(new Version(5, 10, 1)), true);
-        assert.equal(new Version(5, 10, 0).isGreaterThanOrEqual(new Version(5, 9, 0)), true);
-        assert.equal(new Version(5, 10, 1).isGreaterThanOrEqual(new Version(5, 10, 0)), true);
-        assert.equal(new Version(6, 0, 0).isGreaterThanOrEqual(new Version(5, 10, 1)), true);
+        expect(new Version(5, 10, 1).isGreaterThanOrEqual(new Version(6, 0, 0))).to.equal(false);
```
award999, line 106: I'd argue to.be.false or true or undefined read more naturally to make clear why chai makes things clearer

award999 (changes requested):

award999:
This is looking awesome, it helps with a lot of the issues I've been wrestling with as I write test cases :)

package.json
michael-weng: Mocking the global module fs also works, for example:
https://github.com/michael-weng/vscode-swift/blob/9204486e4704a4544cd7a14af9817a79e971f116/test/unit-tests/debugger/lldb.test.ts

docs/contributor/writing-effective-tests.md
michael-weng, line 1: I've had the opportunity to work with the patch and write multiple tests under the new setup. I’m pleased to report that the new arrangement is ergonomic and has proven to be smooth—so far, we haven’t encountered any issues with test setup.

I also think previous pattern established by ts-mockito provided good foundation that let us know what our test pattern and setup would look like.

Also great work on the thorough documentation, this will be super helpful and would be neat to see what the community come up with.

test/MockUtils.ts
michael-weng, line 1: Praise: the doc strings here that highlights the use case and caveat are fantastic.

src/FolderContext.ts
matthewbastien: Unfortunately, it's not just a rename. I had to modify the `WorkspaceContext` to use VS Code Events so that it was testable with `MockUtils`.

test/unit-tests/utilities/version.test.ts
matthewbastien: Good point. Done.

docs/contributor/writing-effective-tests.md
matthewbastien: Good catch!

docs/contributor/writing-effective-tests.md
matthewbastien: I added more info on this in the `Mocking global modules` section.

docs/contributor/writing-effective-tests.md
matthewbastien: Yeah that makes sense. I reworded the `CONTRIBUTING.md` document to reference this document.

package.json
matthewbastien: Yeah, it's not a great example since that one can be easily replaced with `mockGlobalModule()`. I've removed `mock-fs` for now until we can find a proper use case for it.

src/FolderContext.ts
award999: `type FolderEvent = FolderOperation;`? Just feels like a lot of noise and source of merge conflicts for what's on the go

docs/contributor/writing-effective-tests.md
```diff
@@ -0,0 +1,359 @@
...
+```typescript
+import { expect } from "chai";
+import { mockFn, mockObject, MockedObject } from "../MockUtils";
+import * as vscode from "vscode";
+
+suite("Mocking a Global Interface", async function () {
+    let mockedVSCodeWindow: MockedObject<typeof vscode["window"]>;
+    let originalWindowObject: typeof vscode["window"];
+
+    setup() {
```
award999, line 238: wouldn't this setup and teardown be hidden behind a utility function? Would be more prone to errors if developers had to do this themselves

src/FolderContext.ts
matthewbastien: The `FolderEvent` type was replaced with an interface that encapsulates all of the event arguments since `vscode.Event` only accepts a single argument listener. `FolderOperation` is the enumeration that used to be called `FolderEvent`:

```typescript
/** Workspace Folder Operation types */
export enum FolderOperation {
    // Package folder has been added
    add = "add",
    // Package folder has been removed
    remove = "remove",
    // Workspace folder has gained focus via a file inside the folder becoming the actively edited file
    focus = "focus",
    // Workspace folder loses focus because another workspace folder gained it
    unfocus = "unfocus",
    // Package.swift has been updated
    packageUpdated = "packageUpdated",
    // Package.resolved has been updated
    resolvedUpdated = "resolvedUpdated",
}

/** Workspace Folder Event */
export interface FolderEvent {
    operation: FolderOperation;
    workspace: WorkspaceContext;
    folder: FolderContext | null;
}
```

I don't think there's a way to keep backward compatibility here. Changing the name isn't going to work since the listener function will still be different and I don't think making a one-off mock event listener that conforms to the old function really makes sense.

What PR does this conflict with?

docs/contributor/writing-effective-tests.md
matthewbastien: Yeah the continuation after this example shows how using `mockGlobalObject()` is easier. This was just a hypothetical example to show why `mockGlobalObject()` is easier to work with:

```markdown
However, you can remove the need for custom `setup()` and `teardown()` methods ...
```

src/FolderContext.ts
award999: k, makes sense

docs/contributor/writing-effective-tests.md
award999: I missed the wording. I don't think we need to show the anti-pattern

matthewbastien: Makes sense. I replaced that code example with the mockGlobalObject() one pointing out how it must be called at the `suite()` level since it handles `setup()` and `teardown()` for you.

award999 (approved):
🥇

plemarquand (approved):

PR #1094, opened 2024-09-25
Add ability to run tests on VS Code Insiders

plemarquand:
Add a new docker compose file that lets us run the tests on VS Code insiders so we can catch issues before they show up for most users in the mainline VS Code release.


docker/docker-compose.2004.main.insiders.yaml
```diff
@@ -0,0 +1,15 @@
+services:
+
+  runtime-setup:
+    image: vscode-swift:22.04-main
```
award999, line 4: do we maybe do 6.0, i.e. latest release so we know the toolchain is at least more or less stable? Nightly toolchain and nightly VS Code could make it a little harder to know the failure source

plemarquand: Good point, probably better not to have two moving targets.

award999 (approved):
LGTM

award999 (approved):

Issue #1100, opened 2024-09-25
XCTest output is not recorded when using CodeLLDB


plemarquand:
**Describe the bug**
When using CodeLLDB as the debug adapter XCTest output is not captured and so test results are not recorded. If `swift.debugger.useDebugAdapterFromToolchain` is true then output is captured and tests work as expected.

**To Reproduce**
Create a simple XCTest:

```swift
import XCTest

class MyTests: XCTestCase {
    func testFoo() {
        print("x")
    }
}
```

- Right click `testFoo()`'s test icon, Debug Test
- Notice there is Debug Console output for the test, but the test is not marked as passed.

**Environment**
 - OS: macOS
 - Swift version: swift-driver version: 1.115 Apple Swift version 6.0 (swiftlang-6.0.0.9.11 clang-1600.0.26.2)
 - vscode-swift version: 1.11.2

michael-weng:
I will verify this one.

michael-weng:
Verified with and without the patch, the bug is fixed with the patch applied.

PR #1101, opened 2024-09-26
Revert using universal debug adapter name

plemarquand:
The CodeLLDB extension expects the debug adapter to be called `lldb`. If we are using it and we don't call `registerDebugAdapterTrackerFactory` with `lldb` as the adapter name then we fail to get logs in the debugSessionCallback when running tests. This prevents XCTests results from being parsed, and hides the swift-testing output.

Revert a portion of #1024 that always set the debug adapter name to `swift-lldb` and use `lldb` when we're using CodeLLDB as the debug adapter.

Issue #1100


award999 (approved):
LGTM

plemarquand:

matthewbastien (approved):
😢

PR #1105, opened 2024-09-26
Add script that runs test in the same image as the ci test job

award999:


package.json
```diff
@@ -1277,6 +1277,7 @@
     "format": "prettier --check *.json src test",
     "pretest": "npm run compile-tests",
     "test": "vscode-test",
+    "test-ci": "docker compose -f docker/docker-compose.yaml -f docker/docker-compose.*.${SWIFT_VERSION:=main}.yaml -p swift-vscode-${SWIFT_VERSION:=main}-prb run --rm test",
```
plemarquand, line 1280: Could we move this in to a JS script in `/scripts` that looks like

```js
const { execSync } = require('child_process');

// Default to 'main' if no argument is provided
const swiftVersion = (process.argv[2] || 'main').replace(/\./g, '');
const command = `docker compose -f docker/docker-compose.yaml -f docker/docker-compose.*.${swiftVersion}.yaml -p swift-vscode-${swiftVersion}-prb run --rm test`;

console.log(`Running: ${command}`);
execSync(command, { stdio: 'inherit' });
```

That way instead of:
`SWIFT_VERSION=60 npm run test-ci`

We can the slightly more ergonomic:
`npm run test-ci 6.0`

award999: sure, great idea

matthewbastien (approved):

PR #1107, opened 2024-09-27
Add a script for running our nightly suite of tests

award999:
Right now the only difference to CI is we run @slow plus CI tests


docker/docker-compose-nightly.yaml
```diff
@@ -0,0 +1,38 @@
+# this file is not designed to be run directly
+# instead, use the docker-compose.<os>.<swift> files
+# eg docker-compose -f docker/docker-compose.yaml -f docker/docker-compose.1804.50.yaml run test
+
+services:
+
+  runtime-setup:
```
plemarquand, line 7: I dont think we need to duplicate everything, this file can be reduced to one that just sets `FAST_TEST_RUN=false`. 

Then the docker compose command for the nightly job(s) can compose three files together, starting with the existing `docker-compose.yaml`, then this one, then the one that specifies the toolchain.

The last set env var will be used, so as long as `docker-compose-nightly.yaml` appears later in the docker compose command than `docker-compose.yaml` we should have `FAST_TEST_RUN` set to `false`.

docker/test-nightly.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
+
+const { execSync } = require('child_process');
+
+// Default to 'main' if no argument is provided
+const swiftVersion = (process.argv[2] || 'main').replace(/\./g, '');
```
plemarquand, line 6: I'd adapt [test-ci.sh](https://github.com/swiftlang/vscode-swift/pull/1105/files#diff-b2b7bb5d8b2aa0763b5a3e50d5f717357080910cc4f81955ca8fd247f966a2e1) to take another argument that specifies we want to run the nightly tests. Then the npm script can do `docker/test-ci.sh nightly` or something similar and we can avoid duplicating this script.

package.json
```diff
@@ -1278,7 +1278,7 @@
     "pretest": "npm run compile-tests",
     "test": "vscode-test",
     "test-ci": "docker/test-ci.sh",
```
plemarquand, line 1280: Does this need to be updated to `docker/test-ci.sh ci` so we can still do `npm run test-ci 6.0`?

award999: ya sure

plemarquand (approved):

PR #1109, opened 2024-09-30
Unit test for debug logTracker

michael-weng:
Issue: https://github.com/swiftlang/vscode-swift/issues/1023

### DIFF
--- a/test/unit-tests/debugger/logTracker.test.ts
+++ b/test/unit-tests/debugger/logTracker.test.ts
@@ -1,227 +1,229 @@
 import * as vscode from "vscode";
+import * as sinon from "sinon";
 import { expect } from "chai";
 import { mockObject, instance, MockedObject } from "../../MockUtils";
 import {
     LoggingDebugAdapterTracker,
     LoggingDebugAdapterTrackerFactory,
 } from "../../../src/debugger/logTracker";
-import sinon = require("sinon");
 
-suite("LoggingDebugAdapterTrackerFactory Test Suite", () => {
-    const factory = new LoggingDebugAdapterTrackerFactory();
+suite("logTracker Tests", () => {
+    suite("LoggingDebugAdapterTrackerFactory Test Suite", () => {
+        const factory = new LoggingDebugAdapterTrackerFactory();
 
-    // Clean up the static members before each test to ensure isolation
-    teardown(() => {
-        LoggingDebugAdapterTracker["debugSessionIdMap"] = {};
-    });
-
-    test("should create LoggingDebugAdapterTracker instance", () => {
-        const mockSession = mockObject<vscode.DebugSession>({ id: "session-1" });
-        const tracker = factory.createDebugAdapterTracker(instance(mockSession));
-
-        expect(tracker).to.be.instanceOf(LoggingDebugAdapterTracker);
-        expect((tracker as LoggingDebugAdapterTracker).id).to.equal("session-1");
-    });
-
-    test("should handle zero, one, and multiple sessions", () => {
-        const mockSession1 = mockObject<vscode.DebugSession>({ id: "session-1" });
-        const mockSession2 = mockObject<vscode.DebugSession>({ id: "session-2" });
-
-        // Test 0 sessions: map should be empty
-        expect(Object.keys(LoggingDebugAdapterTracker["debugSessionIdMap"])).to.be.empty;
-
-        // Test 1 session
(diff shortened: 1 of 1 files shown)


test/unit-tests/debugger/logTracker.test.ts
```diff
@@ -0,0 +1,229 @@
...
+            expect(tracker).to.be.instanceOf(LoggingDebugAdapterTracker);
+            expect((tracker as LoggingDebugAdapterTracker).id).to.equal("session-1");
+        });
+
+        test("should handle zero, one, and multiple sessions", () => {
+            const mockSession1 = mockObject<vscode.DebugSession>({ id: "session-1" });
+            const mockSession2 = mockObject<vscode.DebugSession>({ id: "session-2" });
+
+            // Test 0 sessions: map should be empty
+            expect(Object.keys(LoggingDebugAdapterTracker["debugSessionIdMap"])).to.be.empty;
```
award999, line 32: I'm assuming this is a private or static property you're accessing with ["debugSessionIdMap"]? This would be a case I would not add a unit test for because if we change the internal workings, the test breaks. We don't need 100% coverage if it doesn't make sense, or if the code does not easily allow unit testing, then we can test anything important at the integration level

test/unit-tests/debugger/logTracker.test.ts
```diff
@@ -0,0 +1,229 @@
...
+            const mockSession2 = mockObject<vscode.DebugSession>({ id: "session-2" });
+
+            const tracker1 = factory.createDebugAdapterTracker(
+                instance(mockSession1)
+            ) as LoggingDebugAdapterTracker;
+            const tracker2 = factory.createDebugAdapterTracker(
+                instance(mockSession2)
+            ) as LoggingDebugAdapterTracker;
+
+            expect(Object.keys(LoggingDebugAdapterTracker["debugSessionIdMap"])).to.have.length(2);
```
award999, line 58: same

test/unit-tests/debugger/logTracker.test.ts
award999: Or alternatively, do we in a separate commit rework the code so it can be unit tested

test/unit-tests/debugger/logTracker.test.ts
```diff
@@ -0,0 +1,229 @@
...
+                const sessionId = `session-${i}`;
+                const mockSession = mockObject<vscode.DebugSession>({ id: sessionId });
+                const tracker = factory.createDebugAdapterTracker(
+                    instance(mockSession)
+                ) as LoggingDebugAdapterTracker;
+                mockSessions.push(tracker);
+            }
+
+            // Verify that all 1000 sessions are in the map
+            expect(Object.keys(LoggingDebugAdapterTracker["debugSessionIdMap"])).to.have.length(
```
award999, line 87: same

test/unit-tests/debugger/logTracker.test.ts
```diff
@@ -0,0 +1,229 @@
...
+        let mockSession: MockedObject<vscode.DebugSession>;
+
+        // Before each test, reset the callback spy, static member, and create a new mock session
+        setup(() => {
+            callbackSpy = sinon.spy();
+            mockSession = mockObject<vscode.DebugSession>({ id: "session-1" });
+        });
+
+        teardown(() => {
+            LoggingDebugAdapterTracker["debugSessionIdMap"] = {}; // Reset static state
```
award999, line 112: same

award999 (changes requested):

test/unit-tests/debugger/logTracker.test.ts
michael-weng: I think it's case dependent, here all the test is doing is checking proper set up and clean up of the data structure, which imo is rather valid. If we are not to use this implementation, the test can be safely removed.

award999: Overall manipulating static or global values where they can influence other tests is a bit of a slippery slope we need to be careful about.

Looked through the code under test and there doesn't appear to be a lot going on here to justify writing a test that is dependent on the private inner workings. The pieces like message handling would likely be implicitly covered by integration or e2e level testing. This would be a case where we may not have any unit testing, but that's ok if we don't think it provides value

award999:
now that we've set new guidelines, can we close this? Doesn't fit into our definition of unit tests

michael-weng:
Sure.

PR #1114, opened 2024-10-01
Address soundness failure

michael-weng:


src/debugger/debugAdapterFactory.ts
```diff
@@ -12,6 +11,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 //===----------------------------------------------------------------------===//
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
matthewbastien, line 14: Why not just remove the unused arguments from the `createDebugAdapterDescriptor` and `resolveDebugConfiguration` functions if they aren't being used?

matthewbastien (changes requested):

src/debugger/debugAdapterFactory.ts
michael-weng: Good call, done.

matthewbastien (approved):

src/debugger/debugAdapterFactory.ts
plemarquand: Is this line necessary anymore after removing them?

plemarquand (approved):

Issue #1113, opened 2024-10-01
Add test cases for Package Dependencies view


award999:
Provide a reasonable set of unit and integration test cases that verify the Package Dependencies view. Initial list of cases:
- Creating correct tree items
- Recursively handles directories
- Notifies of changes

PR #1117, opened 2024-10-02
Package dependency view tests

award999:
* Remove some unused private functions
* Integration tests for package dependencies view
* Add unit tests for how tree items are created

Issue: #1113


test/unit-tests/ui/PackageDependencyProvider.test.ts
```diff
@@ -0,0 +1,62 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import { expect } from "chai";
+import * as vscode from "vscode";
+import { FileNode, PackageNode } from "../../../src/ui/PackageDependencyProvider";
+
+suite("PackageDependencyProvider Unit Test Suite", function () {
+    suite("FileNode", () => {
+        test("toTreeItem, file", () => {
```
matthewbastien, line 21: The test title should give someone an overview of what the test is doing without actually having to read the code. `toTreeItem, file` doesn't really tell me what the test is supposed to do. Something like `can create a VSCode TreeItem that opens a file` would make it clearer what's going on.

test/unit-tests/ui/PackageDependencyProvider.test.ts
```diff
@@ -0,0 +1,62 @@
...
+            expect(item.label).to.equal("Foo");
+            expect(item.resourceUri).to.deep.equal(expectedUri);
+            expect(item.command).to.deep.equal({
+                command: "vscode.open",
+                arguments: [expectedUri],
+                title: "Open File",
+            });
+        });
+
+        test("toTreeItem, Directory", () => {
```
matthewbastien, line 35: Same issue here. Something like `can create a VSCode TreeItem that opens a file` would make it clearer what's going on.

test/unit-tests/ui/PackageDependencyProvider.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+            const expectedUri = vscode.Uri.file("/path/to/Foo");
+            expect(item.label).to.equal("Foo");
+            expect(item.resourceUri).to.deep.equal(expectedUri);
+            expect(item.command).to.be.undefined;
+        });
+    });
+
+    suite("PackageNode", () => {
+        test("toTreeItem", () => {
```
matthewbastien, line 47: `can create a VSCode TreeItem that represents a Swift package`

matthewbastien (changes requested):

assets/test/dependencies/Sources/main.swift
```diff
@@ -0,0 +1,3 @@
+import SwiftSyntax
```
plemarquand, line 1: I know this doesn't actually affect the tests but should this import Swift Markdown instead of Swift Syntax?

test/integration-tests/ui/PackageDependencyProvider.test.ts
```diff
@@ -0,0 +1,118 @@
...
+import { folderContextPromise, globalWorkspaceContextPromise } from "../extension.test";
+import { executeTaskAndWaitForResult, waitForNoRunningTasks } from "../../utilities";
+import { getBuildAllTask, SwiftTask } from "../../../src/tasks/SwiftTaskProvider";
+import { testAssetPath } from "../../fixtures";
+import { Version } from "../../../src/utilities/version";
+
+suite("PackageDependencyProvider Test Suite", function () {
+    let treeProvider: PackageDependenciesProvider;
+    suiteSetup(async function () {
+        this.timeout(2 * 60 * 1000); // Allow up to 2 minutes to build
```
plemarquand, line 29: nit, other tests set the timeout at the top level using `this.timeout(...)` instead of inside a suiteSetup

assets/test/dependencies/Sources/main.swift
award999: Awesome catch, I swapped libs

test/integration-tests/ui/PackageDependencyProvider.test.ts
award999: done

matthewbastien (approved):

plemarquand (approved):

src/ui/PackageDependencyProvider.ts
```diff
@@ -257,7 +257,7 @@ export class PackageDependenciesProvider implements vscode.TreeDataProvider<Tree
         const workspacePath = folderContext.folder.fsPath;
 
         const showingDependencies: Set<string> = new Set<string>();
-        const stack: Dependency[] = rootDependencies;
+        const stack: Dependency[] = rootDependencies.slice();
```
michael-weng, line 260: Why is a shallow copy needed here?

award999: So this was a legit bug in the source, since popping below the dependency was lost, so subsequent times fetching children resulted in missing dependencies. Basically refreshes to the tree can cause stuff to be missed

Issue #1077, opened 2024-09-18
error: Activating Swift extension failed: Failed to find lldb-dap in Swift toolchain


thaneliu:
## Environment:
- CPU: Intel x64
- OS: macOS Sequoia Version 15.0
- Xcode: Xcode Version 16.0(16A242d)
- VSCode: VSCode Version: 1.93.1
- vscode-swift: v1.110

## Reproduce:
1. upgrade to macOS Squoia 15.0
2. upgrade to Xcode 16.0
3. upgrade to VSCode 1.93.1
4. create a iOS App project by XCode 16.0
5. open the project by VSCode

## Screenshots:
<img width="210" alt="image" src="https://github.com/user-attachments/assets/e3adc956-091e-4a35-919a-dddae490422b">

<img width="566" alt="image" src="https://github.com/user-attachments/assets/af5591d6-8daa-4953-a2f1-f3132a9f446e">

<img width="1084" alt="image" src="https://github.com/user-attachments/assets/f01bfb35-45b2-4fcf-8f66-ec71dc493efb">

<img width="296" alt="image" src="https://github.com/user-attachments/assets/230c8762-f1bc-4507-95ab-c45df883e4b2">

plemarquand:
@fearnliu could you run `> Swift: Capture VS Code Swift Diagnostic Bundle`, choose Minimal Bundle and attach the two files to this issue?

fearnliu:
> @fearnliu could you run `> Swift: Capture VS Code Swift Diagnostic Bundle`, choose Minimal Bundle and attach the two files to this issue?
> 
> Have you updated to the newest Xcode (Xcode 16)?

Hello~ thanks for your reply.

I have run the command in VScode. there are two files in attach files.
yep~, Xcode is Version 16.0(has upgraded to the latest version in Mac App Store recently), and the project "SmartAlarmClock"(ios APP) was created by Xcode 16. I have just opened the project in my VSCode.

By the way, my mac cpu is intel x64.

[extension-logs.txt](https://github.com/user-attachments/files/17044807/extension-logs.txt)
[settings.txt](https://github.com/user-attachments/files/17044811/settings.txt)

matthewbastien:
[...]

Issue #1069, opened 2024-09-14
[CursorAI] Activating Swift extension failed: Failed to find lldb-dap in Swift toolchain


xzilja:
**Describe the bug**
I am trying out [CursorAI](https://www.cursor.com) which is an editor based off VSCode 1.91.1. Commonly there are no issues with extensions, but when I add [Swift](https://marketplace.visualstudio.com/items?itemName=sswg.swift-lang) one, I get an error toast  saying "Activating Swift extension failed: Failed to find lldb-dap in Swift toolchain". 

After some googling around, common fix stated is "install swift", however when I run `swift --version` I do get following output

```
swift-driver version: 1.115 Apple Swift version 6.0 (swiftlang-6.0.0.9.10 clang-1600.0.26.2)
Target: arm64-apple-macosx15.0
```

So am currently not sure if there is anything else I could try, hence leading me to think that this might be a bug with either using cursor or swift 6.0?

I also did not get any prompts to set any extras up after installing the extension.

**To Reproduce**
1. Download https://www.cursor.com
2. Add `Swift` Extension, restart editor just in case
3. Open any swift project that uses some import i.e. `SwiftUI` and you'll get errors all across when using any of the views
4. Additionally observe error toast that pops up at the bottom left

**Expected behavior**
Should install and set everything up correctly.

**Environment**
 - OS: macOS Sequoia
 - Swift version: 
 ```
swift-driver version: 1.115 Apple Swift version 6.0 (swiftlang-6.0.0.9.10 clang-1600.0.26.2)
Target: arm64-apple-macosx15.0
```
 - Visual Studio Code version: `Version: 0.40.4`, `VSCode Version: 1.91.1`
 - vscode-swift version: `v1.11.0`

**Additional context**
-

plemarquand:
Hi @xzilja, from inside Cursor could you do `>  Swift: Capture VS Code Swift Diagnostic Bundle`, choose `Capture Minimal Diagnostics` and attach the results to this issue? 

Do you have Xcode installed, have you installed a Swift toolchain from https://www.swift.org/download/, or none of the above?

In the meantime you can try working around this by setting `"swift.debugger.useDebugAdapterFromToolchain": false` in your VS Code user settings.json.

xzilja:
@plemarquand Sure thing, here are the results
[extension-logs.txt](https://github.com/user-attachments/files/17005275/extension-logs.txt)
[...]

PR #1119, opened 2024-10-02
[macOS] Properly handle cases where `lldb-dap` cannot be found with xcrun

matthewbastien:
Some users were seeing the extension fail to activate due to `lldb-dap` not being found in the toolchain. I've tracked this down to three underlying issues with the extension:
1. The `DebugAdapter.verifyDebugAdapterExists()` function was awaited on extension startup
2. The `SwiftToolchain.getLLDB()` and `SwiftToolchain.getLLDBDebugAdapter()` functions can throw errors when the executables are not found. This was not being caught properly by `DebugAdapter.verifyDebugAdapterExists()`.
3. There are inconsistencies between `DebugAdapter.adapterName` and `DebugAdapter.getAdapterType()` that can cause them to return different debug adapters under certain circumstances.

The `DebugAdapter.verifyDebugAdapterExists()` will now catch errors thrown by the SwiftToolchain when finding executables and the extension activation now properly runs this check in the background. I've also updated the docstrings for `getLLDB()` and `getLLDBDebugAdapter()` to make it more obvious that they can throw and added some tests in the hopes that this won't happen again.

I've replaced the `DebugAdapter.adapterName` property with a `DebugAdapter.getAdapterName()` function that matches the signature and behaviour of `DebugAdatper.getAdapterType()`, but instead returns the executable name of the debug adapter. This will hopefully reduce any further issues with adapter mismatches from happening in the future.

I've added `mock-fs` as a dependency now since these new tests rely on a lot of file system checks and it's easier to mock the `fs` module this way. The documentation for writing tests has been updated to reflect this.

Related issues: #1069 #1077


docs/contributor/writing-tests-for-vscode-swift.md
```diff
@@ -113,6 +114,30 @@ suite("ReloadExtension Unit Test Suite", () => {
...
 You may have also noticed that we needed to cast the `"Reload Extensions"` string to `any` when resolving `showWarningMessage()`. Unforunately, this may be necessary for methods that have incompatible overloaded signatures due to a TypeScript issue that remains unfixed.
 
+## Mocking the File System
+
+The [`mock-fs`](https://github.com/tschaub/mock-fs) module can be used to temporarily replace Node's built-in `fs` module with an in-memory file system. This can be useful for testing logic that uses the `fs` module without actually reaching out to the file system. Just make sure that you add a `teardown()` block that restores the `fs` module after each test:
+
+```typescript
+import * as chai from "chai";
+import * as mockFS from "mock-fs";
+import * as fs from "fs/promises";
```
award999, line 124: let's go over this with group. Not clear where the utilities we have are lacking that we need to add a new dependency. Not saying we shouldn't do this, but should make clear the rational since this needs to be adopted for the project's strategy

src/debugger/debugAdapter.ts
```diff
@@ -24,21 +24,19 @@ import { SwiftToolchain } from "../toolchain/toolchain";
...
     }
 
     /** Return debug adapter for toolchain */
-    public static getDebugAdapterType(swiftVersion: Version): "lldb-vscode" | "lldb-dap" {
+    public static getDebugAdapterType(swiftVersion: Version): "lldb" | "lldb-dap" {
         return swiftVersion.isGreaterThanOrEqual(new Version(6, 0, 0)) &&
             configuration.debugger.useDebugAdapterFromToolchain
             ? "lldb-dap"
-            : "lldb-vscode";
+            : "lldb";
```
award999, line 39: so it seems like this string needs to match the launch configuration type provided by CodeLLDB. We should make that more clear with some comments, and probably worth introducing some CODELLDB_TYPE constant to make intent clear

src/debugger/debugAdapterFactory.ts
```diff
@@ -102,7 +102,7 @@ export class LLDBDebugConfigurationProvider implements vscode.DebugConfiguration
         }
 
         // Delegate to CodeLLDB if that's the debug adapter we have selected
-        if (DebugAdapter.getDebugAdapterType(this.swiftVersion) === "lldb-vscode") {
+        if (DebugAdapter.getDebugAdapterType(this.swiftVersion) === "lldb") {
```
award999, line 105: again constant would just mean less places to change in the future

matthewbastien: done

docs/contributor/writing-tests-for-vscode-swift.md
matthewbastien: Fair. The problem with our current setup is that there are multiple ways to test for files existing. You could do `fs.readFile()` and catch failures, or you could do the same with `fs.stat()`. The `mock-fs` module lets you declare in your test which files exist under which directories and backs the entirety of the `fs` module with an in-memory filesystem. This means that the test is more resilient to changes in implementation.

test/unit-tests/debugger/debugAdapterFactory.test.ts
```diff
@@ -45,18 +45,18 @@ suite("Debug Adapter Factory Test Suite", () => {
             const configProvider = new LLDBDebugConfigurationProvider("darwin", swift6);
             const launchConfig = await configProvider.resolveDebugConfiguration(undefined, {
                 name: "Test Launch Config",
-                type: DebugAdapter.adapterName,
+                type: "swift-lldb",
```
michael-weng, line 48: Probably good idea to the use the newly added enum here as well?

matthewbastien: Yup. Good catch!

docs/contributor/writing-tests-for-vscode-swift.md
matthewbastien: I've added a paragraph here to talk about why `mock-fs` is necessary.

docs/contributor/writing-tests-for-vscode-swift.md
```diff
@@ -113,6 +114,35 @@ suite("ReloadExtension Unit Test Suite", () => {
...
+
+suite("mock-fs example", () => {
+    // This teardown step is also important to make sure your tests clean up the
+    // mocked file system when they complete!
+    teardown(() => {
+        mockFS.restore();
```
award999, line 132: in a subsequent change, maybe we should handle this in a utility method so this is done for them. Now that we've created all these cool utilities

matthewbastien: Yeah I could write something up for that in a future PR.

docs/contributor/writing-tests-for-vscode-swift.md
```diff
@@ -113,6 +114,35 @@ suite("ReloadExtension Unit Test Suite", () => {
 
 You may have also noticed that we needed to cast the `"Reload Extensions"` string to `any` when resolving `showWarningMessage()`. Unforunately, this may be necessary for methods that have incompatible overloaded signatures due to a TypeScript issue that remains unfixed.
 
+## Mocking the File System
+
+Mocking file system access can be a challenging endeavor that is prone to fail when implementation details of the unit under test change. This is because there are many different ways of accessing and manipulating files, making it almost impossible to catch all possible failure paths. For example, you could check for file existence using `fs.stat()` or simply call `fs.readFile()` and catch errors with a single function call. Using the real file system is slow and requires extra setup code in test cases to configure.
```
michael-weng, line 119: For pattern where you want to test when the fs functions throws do we use the mockGlobalModule route?

matthewbastien: To test a file not existing you can just create an empty file system which will cause `fs.stat()` or `fs.readFile()` to throw an error. You could also set the mode on the created file using `mockFS.file({ mode: 0o000 })` so that it's not readable/writable by the current user.

matthewbastien: Good question, though. I'm going to add an example for this.

michael-weng: Yeah, because to mock it to throw is like 2 lines. But I think what you said is better as it's more explicit about the expectation around fs and we should establish as such.

plemarquand (approved):

PR #1127, opened 2024-10-08
Disable flakey diagnostics tests in Swift 5.6 and below

plemarquand:
It appears there is an issue in Swift 5.6 where a line of diagnostics output is improperly cleared from the terminal by the build progress prints. When this happens the tests potentially miss diagnostics they expect to see. This only happens on Swift 5.6, so to ensure these tests pass consistently, skip them there.


plemarquand:

award999 (approved):

PR #1128, opened 2024-10-08
Fix unit test that gets skipped if not on Windows

matthewbastien:
There's a unit test that gets skipped if the platform isn't Windows in order to test some functionality of `getLLDBLibPath()`. Use `mockGlobalValue()` here to set the value of `process.platform` so that the test can run on all platforms.


plemarquand:
Try rebasing this, a few test stability fixes are in that should resolve the CI failure seen here.

PR #1129, opened 2024-10-08
Fix swift-testing test not always capturing terminal output

plemarquand:
Occasionally a swift-testing test can start printing output before the test start JSON event is recieved over the named pipe. This output was always recorded to the test run, but wasn't necessarily captured in the structure used to record test results for the vscode-swift unit tests.

This would cause the `swift-testing: Runs passing test` test to occasionally fail when it couldn't find the output line printed during the test, because it was printed before the test start event was recieved.

Record the output no matter when it occurs so that it can be verified in the tests correctly.


plemarquand:

PR #1131, opened 2024-10-09
Ensure that integration tests execute the same code as the extension

matthewbastien:
Build the test code and source into `./dist` rather than `./out`. This fixes two issues:

1. The test code used to import different source modules from VS Code which caused issues during integration tests.
2. The code coverage for integration tests was inconsistent as it was coming from code executed from both the `dist` and `out` folders.

The downside of this approach is that building in `watch` mode can overwrite files when you try to run a test at the same time. Thankfully, this can be easily solved by performing a separate build every time the extension host is launched. There is no noticeable performance impact with this as `esbuild` is quite fast.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -6,11 +6,12 @@
 		{
 			"label": "build",
 			"type": "npm",
-			"script": "esbuild",
-			"problemMatcher": "$esbuild",
+			"script": "watch",
+			"problemMatcher": "$tsc-watch",
 			"presentation": {
 				"reveal": "never"
 			},
+			"isBackground": true,
 			"group": {
 				"kind": "build",
 				"isDefault": true
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -6,9 +6,8 @@
 		{
 			"label": "build",
 			"type": "npm",
-			"script": "esbuild-watch",
-			"problemMatcher": "$esbuild-watch",
-			"isBackground": true,
+			"script": "esbuild",
+			"problemMatcher": "$esbuild",
 			"presentation": {
 				"reveal": "never"
 			},
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "workspaceContains:compile_commands.json",
     "onDebugResolve:swift-lldb"
   ],
-  "main": "./dist/extension.js",
+  "main": "./dist/src/extension.js",
   "contributes": {
     "icons": {
       "swift-icon": {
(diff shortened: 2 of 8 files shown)


award999:
Is there no change in coverage, i.e. same numbers? Does breakpoints work in test source, as well as extension code?

matthewbastien:
Yes, I verified that breakpoints are working still for our launch configurations. As for the code coverage changing: it definitely did, but I'm not sure there's a perfect solution for this.

Before my change:
```
> FAST_TEST_RUN=1 npm run integration-test -- --coverage

---------------------------------|---------|----------|---------|---------|
File                             | % Stmts | % Branch | % Funcs | % Lines |
---------------------------------|---------|----------|---------|---------|
All files                        |   59.29 |    72.73 |   62.59 |   59.29 |
```

After my change:
```
> FAST_TEST_RUN=1 npm run integration-test -- --coverage

---------------------------------|---------|----------|---------|---------|
File                             | % Stmts | % Branch | % Funcs | % Lines |
---------------------------------|---------|----------|---------|---------|
All files                        |   74.33 |    78.36 |   74.68 |   74.33 |
```

Neither of these numbers are going to be particularly accurate as the extension host is executing code that we aren't actually performing any assertions on. It is, however, a more accurate representation of what code is _**actually**_ being executed as its all coming from the same source. Before, unexecuted code from `out/src/**` and `dist/**` were being combined to decrease the code coverage score.

It's impossible to get accurate information here, unfortunately. We cannot differentiate between the two even with the old method as sometimes we were executing code by retrieving the extension API which would execute code from `./dist/**` and other times we were importing modules directly from `./out/src/**`. I think this unification of our code coverage output just makes it easier to consume.

package.json
```diff
@@ -1267,11 +1267,9 @@
   ],
   "scripts": {
     "vscode:prepublish": "npm run esbuild-bundle",
-    "esbuild-base": "del-cli ./dist && esbuild \"./src/**/*.ts\" --outdir=dist --format=cjs --platform=node --target=node18",
-    "esbuild": "npm run esbuild-base",
-    "esbuild-watch": "npm run esbuild-base -- --watch",
-    "esbuild-bundle": "del-cli ./dist && esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node --target=node18 --minify --sourcemap",
-    "compile": "tsc",
+    "esbuild": "del-cli ./dist && esbuild \"./src/**/*.ts\" --outdir=dist/src --format=cjs --platform=node --target=node18 --sourcemap",
```
award999, line 1270: does it make sense to keep this around anymore? Use esbuild to bundle, tsc as pre launch task. We can have a single watch job whether running extension or tests

matthewbastien: Yeah that will definitely work, good catch! Let me fix it up.

award999 (approved):

.vscode/tasks.json
```diff
@@ -6,12 +6,12 @@
 		{
 			"label": "build",
 			"type": "npm",
-			"script": "esbuild-watch",
-			"problemMatcher": "$esbuild-watch",
-			"isBackground": true,
+			"script": "watch",
+			"problemMatcher": "$tsc-watch",
```
plemarquand, line 10: Any differences of note between using esbuild-watch and tsp-watch?

plemarquand (approved):

.vscode/tasks.json
matthewbastien: ESBuild had a different output from the TypeScript Compiler which is what this PR uses for development builds. So, I had to update this to match. As a side note: this actually works better because `esbuild-watch` never reported issues like `tsc-watch` does. So, you'll get notified if your build fails to compile properly before you launch a debug session.

PR #1134, opened 2024-10-10
Fix error in extension.test.ts

matthewbastien:
VS Code recently updated its built-in TypeScript Language Server to v5.6.2 which shows an error in `extension.test.ts` that is not present in CI (currently this repo uses TypeScript 5.5.2). However, this is a legitimate issue as the `rootWorkspaceFolder` variable can indeed be undefined.

I've simply fixed up the code in this area to properly handle the fact that the variable can be undefined as well as an issue where a `throw` would not have been handled properly. However, I'm hesitant to jump to TypeScript 5.6.2 since TypeScript ESLint currently does not support 5.6.

### DIFF
--- a/test/integration-tests/extension.test.ts
+++ b/test/integration-tests/extension.test.ts
@@ -14,32 +14,39 @@
 
 import * as vscode from "vscode";
 import * as assert from "assert";
-import * as swiftExtension from "../../src/extension";
+import { Api } from "../../src/extension";
 import { WorkspaceContext } from "../../src/WorkspaceContext";
 import { getBuildAllTask } from "../../src/tasks/SwiftTaskProvider";
 import { SwiftExecution } from "../../src/tasks/SwiftExecution";
 import { testAssetUri } from "../fixtures";
 import { FolderContext } from "../../src/FolderContext";
 
-export const rootWorkspaceFolder = vscode.workspace.workspaceFolders?.values().next().value;
-export const globalWorkspaceContextPromise = new Promise<WorkspaceContext>(resolve => {
-    const workspaceFolder = vscode.workspace.workspaceFolders?.values().next().value;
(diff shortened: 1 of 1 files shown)


test/integration-tests/extension.test.ts
```diff
@@ -14,32 +14,39 @@
...
 import * as assert from "assert";
-import * as swiftExtension from "../../src/extension";
+import { Api } from "../../src/extension";
 import { WorkspaceContext } from "../../src/WorkspaceContext";
 import { getBuildAllTask } from "../../src/tasks/SwiftTaskProvider";
 import { SwiftExecution } from "../../src/tasks/SwiftExecution";
 import { testAssetUri } from "../fixtures";
 import { FolderContext } from "../../src/FolderContext";
 
-export const rootWorkspaceFolder = vscode.workspace.workspaceFolders?.values().next().value;
```
award999, line 24: assert(rootWorkspaceFolder) would work too

matthewbastien: done

award999 (approved):

award999:
> ESLint currently does not support 5.6.
we should raise a bug for this, the fact we don't get tsc errors when building we cannot let stay. Need to watch for ESLint release so we can upgrade here. Can we temporarily downgrade `typescript`?

matthewbastien:
> we should raise a bug for this, the fact we don't get tsc errors when building we cannot let stay.

I can raise an issue for this and point it at the relevant issue from TypeScript ESLint.

> Can we temporarily downgrade `typescript`?

Our version of TypeScript right now is fine. I just don't want to upgrade us to 5.6.2. This PR also tells VS Code to use the version of TypeScript in `node_modules`. So, there at least won't be any discrepancies in errors anymore.

Issue #90, opened 2021-12-24
Publish in the Open VSX Registry


moritzsternemann:
I'd like to suggest to make this extension available through [Open VSX](https://github.com/eclipse/openvsx), a vendor-neutral open-source alternative to the Visual Studio Marketplace.

I'm an avid user of [VSCodium](https://github.com/VSCodium/vscodium#why-does-this-exist), a FooS distribution of Visual Studio Code that provides MIT-licensed binary releases of VS Code that don't include Microsoft's telemetry, for example. Due to licensing concerns, VSCodium uses the Open VSX extension registry over Microsoft's Marketplace. You can [read more here](https://github.com/VSCodium/vscodium/blob/master/DOCS.md#extensions--marketplace).

The process of publishing to Open VSX is described in detail [in their official documentation](https://github.com/eclipse/openvsx/wiki/Publishing-Extensions). If this addition is of interest to you, I'm happy to help with the setup and maintenance!

Last but not least, thank you to everyone involved for this awesome addition to the Swift development ecosystem!

svanimpe:
Can you try it out on Codium first? I've seen reports from users that Codium can behave differently from Code, so the things we rely on (such as CodeLLDB) may not work there.

fourplusone:
+1 on this one. Publishing the extension on the OpenVSX registry would also make it available for use in non Microsoft services like https://gitpod.io. I can also confirm it works on VSCodium, tested on

```
Version: 1.66.2
Commit: dfd34e8260c270da74b5c2d86d61aee4b6d56977
Date: 2022-04-13T00:45:34.343Z (4 days ago)
Electron: 17.2.0
Chromium: 98.0.4758.109
Node.js: 16.13.0
V8: 9.8.177.11-electron.0
OS: Darwin x64 21.4.0
```

filiptronicek:
Hi @svanimpe, a [Gitpod](https://github.com/gitpod-io/gitpod/) and [OpenVSX](https://github.com/eclipse/openvsx/) commiter here 👋.

[...]

PR #1137, opened 2024-10-11
Rename the extension from `sswg.swift-lang` to `swiftlang.vscode-swift`

plemarquand:
In preparation for the move in the VS Code marketplace, rename the extension ID and associated references in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,17 +12,19 @@ This extension adds language support for Swift to Visual Studio Code. It support
 
 Swift support uses [SourceKit LSP](https://github.com/apple/sourcekit-lsp) for the [language server](https://microsoft.github.io/language-server-protocol/overviews/lsp/overview/) to power code completion and [LLDB](https://github.com/vadimcn/vscode-lldb) to enable debugging.
 
-The extension is developed by members of the Swift Community and maintained by the [SSWG](https://www.swift.org/sswg/). The aim is to provide a first-class, feature complete extension to make developing Swift applications on all platforms a seamless experience.
+The aim is to provide a first-class, feature complete extension to make developing Swift applications on all platforms a seamless experience.
 
 If you experience any issues or want to propose new features please [create an issue](https://github.com/swiftlang/vscode-swift/issues/new) or post on the [swift.org forums](https://forums.swift.org) in the [VS Code Swift Extension category](https://forums.swift.org/c/related-projects/vscode-swift-extension/).
 
 ## Contributing
 
-The Swift for Visual Studio Code extension is a community driven project, developed by the amazing Swift community. Any kind of contribution is appreciated, including code, tests and documentation. For more details see [CONTRIBUTING.md](CONTRIBUTING.md).
+The Swift for Visual Studio Code extension is a community driven project originally created by the [Swift Server Working Group](https://www.swift.org/sswg/) and now maintained as part of the [swiftlang organization](https://github.com/swiftlang/). Contributions are appreciated, including code, tests and documentation. For more details see [CONTRIBUTING.md](CONTRIBUTING.md).
+
+To give clarity of what is expected of our members, Swift has adopted the code of conduct defined by the Contributor Covenant. This document is used across many open source communities, and we think it articulates our values well. For more, see the [Code of Conduct](https://swift.org/code-of-conduct/).
 
 ## Installation
 
-For the extension to work, you must have Swift installed on your system. Please see the [Getting Started Guide on Swift.org](https://www.swift.org/getting-started/) for details on how to install Swift on your system. Install the extension from [VS Code Marketplace](https://marketplace.visualstudio.com/items?itemName=sswg.swift-lang) and open a Swift package! You'll be prompted to install and configure the CodeLLDB extension, which you should do so.
+For the extension to work, you must have Swift installed on your system. Please see the [Getting Started Guide on Swift.org](https://www.swift.org/getting-started/) for details on how to install Swift on your system. Install the extension from [VS Code Marketplace](https://marketplace.visualstudio.com/items?itemName=swiftlang.vscode-swift) and open a Swift package! You'll be prompted to install and configure the CodeLLDB extension, which you should do so.
 
 ## Features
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -12,17 +12,19 @@ This extension adds language support for Swift to Visual Studio Code. It support
...
 
 ## Installation
 
-For the extension to work, you must have Swift installed on your system. Please see the [Getting Started Guide on Swift.org](https://www.swift.org/getting-started/) for details on how to install Swift on your system. Install the extension from [VS Code Marketplace](https://marketplace.visualstudio.com/items?itemName=sswg.swift-lang) and open a Swift package! You'll be prompted to install and configure the CodeLLDB extension, which you should do so.
+For the extension to work, you must have Swift installed on your system. Please see the [Getting Started Guide on Swift.org](https://www.swift.org/getting-started/) for details on how to install Swift on your system. Install the extension from [VS Code Marketplace](https://marketplace.visualstudio.com/items?itemName=swiftlang.vscode-swift) and open a Swift package! You'll be prompted to install and configure the CodeLLDB extension, which you should do so.
```
0xTim, line 27: Is the name fixed? `vscode-swift` both seems redundant and would suggest it excludes other IDEs that the extension should work in (e.g. #90 )

FWIW all the other 1st class language extensions just use the language name, like [`golang.Go`](https://marketplace.visualstudio.com/items?itemName=golang.Go) and [`rust-lang.rust`](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust)

adam-fowler: I've seen Microsoft do this eg `Microsoft.vscode-python` but yeah the `vscode-` still does seem kind of redundant.

test/integration-tests/extension.test.ts
```diff
@@ -29,9 +29,9 @@ function getRootWorkspaceFolder(): vscode.WorkspaceFolder {
 
 export const globalWorkspaceContextPromise: Promise<WorkspaceContext> = (async () => {
     const workspaceFolder = getRootWorkspaceFolder();
-    const ext = vscode.extensions.getExtension<Api>("sswg.swift-lang");
+    const ext = vscode.extensions.getExtension<Api>("swiftlang.swift-lang");
```
lokesh-tr, line 32: I find this to be inconsistent with other places where you refactored them as `"swiftlang.vscode-swift"`. Same below.

README.md
lokesh-tr: Not sure what the convention is, but I think it would be nice to spell it out as: `swiftlang.extension` (can be read as "Swift Language Extension"). It feels natural to me, would like to know what you guys think?

lokesh-tr: I don't find `swiftlang.swift` ("Swift Language Swift") and `swiftlang.vscode-swift` ("Swift Language VS Code Swift") to read out well and it feels redundant

lokesh-tr: Checking with the extension's marketplace page, it describes the extension as "Swift Language Support for Visual Studio Code". So, It feels natural for me to name the extension ID as `swiftlang.support` or `swiftlang.extension`. If not for these two, going with `swiftlang.swift` will be better than going with `swiftlang.vscode-swift`. :)

0xTim: It's `org.extension_name` so `support` or `extension` could be confusion if any other extensions were added under the org

plemarquand: I'm not married to the name but I would like to try and avoid conflicts in the "swiftlang" OSS namespace. The extension ID `swiftlang.swift` might cause confusion by conflicting with the swiftlang/swift repo. 

My thinking was if the publisher+name matched the GitHub org+repo it would be more consistent and discoverable.

plemarquand: @0xTim I like your point about not being specifically for VS Code; compatible forks are gaining popularity (VS Codium, Cursor). It may be worth doing a pass on the language in the documentation in a follow on PR to refine the language so it's more generic.

timsneath: @0xTim, totally get your point about other forks that support VSCode extensions. If there were a consistent generic term that was well understood, that might be an alternative. VSIX is the extension name, but it's a little overloaded with connotations from "full" Visual Studio; there's also VSX (as in, the [Open VSX repository](https://open-vsx.org)). 

That said, even in the open repository, they describe it as "extensions for VS Code-compatible editors", which suggests that no other term has yet reached the desired level of general understanding. 

The other considerations that went into proposing `swiftlang.vscode-swift` were:

1. It matches the repository name, so anyone encountering this extension already has an inkling of where to find its home.
2. It is confusion-free both in the Swift context and the Visual Studio context. A user who encounters this on their system will understand what it is and where it comes from. 
3. This name is purely used as a unique identifier. We wouldn't / shouldn't use it in user-facing documentation. 

@plemarquand Good call on making the language more generic (although we'd need to test the extension in a few other places to really feel comfortable that it doesn't have any hidden dependencies on VSCode: I don't think we test it anywhere else, do we?)

plemarquand: @timsneath there aren't any formalized tests with other forks but I've used Cursor casually and the extension has worked great. 

VS Codium is basically identical to VS Code but with telemetry disabled so I've got high confidence the extension will work, but I haven't tried loading the .vsix in to it yet. Will make a note that needs to be tested there.

adam-fowler: > @plemarquand Good call on making the language more generic (although we'd need to test the extension in a few other places to really feel comfortable that it doesn't have any hidden dependencies on VSCode: I don't think we test it anywhere else, do we?)

I think Apple legal still have to review the terms and conditions to publish on OpenVSX. So we never really treated it as an alternative platform as we weren't publishing to it (although OpenVSX have an unofficial copy up https://open-vsx.org/extension/sswg/swift-lang). Maybe now everything is moving to swiftlang we can organise an OpenVSX version.

test/integration-tests/extension.test.ts
plemarquand: Changed, thanks

plemarquand:
The `soundness` CI failure can be ignored, it is checking for broken links. The updated VS Code marketplace link doesn't exist yet but it will shortly after a new version of the extension is published.

README.md
daveyc123: Yes, once we've wrapped up the work to get the newly identified extension up on the VSCode Marketplace, we'll be working with Apple legal to move forward with the OpenVSX work.

award999 (approved):

matthewbastien (approved):

Issue #1052, opened 2024-09-04
Dependency names are lower-cased, disallowing setting a local version


nmggithub:
**Describe the bug**

I'm currently developing my reverse-engineering toolkit, [Kass](https://github.com/nmggithub/Kass) on my device in a folder named `Kass` (uppercase K). [The package itself is also named with an uppercase K](https://github.com/nmggithub/Kass/blob/a88555f822a9097abc405a7e346cc85f675e6677/Package.swift#L23), which I think is the actual causal contribution to this.

I have a package in a separate folder I use for testing it as a library, pulling in the latest `main` via my `Package.swift` like so:

```swift
let package = Package(
  ...
  dependencies: [
    ...
    .package(url: "https://github.com/nmggithub/Kass", revision: "main"),
    ...
  ],
  ...
)
```

It appears in my dependencies on VSCode like this:
<img width="165" alt="Screenshot 2024-09-04 at 11 22 09" src="https://github.com/user-attachments/assets/44534074-c819-4e2b-ac06-bf51011ead8e">

It appears as lowercase, which isn't ideal, but it doesn't affect my ability to use it. It *does* affect my ability to select my local development version as a path, though.

**To Reproduce**
1. Right click the "kass" dependency
2. Select "Use Local Version"
3. Navigate to folder (potentially named "Kass") with Kass inside and select it
4. See error:
```
'kass': package at '{/path/to}/Kass' is Kass but was expecting kass
```

**Expected behavior**
The local version should be selected.

**Environment**
 - OS: macOS Sonoma 14.6.1
 - Swift version (output of `swift --version`)
 ```
swift-driver version: 1.90.11.1 Apple Swift version 5.10 (swiftlang-5.10.0.13 clang-1500.3.9.4)
Target: arm64-apple-macosx14.0
```
 - Visual Studio Code version:
 ```
Version: 1.92.2 (Universal)
Commit: fee1edb8d6d72a0ddff41e5f71a671c23ed924b9
Date: 2024-08-14T17:29:30.058Z
Electron: 30.1.2
ElectronBuildId: 9870757
Chromium: 124.0.6367.243
Node.js: 20.14.0
V8: 12.4.254.20-electron.0
OS: Darwin arm64 23.6.0
```
 - vscode-swift version: v1.11.0

**Additional context**
It seems the VSCode plugin is running this command behind the scenes:
```
swift package edit --path {/path/to}/Kass kass
```

[...]

PR #1144, opened 2024-10-16
Add integration tests for package dependencies

michael-weng:
- Verify various contract calls with spm
- Exercises use local version workflow

Issue: #1052


test/integration-tests/ui/PackageDependencyProvider.test.ts
```diff
@@ -13,107 +13,184 @@
...
+        test("Use local dependency", async () => {
+            // Expect to fail without setting up local version
+            const tasks = (await getBuildAllTask(folderContext)) as SwiftTask;
+            let { exitCode, output } = await executeTaskAndWaitForResult(tasks);
+            expect(exitCode).to.not.equal(0);
+            expect(output).to.include("PackageLib");
+            expect(output).to.include("required");
+
+            // Contract: spm reset, resolve, update should work
+            const executeTaskSpy = sinon.spy(utilities, "executeTaskWithUI");
```
award999, line 159: this test gets into needing some intimate knowledge of the implementation for spying / mocking. If this is tied to a command, we should execute the command. Instead of mocking the user selection, you can let the command take a parameter and pass that parameter when running executeCommand

test/integration-tests/ui/PackageDependencyProvider.test.ts
```diff
@@ -13,107 +13,184 @@
...
+            workspaceContext = await globalWorkspaceContextPromise;
+            await waitForNoRunningTasks();
+            folderContext = await folderContextPromise("dependencies");
+            await workspaceContext.focusFolder(folderContext);
+        });
+
+        test("Use local dependency", async () => {
+            // Expect to fail without setting up local version
+            const tasks = (await getBuildAllTask(folderContext)) as SwiftTask;
+            let { exitCode, output } = await executeTaskAndWaitForResult(tasks);
```
award999, line 153: probably could move this to a setup hook so it is separate from test purpose

test/integration-tests/ui/PackageDependencyProvider.test.ts
```diff
@@ -13,107 +13,184 @@
...
+            ({ exitCode, output } = await executeTaskAndWaitForResult(tasks));
+            expect(exitCode).to.equal(0);
+            expect(output).to.include("defaultpackage");
+            expect(output).to.include("not used by any target");
+
+            // Contract: spm unedit
+            const updateWorkspaceSpy = sinon.spy(vscode.workspace, "updateWorkspaceFolders");
+            // We would love to call uneditDependency for coverage but there's no clean way to get
+            // a synchronize point for deterministic task completion so just call this function direct
+            await uneditFolderDependency(workspaceContext.currentFolder!, id, workspaceContext);
```
award999, line 186: > "full workflow", i.e. multiple paths and actions, is more smoke level. Find to group a feature in a suite, but > let's make sure each test case is one path.

This is sort of what I meant. I'm not quite sure what we're testing here. We're checking exit codes above, further actions and assertions in subsequent steps. Ideally 1 action, 1 result, 1 set of assertion(s) on that result for each test case

test/integration-tests/ui/PackageDependencyProvider.test.ts
```diff
@@ -13,107 +13,184 @@
...
-        expect(childFolder).to.not.be.undefined;
-
-        expect(childFolder?.path).to.equal(`${path}/Sources/CAtomic`);
-
-        const files = await treeProvider.getChildren(childFolder);
-        const file = files.find(n => n.name === "CAtomic.c");
-        expect(file).to.not.be.undefined;
-
-        expect(file?.path).to.equal(`${path}/Sources/CAtomic/CAtomic.c`);
+    suite("Full work flow tests", function () {
```
award999, line 138: "full workflow", i.e. multiple paths and actions, is more smoke level. Find to group a feature in a suite, but let's make sure each test case is one path.

assets/test/Swift-Markdown/Package.swift
```diff
@@ -0,0 +1,23 @@
+// swift-tools-version:6.0
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    // FIXME: Can be changed back to Swift-Markdown when
+    // https://github.com/swiftlang/swift-package-manager/issues/7931
+    // is released in the toolchain
+    name: "swift-markdown",
```
award999, line 10: is this arbitrary or must it be "swift-markdown" because we want it to overshadow the remote dependency? If the latter, should have a comment "needs to be swift-markdown for .... test to verify ...."

award999 (changes requested):

assets/test/Swift-Markdown/Package.swift
michael-weng: You are right this is not arbitrary, I will add comment to make this more obvious.

test/integration-tests/ui/PackageDependencyProvider.test.ts
michael-weng: Done, switched to assert on executeCommand.

test/integration-tests/ui/PackageDependencyProvider.test.ts
michael-weng: Done.

test/integration-tests/ui/PackageDependencyProvider.test.ts
michael-weng: Done.

test/integration-tests/ui/PackageDependencyProvider.test.ts
michael-weng: Done.

src/commands.ts
```diff
@@ -64,37 +64,50 @@ export function registerToolchainCommands(
     ];
 }
 
+export enum COMMANDS {
```
award999, line 67: praise: great idea 👍

src/commands/utilities.ts
```diff
@@ -31,7 +31,7 @@ export async function executeTaskWithUI(
     task: vscode.Task,
     description: string,
     folderContext: FolderContext,
-    showErrors = false,
+    showErrors = true,
```
award999, line 34: was this intentional or just for development? Let's chat before flipping the default here

test/integration-tests/ui/PackageDependencyProvider.test.ts
```diff
@@ -13,107 +13,213 @@
...
+            await workspaceContext.focusFolder(folderContext);
+        });
+
+        test("Contract: spm resolve", async () => {
+            const result = await vscode.commands.executeCommand(COMMANDS.ResolveDependencies);
+            expect(result).to.be.true;
+        });
+
+        test("Contract: spm update", async () => {
+            const result = await vscode.commands.executeCommand(COMMANDS.UpdateDependencies);
```
award999, line 149: structurally these cases are not part of the tree provider. This and the above case, can be in separate suites. For example this one mapping to tests/commands/dependencies/useLocal.test.ts

test/integration-tests/ui/PackageDependencyProvider.test.ts
```diff
@@ -13,107 +13,213 @@
...
+            // Expect to fail again now dependency is missing
+            const { exitCode, output } = await executeTaskAndWaitForResult(tasks);
+            expect(exitCode).to.not.equal(0);
+            expect(output).to.include("PackageLib");
+            expect(output).to.include("required");
+        });
+
+        const useLocalDependencyTest = async () => {
+            // Contract: spm edit with user supplied local version of dependency
+            const windowMock = sinon.stub(vscode.window, "showOpenDialog");
```
award999, line 196: let's use the utility so teardown is automatic

test/integration-tests/ui/PackageDependencyProvider.test.ts
```diff
@@ -13,107 +13,213 @@
...
-        expect(childFolder).to.not.be.undefined;
-
-        expect(childFolder?.path).to.equal(`${path}/Sources/CAtomic`);
-
-        const files = await treeProvider.getChildren(childFolder);
-        const file = files.find(n => n.name === "CAtomic.c");
-        expect(file).to.not.be.undefined;
-
-        expect(file?.path).to.equal(`${path}/Sources/CAtomic/CAtomic.c`);
+    suite("Full Work Flow Test Suite", function () {
```
award999, line 154: praise: this suite is setup nicely

award999 (changes requested):

src/commands.ts
plemarquand: Agreed this is a good idea! A nit though, stylistically having this in all caps is inconsistent, I believe it should just be named `Commands`

src/commands/utilities.ts
michael-weng: Ooops, good catch!

plemarquand:

plemarquand:
@swift-server-bot add to whitelist

test/integration-tests/commands/dependency.test.ts
```diff
@@ -0,0 +1,137 @@
...
+import { getBuildAllTask, SwiftTask } from "../../../src/tasks/SwiftTaskProvider";
+import { testAssetUri } from "../../fixtures";
+import { FolderContext } from "../../../src/FolderContext";
+import { WorkspaceContext } from "../../../src/WorkspaceContext";
+import * as sinon from "sinon";
+import { Commands } from "../../../src/commands";
+
+suite("Dependency Commmands Test Suite", function () {
+    // full workflow's interaction with spm is also longer than the default timeout
+    this.timeout(2 * 60 * 1000);
```
award999, line 32: if this is the case, we should mark suite or test cases with @slow

award999 (commented):
otherwise looks good

award999 (changes requested):
still unresolved comment

michael-weng:
> still unresolved comment

I offered solution, do you wish to reply?

test/integration-tests/commands/dependency.test.ts
award999: as long as it doesn't take 2 minutes in CI then yes makes sense to run in CI. The timeout seems high, if it can be lowered then sure can bump down

michael-weng: Done.

award999 (approved):

award999 (approved):

Issue #1147, opened 2024-10-17
Run/Debug/Coverage buttons appear on swift-testing argument TestItems


adam-fowler:
**Describe the bug**
Run/Debug/Coverage buttons appear on swift-testing argument TestItems and nothing happens when you click on them

**Expected behavior**
Run/Debug/Coverage buttons do not appear. Although ideally they should be there and run the test with that argument.

**Environment**
 - OS: macOS 15
 - Swift version: 6
 - Visual Studio Code version: 1.94.2
 - vscode-swift version: 1.11.3

michael-weng:
Verified on e7f81c930e752bb049fc7c5fae8e98684bd8b54d 
Snippet below will no longer display Run/Debug/Coverage buttons
```
import Testing

struct Day01Tests {
    @Test(arguments: [1,3,2]) func part1(_ x: Int){}
}
```

PR #1151, opened 2024-10-21
Don't mark swift-testing test case results as runnable

plemarquand:
When running a swift-testing test that takes arguments the results are added to the Test Explorer showing their pass/fail status. At this time these tests cannot be run with a specific argument so these tests should not be marked as runnable. This prevents the "Run", "Debug" and "Coverage" buttons from appearing on the test items in the explorer.

Issue: #1147

### DIFF
--- a/src/TestExplorer/TestDiscovery.ts
+++ b/src/TestExplorer/TestDiscovery.ts
@@ -231,8 +231,13 @@ export function upsertTestItem(
     // parent tags down the tree as children are upserted.
     testItem = applyTagsToChildren(testItem);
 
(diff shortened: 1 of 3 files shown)


plemarquand:
To verify:
- Create a swift-testing test with argument: `@Test(arguments: [1,2,3]) func test(_ arg: Int) {}`
- Run it
- Check the test explorer and see 1,2,3 as entries in the list, verify that on hover they don't have the buttons to run them (while the parent 'test' entry does.

award999 (approved):
post-merge is there a test we can add here?

Issue #1145, opened 2024-10-17
Can no longer cancel running tests


plemarquand:
I wrote a test that accidentally never completed. The only way to halt the test run in VS Code was to restart the editor. We need a way to stop a test run. The little square stop button in the attached screenshot does nothing.

<img width="356" alt="Screenshot 2024-10-17 at 11 44 46 AM" src="https://github.com/user-attachments/assets/8762324e-90b6-456f-aeee-fc844d9895a1">

adam-fowler:
That used to work

plemarquand:
A clarification, the Cancel button in the Test Explorer _does_ work and stops a build/cancels running tests as you'd expect, but the Cancel button in the list panel to the right of the test results output (see screenshot) is the one that does nothing.

michael-weng:
Verified on e7f81c930e752bb049fc7c5fae8e98684bd8b54d that a test can be properly cancelled.

PR #1153, opened 2024-10-21
Improve Test Cancellation

plemarquand:
There are two ways to cancel a running test, one via the Cancel button at the top of the test explorer, and the other using the Cancel button in the test run tree in the bottom right of the Test Results output panel.

We were handling cancellation via the first method, but the one in the test run tree uses a different cancellation token to indicate a test run should be cancelled. It is is found on the constructed vscode.TestRun.

In our situation we want both of these to do the same thing, cancel the build if its running and stop the test run if its running.

Create a `CompositeCancellationToken` that wraps both these child tokens and performs cancellation if either of them request cancellation.

This approach also centralizes the cancellation logic to the `TestRunProxy`, removing the need to pass the test explorer's token around to the `TestRunner` methods.

This patch also improves cancellation when debugging, ensuring that if you're doing a swift-testing + XCTest debugging run and you cancel during the first test type the second one wont start.

Finally, print a "Test run cancelled" message in the test run output to make it easier to see if a run was cancelled. This is helpful when you're looking through the test results tree at old runs.

Issue: #1145

### DIFF
--- a/src/TestExplorer/TestRunner.ts
+++ b/src/TestExplorer/TestRunner.ts
@@ -39,6 +39,7 @@ import { TestCoverage } from "../coverage/LcovResults";
 import { BuildConfigurationFactory, TestingConfigurationFactory } from "../debugger/buildConfig";
 import { TestKind, isDebugging, isRelease } from "./TestKind";
 import { reduceTestItemChildren } from "./TestUtils";
+import { CompositeCancellationToken } from "../utilities/cancellation";
 
 export enum TestLibrary {
     xctest = "XCTest",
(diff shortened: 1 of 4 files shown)


plemarquand:
Its easiest to verify this with a long running test you can cancel:

```
final class LongRunningTests: XCTestCase {
    func testLongRunning() async throws {
        try await Task.sleep(nanoseconds: NSEC_PER_SEC * 60 * 10)
    }
}
```

Start this test and then hit the cancel button in both the test explorer and the test run list and see if the build/execution cancels as expected.

award999 (approved):

Issue #1154, opened 2024-10-21
Make sure all callbacks in registrerCommand return objects properly


michael-weng:
**Describe the bug**
Found this when implementing (https://github.com/swiftlang/vscode-swift/pull/1144)
function commands.registerCommand(command: string, callback: (...args: any[]) => any, thisArg?: any): vscode.Disposable
As show the callback param can return type any, and if we are not care with implicit return of arrow function expressions it's easy to miss this and as a result execute command will not be able to return the promise as expected.

example:
        vscode.commands.registerCommand(COMMANDS.UseLocalDependency, item => {
            if (item instanceof PackageNode) {
                useLocalDependency(item.name, ctx);
            }
        }),

the above should return the promise returned by useLocalDependency

**Expected behavior**
All executeCommand of registered command properly return the right objects.

**Environment**
 - Impact all platforms

michael-weng:
This will be verified by https://github.com/swiftlang/vscode-swift/pull/1144 passing.

award999:
To resister command callbacks return is propagated but the callback itself may not return anything

michael-weng:
Test checked-in in https://github.com/swiftlang/vscode-swift/pull/1144 and many other tests are now passing in the CI for a while.

Verified.

PR #1155, opened 2024-10-21
Properly return in all registerCommand when not doing implicit return

michael-weng:
- If the exist function return void, just return true in the end after the function complete execution.
- Otherwise explicitly return the object.

Issue: #1154


matthewbastien (changes requested):
Not all of these should be returning false by default, most likely. At least for `runTestMultipleTimes()` it should probably just return `undefined` or an empty array to indicate that no tests actually ran. Whatever makes sense for each individual command depending on what it's trying to do.

Taking a look at the `runTestMultipleTimes()` function itself, it looks like it doesn't actually specify a return type despite returning an array at the end which is a separate issue you might want to resolve with this patch.

michael-weng:
> Not all of these should be returning false by default, most likely. At least for `runTestMultipleTimes()` it should probably just return `undefined` or an empty array to indicate that no tests actually ran. Whatever makes sense for each individual command depending on what it's trying to do.
> 
> Taking a look at the `runTestMultipleTimes()` function itself, it looks like it doesn't actually specify a return type despite returning an array at the end which is a separate issue you might want to resolve with this patch.

Hmm `runTestMultipleTimes()` can also just return void if user provide no input, I am just going to remove return false and reserve undefined for these situations.

plemarquand:
I'm not sure we need to be returning anything here. From what I can tell the return value is not being used in any meaningful way by VS Code. 

I have a feeling the signature is defined with `any` as a return type to make it easier to use existing code as commands without having to modify their signature.

award999:
> Issue: #1154

can we have more info. This is the commit message when merged

award999:
> I'm not sure we need to be returning anything here. From what I can tell the return value is not being used in any meaningful way by VS Code.
> 
> I have a feeling the signature is defined with `any` as a return type to make it easier to use existing code as commands without having to modify their signature.

We may not need it for all commands (promise is less important), but to be able to test these commands using `executeCommand` API, returning say the exit code of a task makes the testing much easier and then we don't have to rely on say the onDidEndTaskProcess event listener or complex mocking. The return value for commands is mostly for testing purposes

award999:
> > Not all of these should be returning false by default, most likely. At least for `runTestMultipleTimes()` it should probably just return `undefined` or an empty array to indicate that no tests actually ran. Whatever makes sense for each individual command depending on what it's trying to do.
> > Taking a look at the `runTestMultipleTimes()` function itself, it looks like it doesn't actually specify a return type despite returning an array at the end which is a separate issue you might want to resolve with this patch.
> 
> Hmm `runTestMultipleTimes()` can also just return void if user provide no input, I am just going to remove return false and reserve undefined for these situations.

typically `undefined` means cancellation so that makes sense when there is no selection. I'd avoid void here

michael-weng:
> > > Not all of these should be returning false by default, most likely. At least for `runTestMultipleTimes()` it should probably just return `undefined` or an empty array to indicate that no tests actually ran. Whatever makes sense for each individual command depending on what it's trying to do.
> > > Taking a look at the `runTestMultipleTimes()` function itself, it looks like it doesn't actually specify a return type despite returning an array at the end which is a separate issue you might want to resolve with this patch.
> > 
> > 
> > Hmm `runTestMultipleTimes()` can also just return void if user provide no input, I am just going to remove return false and reserve undefined for these situations.
> 
> typically `undefined` means cancellation so that makes sense when there is no selection. I'd avoid void here

Sorry I meant return void when current folder is not set vs return false explicit in the callback of registerCommand.

matthewbastien (approved):

michael-weng:
> > Issue: #1154
> 
> can we have more info. This is the commit message when merged

Done

PR #1157, opened 2024-10-22
Add some more unit tests for "filesystem" utility module

award999:
* Lower some tests from integration to the unit level
* Template the expandFilePathTilda function to not mock platform paths
* Add more tests

### DIFF
--- a/src/utilities/filesystem.ts
+++ b/src/utilities/filesystem.ts
@@ -58,9 +58,22 @@ export function isPathInsidePath(subpath: string, parent: string): boolean {
  * @param filepath File path
  * @returns full path
  */
-export function expandFilePathTilda(filepath: string): string {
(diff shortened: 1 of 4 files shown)


src/utilities/filesystem.ts
```diff
@@ -58,9 +58,22 @@ export function isPathInsidePath(subpath: string, parent: string): boolean {
  * @param filepath File path
  * @returns full path
  */
-export function expandFilePathTilda(filepath: string): string {
-    if (process.platform !== "win32" && filepath[0] === "~" && process.env.HOME) {
-        return path.join(process.env.HOME, filepath.slice(1));
+export function expandFilePathTilda(
```
plemarquand, line 61: nit: `tilde` not `tilda`

plemarquand (approved):

Issue #1136, opened 2024-10-10
Update to TypeScript 5.6


matthewbastien:
The latest version of VS Code increased its built-in TypeScript version to 5.6.2 which uncovered an issue in our test code (#1134). We should upgrade to TypeScript 5.6 at the next possible opportunity.

Unfortunately, TypeScript ESLint does not currently support TypeScript 5.6 which means we'll have to wait for typescript-eslint/typescript-eslint#9653 to be completed before we can upgrade.

roulpriya:
Typescript has already been migrated to 5.8.2. Let us close this issue.

plemarquand:
Typescript is still at [5.7.3 in the package.json](https://github.com/swiftlang/vscode-swift/blob/main/package.json#L1604). I'm not sure if its going to be clean upgrade, but its worth a try.

PR #1172, opened 2024-10-28
Don't allow major updates of vsce

award999:
Major update breaks dependabot because require going to node.js 20. Will leave until a later time when we can properly migrate to the next major version of node.js and vsce

https://github.com/swiftlang/vscode-swift/pull/1171

Issue: #1136


award999:

PR #1181, opened 2024-10-30
Documentation Preview Editor

matthewbastien:
Adds a new command called `Swift: Preview Documentation` that will open a new documentation preview editor powered by SourceKit-LSP. This command is only enabled if SourceKit-LSP supports converting documentation.


assets/test/SlothCreatorBuildingDocCDocumentationInXcode/Sources/SlothCreator/Care/Activity.swift
```diff
@@ -0,0 +1,40 @@
+/*
```
plemarquand, line 1: nit: this should probably be our actual licence header for the project.

also nit, but I can't comment on the file, but delete the file `assets/test/SlothCreatorBuildingDocCDocumentationInXcode/Sources/SlothCreator/.DS_Store`

matthewbastien: I just removed the license header to be in line with what we normally do for test assets. The .DS_Store file has also been removed and added to the .gitignore for the project.

src/documentation/DocumentationPreviewEditor.ts
```diff
@@ -0,0 +1,166 @@
...
+                    vscode.Uri.file(swiftDoccRenderPath),
+                    ...context.folders.map(f => f.folder),
+                ],
+            }
+        );
+        const webviewBaseURI = this.webviewPanel.webview.asWebviewUri(
+            vscode.Uri.file(swiftDoccRenderPath)
+        );
+        const scriptURI = this.webviewPanel.webview.asWebviewUri(
+            vscode.Uri.file(this.extension.asAbsolutePath("assets/documentation-webview/index.js"))
```
award999, line 58: nit: path.join like below

src/documentation/webview/CommunicationBridge.ts
```diff
@@ -0,0 +1,120 @@
...
+export function createCommunicationBridge(): Promise<CommunicationBridge> {
+    if ("webkit" in window) {
+        throw new Error("A CommunicationBridge has already been established");
+    }
+
+    return new Promise<CommunicationBridge>((resolve, reject) => {
+        try {
+            // Define the window.webkit property in order to receive messages
+            const messageHandlers: Set<(message: VueAppMessage) => void> = new Set();
+            Object.defineProperty(window, "webkit", {
```
award999, line 42: assume the API type is referenced in the swift-docc-renderer project? Maybe good to leave a comment with the link so we know where to start looking if this changes

src/documentation/webview/CommunicationBridge.ts
```diff
@@ -0,0 +1,120 @@
...
+                            },
+                        },
+                    },
+                },
+                writable: false,
+            });
+
+            // Wait for the window.bridge property to be set in order to send messages
+            let windowBridge: unknown;
+            Object.defineProperty(window, "bridge", {
```
award999, line 57: same

src/documentation/webview/webview.ts
```diff
@@ -0,0 +1,123 @@
...
+        } else {
+            // Show the error message
+        }
+    }
+
+    // Notify vscode-swift that we're ready to receive messages
+    vscode.postMessage({ type: "loaded" });
+});
+
+declare global {
```
award999, line 87: other webviews we do could use this so we could move it to a common place

src/documentation/DocumentationPreviewEditor.ts
award999: also is the build placing this script under assets? Is there a reason we cannot place this and swift-docc-renderer under dist like rest of build artifacts so we only have the one build folder.

award999 (commented):
nothing major, looks good

src/documentation/DocumentationPreviewEditor.ts
matthewbastien: Done.

The build does place the script into assets because it needs to be bundled with esbuild to work. The TypeScript compiler doesn't get rid of imports and those don't work in webviews.

src/documentation/webview/CommunicationBridge.ts
matthewbastien: Good point. I added a note to the docstring for createCommunicationBridge to make it clear where the value comes from.

src/documentation/webview/CommunicationBridge.ts
matthewbastien: done

src/documentation/webview/webview.ts
matthewbastien: I agree, but I think we should wait until we actually have another webview first.

src/documentation/DocumentationPreviewEditor.ts
award999: ok, but to confirm, esbuild is putting this under dist/ when building?

award999 (approved):

src/documentation/DocumentationPreviewEditor.ts
matthewbastien: No, it doesn't, unfortunately. This is due to some weirdness in our build where we use `tsc` for local development and `esbuild` for production. I will fix this in a subsequent PR.

Issue #1184, opened 2024-10-30
Integration Tests: Build Commands


michael-weng:
Add integration tests that reflect the workflow of user calling the Swift: Run Build/Clean Build/Debug Build commands.

michael-weng:
Done and passing in CI.

PR #1126, opened 2024-10-07
Add tests for the Swift snippets workflows

award999:
* Created some common utilities for dealing with debug sessions
* Add integration test for run command and debug variant, big difference
  being whether the breakpoint is hit or not

Issue: #1125

PR #1185, opened 2024-10-30
Add integration tests for build commands

michael-weng:
- Validate the workflow of user calling the Swift: Run Build/Clean Build/Debug Build commands.
- Ensure Swift: Run Build will not get blocked by pre-set breakpoint.
- Ensure Swift: Clean Build will result in a cleaned up .build folder.
- Ensure Swift: Debug Build will stop on a breakpoint and resume.

Issue: #1184

### DIFF
--- a/test/integration-tests/commands/build.test.ts
+++ b/test/integration-tests/commands/build.test.ts
@@ -1,0 +1,78 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2024 the VS Code Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of VS Code Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import * as vscode from "vscode";
+import * as fs from "fs";
+import * as path from "path";
+import { expect } from "chai";
+import { folderContextPromise, globalWorkspaceContextPromise } from "../extension.test";
+import { waitForNoRunningTasks } from "../../utilities";
+import { testAssetUri } from "../../fixtures";
+import { FolderContext } from "../../../src/FolderContext";
+import { WorkspaceContext } from "../../../src/WorkspaceContext";
+import { Commands } from "../../../src/commands";
+import { makeDebugConfigurations } from "../../../src/debugger/launch";
+
+suite("Build Commands", function () {
+    let folderContext: FolderContext;
+    let workspaceContext: WorkspaceContext;
+    const uri = testAssetUri("defaultPackage/Sources/PackageExe/main.swift");
+    const breakpoints = [
+        new vscode.SourceBreakpoint(new vscode.Location(uri, new vscode.Position(2, 0))),
+    ];
+
+    suiteSetup(async function () {
+        workspaceContext = await globalWorkspaceContextPromise;
+        await waitForNoRunningTasks();
+        folderContext = await folderContextPromise("defaultPackage");
+        await workspaceContext.focusFolder(folderContext);
+        await vscode.window.showTextDocument(uri);
+        makeDebugConfigurations(folderContext, undefined, true);
+    });
+
+    suiteTeardown(async () => {
+        await vscode.commands.executeCommand("workbench.action.closeAllEditors");
+    });
+
+    test("Swift: Run Build", async () => {
+        // A breakpoint will have not effect on the Run command.
+        vscode.debug.addBreakpoints(breakpoints);
+
+        const result = await vscode.commands.executeCommand(Commands.RUN);
+        expect(result).to.be.true;
+
+        vscode.debug.removeBreakpoints(breakpoints);
+    });
+
+    test("Swift: Clean Build", async () => {
+        const buildPath = path.join(folderContext.folder.fsPath, ".build");
+        const beforeItemCount = fs.readdirSync(buildPath).length;
+
+        const result = await vscode.commands.executeCommand(Commands.CLEAN_BUILD);
+        expect(result).to.be.true;
+
+        const afterItemCount = fs.readdirSync(buildPath).length;
+        expect(afterItemCount).to.be.lessThan(beforeItemCount);
+    });
+
+    test("Swift: Debug Build", async () => {
+        vscode.debug.addBreakpoints(breakpoints);
+
+        const result = vscode.commands.executeCommand(Commands.DEBUG);
+        expect(result).to.eventually.be.true;
+
+        await vscode.commands.executeCommand("workbench.action.debug.continue");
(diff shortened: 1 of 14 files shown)


test/integration-tests/commands/build.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    const breakpoints = [
+        new vscode.SourceBreakpoint(new vscode.Location(uri, new vscode.Position(2, 0))),
+    ];
+
+    suiteSetup(async function () {
+        workspaceContext = await globalWorkspaceContextPromise;
+        await waitForNoRunningTasks();
+        folderContext = await folderContextPromise("defaultPackage");
+        await workspaceContext.focusFolder(folderContext);
+        await vscode.window.showTextDocument(uri);
```
award999, line 40: `showTextDocument` usually doesn't take a URI the pattern is usually:
```
const document = await vscode.workspace.openTextDocument(cUri);
await vscode.window.showTextDocument(document);
```

award999: surprised there wasn't some tsc error

test/integration-tests/commands/build.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        makeDebugConfigurations(folderContext, undefined, true);
+    });
+
+    suiteTeardown(async () => {
+        await vscode.commands.executeCommand("workbench.action.closeAllEditors");
```
award999, line 45: since we do this in a few other places, we can have some enum under the test/ folder with the vscode command IDs we use. Similar to the Commands enum you created

test/integration-tests/commands/build.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        expect(afterItemCount).to.be.lessThan(beforeItemCount);
+    });
+
+    test("Swift: Debug Build", async () => {
+        vscode.debug.addBreakpoints(breakpoints);
+
+        const result = vscode.commands.executeCommand(Commands.DEBUG);
+        expect(result).to.eventually.be.true;
+
+        await vscode.commands.executeCommand("workbench.action.debug.continue");
```
award999, line 75: same can make common

award999: we haven't confirmed if a breakpoint is hit

test/integration-tests/commands/build.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        const beforeItemCount = fs.readdirSync(buildPath).length;
+
+        const result = await vscode.commands.executeCommand(Commands.CLEAN_BUILD);
+        expect(result).to.be.true;
+
+        const afterItemCount = fs.readdirSync(buildPath).length;
+        expect(afterItemCount).to.be.lessThan(beforeItemCount);
+    });
+
+    test("Swift: Debug Build", async () => {
```
award999, line 69: if we follow the same file name pattern, this would be under launch.test.ts

test/integration-tests/commands/build.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    test("Swift: Clean Build", async () => {
+        const buildPath = path.join(folderContext.folder.fsPath, ".build");
+        const beforeItemCount = fs.readdirSync(buildPath).length;
+
+        const result = await vscode.commands.executeCommand(Commands.CLEAN_BUILD);
+        expect(result).to.be.true;
+
+        const afterItemCount = fs.readdirSync(buildPath).length;
+        expect(afterItemCount).to.be.lessThan(beforeItemCount);
```
award999, line 66: would this always be the case? If there is an explanation can you add some comment code to explain

test/integration-tests/commands/build.test.ts
award999: Checkout https://github.com/swiftlang/vscode-swift/pull/1126/files#diff-8df4c1aa97a833b0df73f92de09501c686ab0a49369087ad2c6c3bf59db6bedfR72

when we get stacktrace message from debug adapter, we know a breakpoint is hit

test/integration-tests/commands/build.test.ts
award999: also since this is exercising debugger it is a smoke test, lets setup that suite

award999 (changes requested):

test/integration-tests/commands/build.test.ts
michael-weng: There wasn't a tsc error because this is a valid overload:

/**
 * A short-hand for `openTextDocument(uri).then(document => showTextDocument(document, options))`.
 *
 * @see {@link workspace.openTextDocument}
 *
 * @param uri A resource identifier.
 * @param options {@link TextDocumentShowOptions Editor options} to configure the behavior of showing the {@link TextEditor editor}.
 * @returns A promise that resolves to an {@link TextEditor editor}.
 */
export function showTextDocument(uri: Uri, options?: TextDocumentShowOptions): Thenable<TextEditor>;

test/integration-tests/commands/build.test.ts
michael-weng: Done.

test/integration-tests/commands/build.test.ts
michael-weng: Done.

test/integration-tests/commands/build.test.ts
michael-weng: Done. I will take the piece I need if you don't mind.

test/integration-tests/commands/build.test.ts
michael-weng: > if we follow the same file name pattern, this would be under launch.test.ts

No, the corresponding commands are under src/commands/build.ts so this is correct name.

test/integration-tests/commands/build.test.ts
```diff
@@ -0,0 +1,91 @@
...
+        const bpPromise = waitForDebugAdapterCommand(
+            "Debug PackageExe (defaultPackage)",
+            "stackTrace",
+            workspaceContext
+        );
+
+        const result = vscode.commands.executeCommand(Commands.DEBUG);
+        expect(result).to.eventually.be.true;
+
+        await bpPromise.then(() => continueSession());
```
award999, line 88: if we're hitting breakpoints, crossing into smoke level, let's create a smoke suite and put that there

award999 (commented):
otherwise looks good

test/integration-tests/commands/build.test.ts
award999: still should split into separate suite. Is it actually slow or just trying to keep out of CI build?

michael-weng: What are we trying to achieve to split this into a separate suite? Is it for dependency management?

Given the test run fast and this is a command that should probably always work, I am not sure if we want keep this out of CI.

award999: will leave this for until long term approach is decided

award999 (approved):

PR #1191, opened 2024-11-02
Support Swift SDKs

kabiroberai:
This PR adds support for Swift SDKs when using the SwiftPM build system, as defined by [SE-0387](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0387-cross-compilation-destinations.md). These are already supported by SourceKit LSP, so adding support here is primarily a matter of plumbing the `swift.swiftSDK` parameter along to the `swiftPM.swiftSDK` field of SourceKit LSP's startup options.

~~Additionally, with https://github.com/swiftlang/swift-package-manager/pull/6828 it is now possible to select Darwin triples as Swift SDKs on macOS — that is, you can build for iOS with `swift build --swift-sdk arm64-apple-ios` and so on. This support also applies to SourceKit LSP, and so we can update the "Select Target Platform" action to pick a Swift SDK instead.~~

## Open questions

- ~~This PR _mostly_ obsoletes the old `swift.SDK` parameter, but I haven't removed it yet because 1) Swift SDK support only works with SwiftPM right now and 2) I'm not sure whether removing `swift.SDK` would break the Windows-specific use cases for that parameter. Removing it would clean up a lot of code though. Would this be reasonable to do?~~
- ~~Support for Darwin SDKs as Swift SDKs is on main and has landed in time for the 6.1 release cut, but doesn't exist in 6.0. Is there a recommended pattern we can use for feature detection to only surface the "Select Target Platform" command if the compiler is new enough?~~

TODO:
- [x] Tests


stevapple:
We need a https://github.com/swiftlang/swift-package-manager/pull/6828 like solution on Windows before we can get rid of `swift.SDK`…

award999:
@kabiroberai can you rebase please? We recently moved to GitHub actions to verify PRs

kabiroberai:
@award999 done!

kabiroberai:
Heya @award999, any updates on this? CI is green now ✅

daveyc123:
Thanks for the contribution @kabiroberai!

In answer to your questions:

> This PR mostly obsoletes the old swift.SDK parameter, but I haven't removed it yet because 1) Swift SDK support only works with SwiftPM right now and 2) I'm not sure whether removing swift.SDK would break the Windows-specific use cases for that parameter. Removing it would clean up a lot of code though. Would this be reasonable to do?
As long as SwiftPM continues to support the --sdk option, the old swift.SDK parameter should stick around.

> Support for Darwin SDKs as Swift SDKs is on main and has landed in time for the 6.1 release cut, but doesn't exist in 6.0. Is there a recommended pattern we can use for feature detection to only surface the "Select Target Platform" command if the compiler is new enough?
The extension typically follows the pattern that the command is always enabled. If the command can't run, it presents an errorMessage. This allows presenting to the user the reason why the command couldn't be run (as opposed to it just not being there). For example "You don't have an SDK selected" vs. "This feature is only supported on Swift 6.0 and higher".

An example of this behaviour can be found at https://github.com/swiftlang/vscode-swift/blob/c3e87fde3119be3a0f712096b1a7e133c91da84f/src/commands/createNewProject.ts#L40

award999:
@kabiroberai looks like there is a merge conflict to resolve

kabiroberai:
@award999 fixed!

award999:
@matthewbastien when you're back can you drive this code review being the most familiar with toolchain/sdk selection?

matthewbastien (approved):
Looks great! Thank you for creating this PR @kabiroberai!

PR #1193, opened 2024-11-03
feat: Use toolchain's swift-format as a formatting provider

### LINKED ISSUE #1190: Formatter situation?
What are the plans about providing a formatter in this repo now that `swift-format` is bundled with the toolchains?

Currently it this extension does show up as a formatter but apparently it just triggers prettier?

This would look like a bit unfair for `SwiftFormat` (Nick Lockwood's) which is what we currently use as well since `swift-format` can be hard to bear with at times ... but it still makes sense since `swift-format` is part of the toolchain.

MahdiBM:
resolves #1190 

Inspired by https://github.com/vknabel/vscode-apple-swift-format

Checklist:
* [x] Create and register a formatting provider that calls `swift format`.
* [ ] Test the PR locally.
* [ ] Proper error handling.
* [ ] Make sure user toolchain has swift-format bundled with it.

I'm not a typescript dev, maintainer folks you can feel free to show the obvious problems, or even nit pick.

### DIFF
--- a/src/WorkspaceContext.ts
+++ b/src/WorkspaceContext.ts
@@ -33,6 +33,7 @@ import { DebugAdapter, LaunchConfigType } from "./debugger/debugAdapter";
 import { SwiftBuildStatus } from "./ui/SwiftBuildStatus";
 import { SwiftToolchain } from "./toolchain/toolchain";
 import { DiagnosticsManager } from "./DiagnosticsManager";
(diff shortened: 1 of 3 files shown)


MahdiBM:
Looks like sourcekit-lsp is the proper place for this implementation to go to: https://github.com/swiftlang/vscode-swift/issues/1190#issuecomment-2457694387

PR #1195, opened 2024-11-04
Setup Windows GH actions job

award999:
We cannot use the existing windows job setup because it launches a docker container which we cannot open a VS Code window in to run the tests as no tool like Xvfb exists for Windows

Fix up some tests that fail on windows (mostly related to path separator mismatches)


.github/workflows/pull_request.yml
```diff
@@ -37,3 +37,33 @@ jobs:
...
       linux_build_command: ./docker/test.sh
       enable_windows_checks: false
+
+  tests-windows:
+    name: Windows (${{ matrix.swift_version }})
+    runs-on: ${{ contains(matrix.swift_version, 'nightly') && 'windows-2019' || 'windows-2022' }}
+    strategy:
+      fail-fast: false
+      matrix:
+        swift_version: ['5.9', '5.10', '6.0', 'nightly', 'nightly-6.0']
```
plemarquand, line 47: probably want to exclude nightly from the list as we do on linux.

award999:
@swift-ci Please test

src/TestExplorer/TestRunner.ts
```diff
@@ -40,6 +40,8 @@ import { BuildConfigurationFactory, TestingConfigurationFactory } from "../debug
 import { TestKind, isDebugging, isRelease } from "./TestKind";
 import { reduceTestItemChildren } from "./TestUtils";
 import { CompositeCancellationToken } from "../utilities/cancellation";
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+import stripAnsi = require("strip-ansi");
```
matthewbastien, line 44: nit: I think we have ES module support enabled in our TS config which should allow you to do something like:
```typescript
import * as stripAnsi from "strip-ansi";
```
Either that or disable the no-require-imports rule in our eslint config. I'm not a fan of adding es lint ignores everywhere.

src/commands/build.ts
```diff
@@ -74,16 +74,26 @@ export async function debugBuildWithOptions(
 ) {
     const current = ctx.currentFolder;
     if (!current) {
+        console.log("debugBuildWithOptions: No current folder on WorkspaceContext");
```
matthewbastien, line 77: We should be using the output channel to log and not console. Do we not have the no-console lint rule enabled?

matthewbastien (approved):
I just have a few minor things I'd like fixed, but I'm fine with fixing them in a future PR since this gets Windows CI up and running.

src/TestExplorer/TestRunner.ts
plemarquand: We don't have esModuleInterop set in our tsconfig.json. I think we could try turning that on, but I'd like to do it in a follow on patch.

src/commands/build.ts
plemarquand: Done

src/TestExplorer/TestRunner.ts
matthewbastien: Whoops, thought we already had that enabled. Yeah I think that's best turned on in another patch.

Issue #1197, opened 2024-11-04
SwiftOutputChannel test prints lots of errors


award999:
With latest vscode there are a lot of error messages printed to the console:

```
  SwiftOutputChannel
    ✔ Appends logs
Error: Trying to add a disposable to a DisposableStore that has already been disposed of. The added object will be leaked!
    ✔ Appends lines
    at H3.add (file:///d:/a/vscode-swift/vscode-swift/.vscode-test/vscode-win32-x64-archive-1.95.1/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:24:1439)
    ✔ Appends lines and rolls over
    at yw.n (file:///d:/a/vscode-swift/vscode-swift/.vscode-test/vscode-win32-x64-archive-1.95.1/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:120:1748)
    ✔ Appends and rolls over

```

michael-weng:
https://github.com/swiftlang/vscode-swift/actions/runs/12149050931/job/33879420980

Verified no excessive log output unless on test failure.

PR #1200, opened 2024-11-05
Fix warnings printed during SwiftOutputChannel tests

plemarquand:
Output channels were not designed to be created and immediately disposed because they're added to the disposable store internal to VS Code asynchronously.

Issue #1197


award999 (approved):
LGTM

Issue #1205, opened 2024-11-13
Integration Tests: Inline/Expand Macro


michael-weng:
Add integration tests that verify macro action integration with Sourcekit-lsp.
- Similar to https://github.com/swiftlang/vscode-swift/issues/1235, ensure the integration with sourcekit-lsp works as intended, with correct requests sent and responses received.
- Detect potential changes in the LSP response for expand macro feature, which requires extra code in the extension to display the code and position in peek view.

award999:
@michael-weng can you add the plan to the description of what will be tested

michael-weng:
> @michael-weng can you add the plan to the description of what will be tested

Done.

PR #1206, opened 2024-11-14
Add integration tests for Expand Macro actions

michael-weng:
- Validate the workflow of user calling Expand Macro actions on a swift project with macro
- Add test fixture for swift macro
- Verify expand macro by asserting expanded macro document contain the right macro

Issue: #1205


michael-weng:
We can probably optimize with background compilation integration test, but let's do that after this is merged and we have a few more lsp related integration tests.

Issue #1207, opened 2024-11-15
Running debug from TestExplorer target (XCTest) TestItem runs nothing


adam-fowler:
**Describe the bug**
With a project using XCTest, running `Debug Test` from the TestExplorer target TestItem runs 0 tests
 
**To Reproduce**
1. Open project using XCTest for testing
2. Open TestExplorer
3. Right click on TestItem for Test target and select `Debug Test`.
4. Check output in TestResults pane
5. See `Executed 0 tests, with 0 failures (0 unexpected) in 0.000 (0.001) seconds`

**Expected behavior**
All tests in target are run

**Environment**
 - macOS: 15.0.1
 - Swift version: Apple Swift version 6.0.2 (swiftlang-6.0.2.1.2 clang-1600.0.26.4)
 - Visual Studio Code version: 1.95.3
 - vscode-swift version: 1.11.3

**Additional context**
`xctest` is run with command line parameter `-XCTest testTargetName.*`. This won't work. You need to list all the classes in the target to get them to run eg
```
xctest -XCTest testTargetName.testClass1,testTargetName.testClass2 ...
```

plemarquand:
This is certainly a valid issue, but I'm wondering how to best handle it on Windows where the command line max length is only 8,191 characters. It's quite easy to exceed this limit when debugging all tests in an entire target.

adam-fowler:
You'd have to have a lot of TestCase classes to exceed the max length.

PR #1209, opened 2024-11-16
Fix debugging entire XCTest targets

plemarquand:
When right clicking a test target and debugging the tests, no XCTests are run.

The test run argument simplification code would simplify too much when debugging a test target; the `xctest` binary does not accept a test target as an `-XCTest` argument. Only fully specified test names or test suites are permitted.

Issue: #1207


award999 (approved):
LGTM

matthewbastien (approved):

PR #1227, opened 2024-11-27
Reorganize LSP extensions

matthewbastien:
Organize our LSP extensions in the same way that vscode-languageclient does: by using a namespace that includes all of the relevant information related to the Request/Notification. This allows us to use our LSP extensions in the same type-safe manor as we do with built-ins.


src/sourcekit-lsp/LanguageClientFactory.ts
```diff
@@ -0,0 +1,41 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import { LanguageClient, LanguageClientOptions, ServerOptions } from "vscode-languageclient/node";
+
+/**
+ * Used to create a {@link LanguageClient} for use in VS Code.
+ *
+ * This is primarily used to make unit testing easier so that we don't have to
```
award999, line 20: awesome :)

src/sourcekit-lsp/extensions/GetReferenceDocumentRequest.ts
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * ### LSP Extension
+ *
+ * This request is an extension to LSP supported by SourceKit-LSP.
+ *
+ * Enable the experimental client capability `"workspace/getReferenceDocument"` so that the server responds with
+ * reference document URLs for certain requests or commands whenever possible.
+ */
+export namespace GetReferenceDocumentRequest {
+    export const method = "workspace/getReferenceDocument" as const;
```
award999, line 48: "get" doesn't seemed to be used elsewhere in the LSP protocol. I guess sort of implied when there is a response

src/sourcekit-lsp/extensions/index.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Users should not need to rely on this request. The index should always be updated automatically in the background.
+ * Having to invoke this request means there is a bug in SourceKit-LSP's automatic re-indexing. It does, however, offer
+ * a workaround to re-index files when such a bug occurs where otherwise there would be no workaround.
+ *
+ * ### LSP Extension
+ *
+ * This request is an extension to LSP supported by SourceKit-LSP.
+ */
+export namespace ReIndexProjectRequest {
```
award999, line 40: should this be in its own file too?

award999 (commented):
otherwise looks good

src/sourcekit-lsp/extensions/index.ts
matthewbastien: Good point. Done.

src/sourcekit-lsp/extensions/GetReferenceDocumentRequest.ts
matthewbastien: Yeah, I'm just matching the name that SourceKit-LSP gives it here.

award999 (approved):

Issue #562, opened 2023-07-02
Add Swift Documentation Preview Support


Kyle-Ye:
**Is your feature request related to a problem? Please describe.**

When user actives documentation preview panel and is currently on a .swift, .h or .markdown file, preview the documentation on the right panel.

**Describe the solution you'd like**

It can be achieved by using the existing ConvertService API provided by [DocC](https://github.com/apple/swift-docc).

No Xcode magic is required.

> https://twitter.com/franklinschrans/status/1666343700442628098

**Benefits**

Improve the doc-writing experience on non-Xcode IDE.

**Additional context**

Inspiration: Xcode 15's documentation preview feature

See more info https://developer.apple.com/videos/play/wwdc2023/10244/

adam-fowler:
You can use the command `Markdown: Open preview to the side` to give a markdown preview. This won't give you the full docc preview obviously but it is a start.

Kyle-Ye:
> You can use the command `Markdown: Open preview to the side` to give a markdown preview. This won't give you the full docc preview obviously but it is a start.

Yeah. VSCode has a build-in Markdown feature panel.

But I'd like to know if it was possible to render a HTML directly in a associated-panel(The same as Markdown file and Markdown-Preview panel) via the VSCode plugin SDK.

adam-fowler:
You can do it. But managing the preview server, associating the current page with the correct page on the server isn't simple. It is all possible, just complex. 

I have looked into to it a bit and got it displaying a page served by preview server I was running separately. But it is far away from a finished feature.

roydbt:
Any updates on this issue?

adam-fowler:
No update on this. I haven't had anymore time to look into it.

daveyc123:
@adam-fowler Can we grab this one? If so, do you have your prototype changes in a fork somewhere?

adam-fowler:
@daveyc123 its all yours. I had some basic code setup to display the documentation but there was a loads of issues managing the preview server and updating the documentation. Don't think I kept any of the code around though.

adam-fowler:
@daveyc123 I found this https://github.com/adam-fowler/vscode-swift/tree/documentation-preview

matthewbastien:
I just posted a general overview of how this feature could be implemented on the Swift forums: https://forums.swift.org/t/rfc-swift-docc-integration/75537

Please check it out and leave any feedback/suggestions that you might have!

matthewbastien:
[...]

PR #1229, opened 2024-11-27
Add a custom build for swift-docc-render

matthewbastien:
We need to build a special copy of [swiftlang/swift-docc-render](https://github.com/swiftlang/swift-docc-render) that will work in VS Code. However, this build does not need to happen every single time we build the extension. So, as a nice middle-ground I've opted to create a NPM script called `update-swift-docc-render` that will:

1. clone `swift-docc-render` into a temporary directory
2. apply our custom patches to `swift-docc-render`
3. build a copy of `swift-docc-render` into `assets/swift-docc-render`

The asset will be checked into source control so that we don't have to build it unless `swift-docc-render` changes in the future.


matthewbastien:
It's also worth noting that building `swift-docc-render` requires node v18 which is a couple versions behind the version that VS Code uses.

matthewbastien:
Needed for #562

award999:
@swift-ci Please test

plemarquand (approved):

Issue #1230, opened 2024-11-27
Test Plan Implementation - Integration Tests: swift-plugin


michael-weng:
We already have integration tests for the task providers, but just need one to cover the user facing commands.

What we want to test for:
- swift.runPluginTask shows up and can be executed by user when under the right command_plugin file
- swift.runPluginTask command return appropriate tasks from different sources (provided by extension or resolved from tasks.json)

award999:
@michael-weng can you add the plan of what needs to be tested

michael-weng:
> @michael-weng can you add the plan of what needs to be tested

Done.

PR #1231, opened 2024-11-27
Add integration test for swift.runPluginTask command

michael-weng:
- Change runPluginTask to return the argument filter for better testability
- Given SwiftPluginTaskProvider integration test already test the tasks, assert the command arguments are valid and return the right tasks for picking.

Issue: #1230


award999:
Superseded by #1256

Issue #1235, opened 2024-12-02
Test Plan Implementation - Integration Tests: lsp integration


michael-weng:
* Ensure the integration with sourcekit-lsp works as intended, with correct requests sent and responses received.
* Detect potential changes in the LSP response for most essential request like go to definition and find all references.

PR #1236, opened 2024-12-02
Add basic lsp integration test to check language support

michael-weng:
- Roll relevant test under language folder for better organization.
- Given how build all task for defaultPacakge will need to be executed as a set up for this test, remove @slow for other tests that's running this task as the cost will be atmortized in CI.
- The test check lsp client is running and exercise Goto Definition Request and Find References Request.

Issue: #1235


award999:
Superseded by #1259

Issue #1087, opened 2024-09-20
Launch Swift extension if workspace has `buildServer.json` or `compile_flags.txt`


ahoppen:
The VS Code extension currently check for the presence of `Package.swift` or `compile_commands.json` in `isValidWorkspaceFolder`. SourcKit-LSP can also handle `compile_flags.txt` (https://clang.llvm.org/docs/JSONCompilationDatabase.html#alternatives) or files that specify a build server by having `buildServer.json` in their project root. Such projects should also be considered valid workspaces, I think.

aelam:
@ahoppen 
if bsp is enabled i.e. buildServer.json
should some bsp request be sent from the client?

ahoppen:
If `buildServer.json` exists in the workspace root, SourceKit-LSP will launch the BSP server and send the initialize request to it. I don’t know what you mean by *client* here. The editor should not send any request to the BSP server.

aelam:
@ahoppen 
Thanks for replying.

I checked the `xcode-build-server`, only `build/initialize` and `build/initialized` are sent. 
I'm wondering how `workspace/buildTargets` is sent. is it sent from client. (i.e. `vscode-swift` if it supports bsp)

* [build/initialize](https://github.com/swiftlang/sourcekit-lsp/blob/87b928540200708a198d829c4ad1bac37b1a5d69/Contributor%20Documentation/BSP%20Extensions.md#buildinitialize)
* build/initialized
* [workspace/buildTargets
](https://github.com/swiftlang/sourcekit-lsp/blob/87b928540200708a198d829c4ad1bac37b1a5d69/Contributor%20Documentation/BSP%20Extensions.md#workspacebuildtargets)

ahoppen:
This is because we changed the BSP integration between 6.0 and 6.1 (https://forums.swift.org/t/extending-functionality-of-build-server-protocol-with-sourcekit-lsp/74400 has the details). `xcode-build-server` is still using the old integration that uses `textDocument/registerForChanges` and `build/sourceKitOptionsChanged` instead of `workspace/buildTargets`, to the best of my knowledge. If the BSP server returns `data.sourceKitOptionsProvider: true` in the initialize request, we switch to the “new” integration using `workspace/buildTargets`. https://github.com/swiftlang/sourcekit-lsp/blob/main/Contributor%20Documentation/Implementing%20a%20BSP%20server.md has the details on how to implement a BSP server for SourceKit-LSP.

[...]

PR #1240, opened 2024-12-03
Activate extension when compile_flags.txt or buildServer.json is present

plemarquand:
SourceKit-LSP can handle projects configured with the paired down clang `compile_flags.txt` configuration file, as well as projects configured with the Build Server Protocol's `buildServer.json`.

Activate the extension if the folder added to the workspace contains either of these files in the root.

Issue: #1087


award999 (approved):
Looks great

PR #1244, opened 2024-12-04
Prompt to restart extension on `xcode-select`

plemarquand:
If the user performs an `xcode-select` and doesn't have a `swift.path` set in their VS Code settings, prompt them to restart the extension to pick up the new Xcode path. This only applies on macOS where `xcode-select` is possible.

### DIFF
--- a/src/toolchain/SelectedXcodeWatcher.ts
+++ b/src/toolchain/SelectedXcodeWatcher.ts
@@ -1,0 +1,93 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2021-2024 the VS Code Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of VS Code Swift project authors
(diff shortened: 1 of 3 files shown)


src/toolchain/SelectedXcodeWatcher.ts
```diff
@@ -0,0 +1,93 @@
...
+        clearInterval(this.interval);
+    }
+
+    /**
+     * Polls the Xcode symlink location checking if it has changed.
+     * If the user has `swift.path` set in their settings this check is skipped.
+     */
+    private async setup() {
+        this.xcodePath = await this.xcodeSymlink();
+        this.interval = setInterval(async () => {
```
award999, line 69: node.js has file watching APIs https://nodejs.org/docs/latest/api/fs.html#fspromiseswatchfilename-options, I would assume updating the symlink should trigger the watcher

plemarquand: I looked at this initially but it doesn't notify when the target of a symlink changes, only when the symlink at the supplied path is deleted/renamed

award999 (approved):

PR #1271, opened 2024-12-12
Allow triggering nightly build from PR label

plemarquand:
Adding the `full-test-run` label to a PR before creating it will trigger the same workflow as the nightly job.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -30,6 +30,30 @@ jobs:
       windows_build_command: docker\test-windows.ps1
       enable_windows_docker: false
 
+  full_test_run:
+    name: Full Test Run
+    if: contains(github.event.pull_request.labels.*.name, 'full-test-run')
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@main
+    with:
+      # Linux
+      linux_exclude_swift_versions: '[{"swift_version": "nightly-6.0"},{"swift_version": "nightly-main"}]'
+      linux_env_vars: |
+        NODE_VERSION=v18.19.0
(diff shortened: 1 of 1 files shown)


plemarquand:
I'd love to be able to just call the `tests_release` job in the nightly workflow to avoid duplication, but I couldn't figure out a nice way to do it.

.github/workflows/pull_request.yml
```diff
@@ -30,6 +30,30 @@ jobs:
...
+    if: contains(github.event.pull_request.labels.*.name, 'full-test-run')
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@main
+    with:
+      # Linux
+      linux_exclude_swift_versions: '[{"swift_version": "nightly-6.0"},{"swift_version": "nightly-main"}]'
+      linux_env_vars: |
+        NODE_VERSION=v18.19.0
+        NODE_PATH=/usr/local/nvm/versions/node/v18.19.0/bin
+        NVM_DIR=/usr/local/nvm
+        CI=1
```
award999, line 45: is the main different that not setting FAST_TEST_RUN? If so I think we could reuse the job above and do something along lines of
FAST_TEST_RUN=${{ contains(github.event.pull_request.labels.*.name, 'full-test-run') && '0' || '1'}}

See https://github.com/actions/runner/issues/409#issuecomment-752775072

plemarquand: Excellent idea

plemarquand:
Since some nightly tests are still failing this PR won't show green CI, but the fact that the nightly tests are running with this label shows its working

award999 (approved):
LGTM

PR #1276, opened 2024-12-13
Only change scheme of `.swiftinterface` files to `readonly` if old scheme was `file`

ahoppen:
I’m working on adding semantic functionality to generated interfaces in SourceKit-LSP and this will generate reference documents with a `sourcekit-lsp:` scheme and a `.swiftinterface` file. We shouldn’t change those URIs to a `readonly` scheme.


plemarquand:
What happens if the user modifies the contents of a document with the `sourcekit-lsp:` scheme?

ahoppen:
Non-file URL schemes are automatically marked as non-editable by VS Code.

plemarquand (approved):

Issue #1269, opened 2024-12-12
Build Commands suite is flaky


award999:
Flaking in nightly build https://github.com/award999/vscode-swift/actions/runs/12287793290/job/34290458188
```
  2) Build Commands
       Swift: Debug Build @slow:
     Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/tmp/code/dist/test/integration-tests/commands/build.test.js)
```

rbenegal:
Confirmed that the `Build Commands` tests now pass successfully locally and on github actions. It does seem like they spew out a whole bench of errors which may be confusing, possibly we can clean this up them later.


<details>
  <summary>
    Output from a local run of build.test.ts on linux
  </summary>

```
 swift -v
Swift version 6.0.3 (swift-6.0.3-RELEASE)
```
on  `vscode 1.96.4` with `Ubuntu 22.04`

build.test.ts:
```
npm test -- --label integrationTests

> swift-lang@1.11.4 pretest
> npm run compile-tests


> swift-lang@1.11.4 compile-tests
> del-cli ./assets/test/**/.build && npm run compile


> swift-lang@1.11.4 compile
> del-cli ./dist/ && tsc --build


> swift-lang@1.11.4 test
> vscode-test --label integrationTests

✔ Validated version: 1.96.4
✔ Found existing install in /home/rishi/src/vscode-swift/.vscode-test/vscode-linux-arm64-1.96.4
✔ Validated version: 1.96.4
✔ Found existing install in /home/rishi/src/vscode-swift/.vscode-test/vscode-linux-arm64-1.96.4
[main 2025-01-20T15:51:04.146Z] update#setState disabled
[main 2025-01-20T15:51:04.149Z] update#ctor - updates are disabled by the environment
Via 'product.json#extensionEnabledApiProposals' extension 'ms-vscode.vscode-selfhost-test-provider' wants API proposal 'attributableCoverage' but that proposal DOES NOT EXIST. Likely, the proposal has been finalized (check 'vscode.d.ts') or was abandoned.
Via 'product.json#extensionEnabledApiProposals' extension 'ms-python.python' wants API proposal 'contribIssueReporter' but that proposal DOES NOT EXIST. Likely, the proposal has been finalized (check 'vscode.d.ts') or was abandoned.
[...]

PR #1287, opened 2024-12-19
Fix debug command test

award999:
* Timing issue waiting for the debug configuration to be created, but we should not have to do this
* Using existing launch config, the  adapter tracker was waiting for "swift-lldb" but we fallback to the "lldb" type for swift versions < 6.0

Issue: #1269

### DIFF
--- a/test/integration-tests/commands/build.test.ts
+++ b/test/integration-tests/commands/build.test.ts
@@ -21,18 +21,14 @@ import { testAssetUri } from "../../fixtures";
 import { FolderContext } from "../../../src/FolderContext";
 import { WorkspaceContext } from "../../../src/WorkspaceContext";
 import { Commands } from "../../../src/commands";
-import { makeDebugConfigurations } from "../../../src/debugger/launch";
 import { Workbench } from "../../../src/utilities/commands";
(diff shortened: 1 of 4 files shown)


test/integration-tests/commands/build.test.ts
```diff
@@ -43,10 +39,11 @@ suite("Build Commands", function () {
 
     activateExtensionForSuite({
         async setup(ctx) {
-            // The description of this package is crashing on Windows with Swift 5.9.x and below,
-            // preventing it from being built. The cleanup in the teardown is failng as well with
-            // an EBUSY error. Skip this test on Windows until the issue is resolved.
-            if (process.platform === "win32") {
+            // The description of this package is crashing on Windows with Swift 5.9.x and below
```
plemarquand, line 42: This comment says 5.9 but code indicates 6.0

plemarquand (approved):

test/integration-tests/commands/build.test.ts
award999: Incoming PR I'll fix that in, but thanks good catch

Issue #1277, opened 2024-12-13
Command plugin whitelist?


MahdiBM:
Running a command plugin asks for permission everytime.
This is specially annoying since the extension can't tell that a plugin requires perms ahead of time, so it can't just switch the panel where the cursor is, to the terminal panel, so i can at least quickly write `yes` and move on.

I think optimally the extension should maintain a whitelist of plugins with the permissions we've allowed them, and if a plugin has been whitelisted, then run it with one of those "allow this perm" SwiftPM plugin flags.

plemarquand:
Fixed by https://github.com/swiftlang/vscode-swift/pull/1297

PR #1297, opened 2025-01-03
Configurable Command Plugin Permissions

plemarquand:
Packages can define their own plugins either directly or through their dependencies. These plugins define commands, and the extension exposes a list of these when you use `> Swift: Run Command Plugin`.

If a command requires special permissions to write to disk or use the network the user is prompted in the integrated terminal to type "yes". This can be bypassed by passing a permission flag to the command such as `--allow-writing-to-package-directory`. The extension does supply permission flags for a small list of well known package plugins, however if the user creates their own or uses one not on this list they must enter "yes" every time they run the command plugin.

This patch introduces a new setting that can be specified globally or on a per workspace folder basis that allows users to configure which permission flags should be used when running the command.

The setting is defined under `swift.pluginPermissions`, and is specified as an object in the following form:

```
{
	"PluginCommandName:intent-name": {
		"allowWritingToPackageDirectory": true,
		"allowWritingToDirectory: "/some/path",
		"allowNetworkConnections: "all",
		"disableSandbox": true,
	}
}
```

- The top level string key is the command id in the form `PluginCommandName:intent-name`. For instance, swift-format's format-source-code command would be specified as `swift-format:format-source-code`
- Each permission in the permissions lookup is optional.
- `allowWritingToDirectory` can also be specified as an array of paths.
- The valid values for `allowNetworkConnections` can be found here: https://github.com/swiftlang/swift-package-manager/blob/0401a2ae55077cfd1f4c0acd43ae0a1a56ab21ef/Sources/Commands/PackageCommands/PluginCommand.swift#L62

Issue: #1277

### DIFF
--- a/package.json
+++ b/package.json
@@ -394,6 +394,12 @@
             "type": "boolean",
             "default": true,
             "markdownDescription": "Controls whether or not the extension will contribute environment variables defined in `Swift: Environment Variables` to the integrated terminal. If this is set to `true` and a custom `Swift: Path` is also set then the swift path is appended to the terminal's `PATH`."
+          },
+          "swift.pluginPermissions": {
+            "type": "object",
+            "default": {},
+            "markdownDescription": "TODO: Document",
+            "scope": "machine-overridable"
           }
         }
       },
--- a/package.json
+++ b/package.json
@@ -398,8 +398,42 @@
           "swift.pluginPermissions": {
             "type": "object",
             "default": {},
-            "markdownDescription": "TODO: Document",
-            "scope": "machine-overridable"
+            "markdownDescription": "Configures a list of permissions to be used when running a command plugins.\n\nPermissions objects are defined in the form:\n\n`{ \"PluginName:command\": { \"allowWritingToPackageDirectory\": true } }`.\n\nA key of `PluginName:command` will set permissions for a specific command. A key of `PluginName` will set permissions for all commands in the plugin.",
+            "scope": "machine-overridable",
+            "patternProperties": {
+              "^([a-zA-Z0-9_-]+(:[a-zA-Z0-9_-]+)?)$": {
+                "type": "object",
+                "properties": {
+                  "disableSandbox": {
+                    "type": "boolean",
+                    "description": "Disable using the sandbox when executing plugins"
+                  },
+                  "allowWritingToPackageDirectory": {
+                    "type": "boolean",
+                    "description": "Allow the plugin to write to the package directory"
+                  },
+                  "allowWritingToDirectory": {
+                    "oneOf": [
+                      {
+                        "type": "string",
+                        "description": "Allow the plugin to write to an additional directory"
+                      },
+                      {
+                        "type": "array",
+                        "items": {
+                          "type": "string"
+                        },
+                        "description": "Allow the plugin to write to additional directories"
+                      }
+                    ]
+                  },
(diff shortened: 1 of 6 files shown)


adam-fowler:
Just to say the way I've got around this so far is click on the cog to create a task in tasks.json and edit the task. I added a field to enable writing to the package directory and a few other options.

This PR is still valid but the solution above should also be advertised as an option as it provides a more fine grained control.

plemarquand:
@adam-fowler thanks for the heads up; I've made sure that this patch still works with the tasks.json permission based setup. It also adds support to pass `--allow-writing-to-directory` and `--allow-network-connections` by setting `allowWritingToDirectory ` and `allowNetworkConnections` on the task respectively.

One of the outstanding things left for this PR is to properly document this behaviour both in the package.json settings markdownDescription and in the additional [settings documentation](https://github.com/swiftlang/vscode-swift/blob/main/docs/settings.md).

MahdiBM:
@plemarquand looks great, thank you.

package.json
```diff
@@ -394,6 +394,46 @@
...
+                      {
+                        "type": "array",
+                        "items": {
+                          "type": "string"
+                        },
+                        "description": "Allow the plugin to write to additional directories"
+                      }
+                    ]
+                  },
+                  "allowNetworkConnections": {
```
MahdiBM, line 430: Can this not be declared as an "enum" with the values SwiftPM accepts?
Even if SwiftPM makes a change, it's trivial for this plugin to update the values and do a release (theoretically at least, not sure what release policies this repo follows).

plemarquand: Almost, but the [`local` and `all` cases](https://github.com/swiftlang/swift-package-manager/blob/0401a2ae55077cfd1f4c0acd43ae0a1a56ab21ef/Sources/Commands/PackageCommands/PluginCommand.swift#L72) have associated values. `local`, `local:1234` and `local:1234,4321` are all valid values, which complicates a straight mapping here in the settings definition.

src/tasks/SwiftPluginTaskProvider.ts
```diff
@@ -196,9 +185,71 @@ export class SwiftPluginTaskProvider implements vscode.TaskProvider {
...
+        const taskDefinitionConfiguration: PluginPermissionConfiguration = {};
+        if (taskDefinition.disableSandbox) {
+            taskDefinitionConfiguration.disableSandbox = true;
+        }
+        if (taskDefinition.allowWritingToPackageDirectory) {
+            taskDefinitionConfiguration.allowWritingToPackageDirectory = true;
+        }
+
+        return this.pluginArguments({
+            ...taskDefinitionConfiguration,
```
award999, line 228: shouldn't task definition take precedence? There are default launch and task settings, but typically what I've seen is the configuration in the .json file takes precedence over settings

award999 (approved):

MahdiBM:
@plemarquand What's the release process in this repo, when can I expect this change to be released? 🙂

plemarquand:
@MahdiBM There is no strict cadence, but it should be available in then next release in the next week or two. If you'd like to try it out beforehand you can follow [these instructions](https://github.com/swiftlang/vscode-swift/blob/main/CONTRIBUTING.md#installing-a-pre-released-version) to build your own `.vsix` package from `main` and install it.

Issue #1304, opened 2025-01-09
Diagnostics are missing on Windows


award999:
`swiftc` parsed diagnostics are missing on windows. This is causing the Windows nightly test run to fail https://github.com/swiftlang/vscode-swift/actions/runs/12677423825/job/35332738251.

It looks like vscode.Uri.fsPath has the tendency to return a lowercase drive letter, ex. "c:\", but what we parse from SwiftPM is uppercase, ex. "C:\". This causes a mismatch where we only get a portion of the diagnostics published to the collection.

rbenegal:
Confirmed that the Diagnostics tests now pass successfully locally and on github actions for Windows versions that use a toolchain greater than 6.0.

<details>
  <summary>
    Output from a local run 
  </summary>

```
swift -v
Swift version 6.2-dev (LLVM 06a77c5cc0ff511, Swift 062061e37a8851b)
Target: x86_64-unknown-windows-msvc
```
on  `vscode 1.96.3` with `Windows 11 24H2`

```
>npm test -- --label integrationTests

> swift-lang@1.11.4 pretest
> npm run compile-tests


> swift-lang@1.11.4 compile-tests
> del-cli ./assets/test/**/.build && npm run compile


> swift-lang@1.11.4 compile
> del-cli ./dist/ && tsc --build


> swift-lang@1.11.4 test
> vscode-test --label integrationTests

✔ Validated version: 1.96.3
✔ Found existing install in C:\Users\rishi\src\vscode-swift\.vscode-test\vscode-win32-arm64-archive-1.96.3
✔ Validated version: 1.96.3
✔ Found existing install in C:\Users\rishi\src\vscode-swift\.vscode-test\vscode-win32-arm64-archive-1.96.3

[6732:0116/115158.254:ERROR:node_bindings.cc(378)] Most NODE_OPTIONs are not supported in packaged apps. See documentation for more details.
[main 2025-01-16T16:51:58.618Z] update#setState disabled
[main 2025-01-16T16:51:58.618Z] update#ctor - updates are disabled by the environment
(node:6732) [DEP0168] DeprecationWarning: Uncaught N-API callback exception detected, please run node with option --force-node-api-uncaught-exceptions-policy=true to handle those exceptions properly.
(Use `Code --trace-deprecation ...` to show where the warning was created)
[...]

PR #1305, opened 2025-01-09
Fix missing windows diagnostics

award999:
Use the Uri's fsPath right away so we don't need to worry about mismatches downstream

Issue: #1304


award999:
@plemarquand past on a nightly I triggered https://github.com/award999/vscode-swift/actions/runs/12692110130

award999:
to clarify....because only was seeing this on main branch toolchain

plemarquand (approved):

Issue #892, opened 2024-06-13
swift-testing: If LSP is disabled, swift-testing tests do not work


adam-fowler:
It appears swift-testing tests are not flagged as swift-testing until they are found via the `workspace/tests` or `textDocument/tests` LSP requests. We need another way to catch this. While disabling SourceKit-LSP is a special situation, it could occur in situations where LSP is particularly slow in large projects.

award999:
Swift-testing is more complex than XCTests so Paul said he would be hesitant to parse this. Some ideas:
1. Update setting description, warning if they disable sourcekit-lsp, they lose this functionality
2. Update warning notification message that shows up to kill the LSP
3. Provide warning in testing view (if we can still show XCTests at same time)

rbenegal:
I can take a look at this. Agree that there should at least be some indication that things won't work as expected if sourcekit-lsp is disabled. Thinking of using a combination of the options mentioned by @award999 above.

For the setting description, perhaps a message like:

```Disabling SourceKit-LSP will turn off features like code completion, error diagnostics and jump-to-definition. Certain features like swift-testing tests may not work correctly.```

And something similar for the message that comes up to turn off the language server.

```You have disabled the Swift language server (SourceKit-LSP), but it is still running. Would you like to stop it now? This will turn off code completion, error diagnostics and jump-to-definition. Certain features like swift-testing tests may not work correctly.```

And a warning in the testing view along the lines of:

```The Swift language server (SourceKit-LSP) is disabled, swift-testing tests may not be detected correctly.```

award999:
Looks like the extra warnings are done, marking as Done

PR #1308, opened 2025-01-13
Improve user feedback if SourceKit-LSP is disabled

rbenegal:
Add some notifications to let users know that several things won't work as expected if SourceKit-LSP is disabled.

Issue: #892

### DIFF
--- a/src/TestExplorer/TestExplorer.ts
+++ b/src/TestExplorer/TestExplorer.ts
@@ -261,6 +261,14 @@ export class TestExplorer {
     async discoverTestsInWorkspaceSPM(token: vscode.CancellationToken) {
         async function runDiscover(explorer: TestExplorer, firstTry: boolean) {
             try {
+                // we depend on sourcekit-lsp to detect swift-testing tests so let the user know
+                // that things won't work properly if sourcekit-lsp has been disabled for some reason
+                if (firstTry === true && configuration.lsp.disable === true) {
+                    vscode.window.showInformationMessage(
+                        `swift-testing tests may not be detected correctly since SourceKit-LSP
+                        has been disabled for this workspace.`
+                    );
+                }
                 const toolchain = explorer.folderContext.workspaceContext.toolchain;
                 // get build options before build is run so we can be sure they aren't changed
                 // mid-build
--- a/src/TestExplorer/TestExplorer.ts
+++ b/src/TestExplorer/TestExplorer.ts
@@ -263,11 +263,31 @@ export class TestExplorer {
             try {
                 // we depend on sourcekit-lsp to detect swift-testing tests so let the user know
                 // that things won't work properly if sourcekit-lsp has been disabled for some reason
-                if (firstTry === true && configuration.lsp.disable === true) {
-                    vscode.window.showInformationMessage(
-                        `swift-testing tests may not be detected correctly since SourceKit-LSP
-                        has been disabled for this workspace.`
-                    );
+                // and provide an option to enable sourcekit-lsp again
+                const ok = "OK";
+                const enable = "Enable SourceKit-LSP";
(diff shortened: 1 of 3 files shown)


src/TestExplorer/TestExplorer.ts
```diff
@@ -261,6 +261,14 @@ export class TestExplorer {
     async discoverTestsInWorkspaceSPM(token: vscode.CancellationToken) {
         async function runDiscover(explorer: TestExplorer, firstTry: boolean) {
             try {
+                // we depend on sourcekit-lsp to detect swift-testing tests so let the user know
+                // that things won't work properly if sourcekit-lsp has been disabled for some reason
+                if (firstTry === true && configuration.lsp.disable === true) {
```
award999, line 266: can just be "firstTry"

src/TestExplorer/TestExplorer.ts
```diff
@@ -261,6 +261,14 @@ export class TestExplorer {
...
+                    vscode.window.showInformationMessage(
+                        `swift-testing tests may not be detected correctly since SourceKit-LSP
```
award999, line 268: @plemarquand is it "may" or is it a guarantee? i.e. "will"

src/TestExplorer/TestExplorer.ts
```diff
@@ -261,6 +261,14 @@ export class TestExplorer {
...
+                        has been disabled for this workspace.`
+                    );
```
award999, line 270: we could also provide a button that re-enables and restarts? Could fewer steps for the user

src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -226,7 +226,9 @@ export class LanguageClientManager implements vscode.Disposable {
                         return;
                     }
                     message =
-                        "You have disabled the Swift language server, but it is still running. Would you like to stop it now?";
+                        `You have disabled the Swift language server, but it is still running. Would you like to stop it now?
+                        This will turn off code completion, error diagnostics and jump-to-definition.
```
award999, line 230: maybe "This will turn off features such as code completion, error diagnostics, jump-to-definition, and test discovery."

award999 (commented):
one question, and left some personal preference suggestions but let me know what you think

award999:
@rbenegal you can run `npm run lint` to see the build failures. You can see these in the editor with the `esbenp.prettier-vscode` and `dbaeumer.vscode-eslint` extensions installed

src/TestExplorer/TestExplorer.ts
rbenegal: Sounds good, will change

src/TestExplorer/TestExplorer.ts
rbenegal: Yes good idea thanks, will try that.

src/TestExplorer/TestExplorer.ts
plemarquand: I'd change this to `will`; XCTest discovery has a fallback in the extension that tries to discover them semantically  if sourcekit-lsp isn't working, but swift-testing test discovery is complicated enough that discovering them requires sourcekit-lsp.

rbenegal: Thank you, will use `will`.

plemarquand:
@swift-ci test

src/TestExplorer/TestExplorer.ts
rbenegal: Added an option to set the sourcekit-lsp setting to enabled. The `LanguageClientManager` should be able to detect the change and prompt the user to restart the language server so didn't want to repeat that logic here.

award999 (approved):
LGTM

Issue #1322, opened 2025-01-21
Build notification is late


MahdiBM:
**Describe the bug**
It appears the notification for building stuff does not appear before the build logs contain "Building for debugging/release..." which is when the `[num/num]` progress indicator starts to show up. That can take a bit to show up so makes the indicator a bit inaccurate.

**To Reproduce**
* Open [vapor/penny-bot](https://github.com/vapor/penny-bot).
* Set `Swift: Show Build Status`  to `notification`.
* Build.

**Expected behavior**
show something like "starting build" instead of not showing the notification at all.

**Environment**
 - OS: [e.g. Ubuntu 20.04] macOS 15.2
 - Swift version (output of `swift --version`): 6.0.3 release
 - Visual Studio Code version: -
 - vscode-swift version: 1.11.4

**Additional context**

MahdiBM:
Reading the extension code, I think it's important to mention what was being logged when the build notification was late:
```bash
Executing task: swift: Building and Running Tests 

> /Applications/Xcode 16.2 beta 3.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swift test --enable-swift-testing --event-stream-version 0 --event-stream-output-path /var/folders/rr/9nl0rwsx2dlcwd49g83wp_jr0000gn/T/vscodemkfifo-1737444536945 --disable-xctest --filter PennyTests\.GHHooksTests\/handlePREvent14\(\)\/ -Xswiftc -diagnostic-style=llvm --force-resolved-versions --disable-xctest

warning: 'swift-evolution-metadata-extractor': 'swift-evolution-metadata-extractor' dependency on 'https://github.com/swiftlang/swift-markdown' conflicts with dependency on 'https://github.com/apple/swift-markdown.git' which has the same identity 'swift-markdown'. this will be escalated to an error in future versions of SwiftPM.
warning: 'leaf-kit': found 1 file(s) which are unhandled; explicitly declare them as resources or exclude from the target
    /Users/mahdibm/Github/penny-bot/.build/checkouts/leaf-kit/Sources/LeafKit/Docs.docc
warning: 'swift-algorithms': found 1 file(s) which are unhandled; explicitly declare them as resources or exclude from the target
    /Users/mahdibm/Github/penny-bot/.build/checkouts/swift-algorithms/Sources/Algorithms/Documentation.docc
Building for debugging...
[68/68] Linking PennyPackageTests
Build complete! (61.36s)
[...]

PR #1323, opened 2025-01-21
Notify build is preparing before progress starts

plemarquand:
The build progress notification would only appear once the progress values were shown in the output, i.e. [12/122]. Until progress values appear, show a "Preparing <task>" notification. This only needs to apply when the `showBuildStatus` value is `notification` or `progress`, since `swiftStatus` already shows a message in the bar immediately.

This patch also only sets the `showBuildStatus` setting on a Task if it is coming from a user defined task. Previously it was being set on synthetic tasks which were being cached, so when the user updated the `showBuildStatus` setting it wouldn't take effect until they restarted the extension.

Issue: #1322

### DIFF
--- a/src/tasks/SwiftTaskProvider.ts
+++ b/src/tasks/SwiftTaskProvider.ts
@@ -155,7 +155,6 @@ export function createBuildAllTask(
                 reveal: getBuildRevealOption(),
             },
             disableTaskQueue: true,
-            showBuildStatus: configuration.showBuildStatus,
         },
         folderContext.workspaceContext.toolchain
     );
@@ -239,7 +238,6 @@ function createBuildTasks(product: Product, folderContext: FolderContext): vscod
                 },
                 disableTaskQueue: true,
                 dontTriggerTestDiscovery: true,
-                showBuildStatus: configuration.showBuildStatus,
             },
             folderContext.workspaceContext.toolchain
         );
@@ -268,7 +266,6 @@ function createBuildTasks(product: Product, folderContext: FolderContext): vscod
                 },
                 disableTaskQueue: true,
                 dontTriggerTestDiscovery: true,
-                showBuildStatus: configuration.showBuildStatus,
             },
             folderContext.workspaceContext.toolchain
         );
@@ -314,9 +311,15 @@ export function createSwiftTask(
             args: args,
             env: env,
             cwd: cwd,
-            showBuildStatus: config.showBuildStatus,
-            disableTaskQueue: config.disableTaskQueue,
-            dontTriggerTestDiscovery: config.dontTriggerTestDiscovery,
+            ...(config.showBuildStatus !== undefined
+                ? { showBuildStatus: config.showBuildStatus }
+                : {}),
+            ...(config.disableTaskQueue !== undefined
+                ? { disableTaskQueue: config.disableTaskQueue }
+                : {}),
+            ...(config.dontTriggerTestDiscovery !== undefined
+                ? { dontTriggerTestDiscovery: config.dontTriggerTestDiscovery }
+                : {}),
         },
         config?.scope ?? vscode.TaskScope.Workspace,
         name,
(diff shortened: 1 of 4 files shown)


plemarquand:
cc @MahdiBM

src/ui/SwiftBuildStatus.ts
```diff
@@ -124,14 +127,27 @@ export class SwiftBuildStatus implements vscode.Disposable {
...
+        // If we've found nothing that matches a known state then put up a temporary
+        // message that we're preparing the build, as there is sometimes a delay before
+        // building starts while the build system is preparing, especially in large projects.
+        // The status bar has a message immediately, so only show this when using a
+        // notification to show progress.
+        if (
+            !state.started &&
+            (showBuildStatus === "notification" || showBuildStatus === "progress")
+        ) {
+            update(`Preparing ${name}`);
```
award999, line 149: to keep your format above, maybe `${name}: Preparing`

src/tasks/SwiftTaskProvider.ts
```diff
@@ -314,9 +311,15 @@ export function createSwiftTask(
             args: args,
             env: env,
             cwd: cwd,
-            showBuildStatus: config.showBuildStatus,
-            disableTaskQueue: config.disableTaskQueue,
-            dontTriggerTestDiscovery: config.dontTriggerTestDiscovery,
+            ...(config.showBuildStatus !== undefined
+                ? { showBuildStatus: config.showBuildStatus }
+                : {}),
```
award999, line 316: wouldn't default be to fallback to `configuration.showBuildStatus`?

MahdiBM (approved):
Thank you. A minor annoyance that will no longer exist!

src/ui/SwiftBuildStatus.ts
award999: same as line 129

src/ui/SwiftBuildStatus.ts
```diff
@@ -71,9 +71,10 @@ export class SwiftBuildStatus implements vscode.Disposable {
                     disposables.forEach(d => d.dispose());
                     res();
                 };
+                const state = { started: false };
                 disposables.push(
                     execution.onDidWrite(data => {
-                        if (this.parseEvents(task, data, update)) {
+                        if (this.parseEvents(task, data, showBuildStatus, update, state)) {
```
award999, line 77: instead of passing down parameters, can't we set that first time this onDidWrite event is triggered, before parsing any events?

award999 (changes requested):
Some minor things

src/tasks/SwiftTaskProvider.ts
plemarquand: No, we don't want to resolve the setting too early because tasks are cached. This means if the user updates their setting, the old value is still cached on the task. Ideally we resolve all settings right before use to avoid this kind of bug. The only case where the setting is attached to the task is when it comes from a user defined task, which should take precedence over the setting on `configuration`.

src/ui/SwiftBuildStatus.ts
plemarquand: I found its possible to get some lines of build output that don't parsed as progress once the build starts, so it would flip back to the "Preparing" notification for a moment mid build. Once we see the first line that parses as progress we need to keep some state so we don't flip back in to Preparing if we encounter a line like this.

src/ui/SwiftBuildStatus.ts
plemarquand: I've made these messages consistent

src/tasks/SwiftTaskProvider.ts
award999: ah see where it is resolved in `handleTaskStatus`

src/ui/SwiftBuildStatus.ts
award999: but couldn't we do something like?
```
executions.onDidWrite(data => {
    if (!state.started) {
        if (showBuildStatus === "notification" || showBuildStatus === "progress") {
           update(`Preparing ${name}`);
        }
        state.started = true;
    }
    if (this.parseEvents(task, data, update)) {
...
}
```

award999: I just meant instead of needing to do this in the function so we don't need 5 parameters

plemarquand: Oh, yes, sorry I misunderstood, this is better

award999 (approved):

award999:
Verified with fa4d12c

Issue #1315, opened 2025-01-15
Nightly Test: Language Client Integration Suite - Expand Macro test failed on Linux (nightly-main - jammy)


rbenegal:
The Expand Macro test failed in the nightly run with:

```
  Language Client Integration Suite @slow
Language client is not ready yet. Retrying in 100 ms...
Language client is not ready yet. Retrying in 100 ms...
    1) Expand Macro
Captured logs during Language Client Integration Suite @slow → Expand Macro:
00:34:27: Activating Swift for Visual Studio Code...
00:34:27: Swift version 6.2-dev (LLVM 81859ac55f8d09a, Swift 8ec8a1229a7ea14)
Platform: linux
Swift Path: /usr/bin
Toolchain Path: /usr
Runtime Library Path: /usr/lib/swift/linux
Default Target: x86_64-unknown-linux-gnu
00:34:27: SourceKit-LSP setup
00:34:27: focus: undefined
00:34:27: test/defaultPackage: add: /tmp/code/assets/test/defaultPackage

...

rejected promise not handled within 1 second: Error: Timed out waiting for sourcekit-lsp document to be opened.
stack trace: Error: Timed out waiting for sourcekit-lsp document to be opened.
    at Timeout._onTimeout (/tmp/code/test/integration-tests/language/LanguageClientIntegration.test.ts:110:24)
    at listOnTimeout (node:internal/timers:581:17)
    at processTimers (node:internal/timers:519:7)

```

Link to failed run: https://github.com/swiftlang/vscode-swift/actions/runs/12778956212/job/35622745761

Seems like it was passing fine previously, so possibly just flaky.

rbenegal:
Failed again in https://github.com/swiftlang/vscode-swift/actions/runs/12848727861/job/35826444826:

```
 1) Language Client Integration Suite @slow
       Expand Macro:
     Error: Unknown error: no edits reported for semantic refactoring action for url file:///tmp/code/assets/test/swift-macro/Sources/swift-macroClient/main.swift
      at handleResponse (node_modules/vscode-jsonrpc/lib/common/connection.js:565:48)
      at handleMessage (node_modules/vscode-jsonrpc/lib/common/connection.js:345:13)
      at processMessageQueue (node_modules/vscode-jsonrpc/lib/common/connection.js:362:17)
      at Immediate.<anonymous> (node_modules/vscode-jsonrpc/lib/common/connection.js:334:13)
      at processImmediate (node:internal/timers:483:21)
[...]

PR #1324, opened 2025-01-21
Remove Expand Macro Test

rbenegal:
Currently, it seems like the Expand Macros test has some issues that makes it fail intermittently when run on GitHub actions. These issues seem to be hard to reproduce and likely require simplifying the dependencies the test depends on. Remove the test until there is a better way to test this to avoid noise.

Issue: #1315


award999 (approved):
Looks good

PR #1339, opened 2025-01-27
Show project dependencies hierarchically

plemarquand:
Currently the Package Dependencies view lists all dependencies in a flat list. This list not only includes those explicitly defined in the project's Package.swift, but also all the dependencies of dependencies.

As such its difficult to tell at a glance how a dependency is included. It can also be confusing to see dependencies in the list that you did not explicitly add.

Instead, show a top level list that is only the dependencies explicitly defined in Package.swift. Expanding one shows any child dependencies of the dependency. The files in the dependency's folder are still shown as well, just after any child deps.

### DIFF
--- a/src/SwiftPackage.ts
+++ b/src/SwiftPackage.ts
@@ -47,10 +47,18 @@ export interface Target {
 /** Swift Package Manager dependency */
 export interface Dependency {
     identity: string;
-    type?: string; // fileSystem, sourceControl or registry
+    type?: string;
     requirement?: object;
     url?: string;
     path?: string;
+    dependencies: Dependency[];
+}
+
+export interface ResolvedDependency extends Omit<Omit<Dependency, "type">, "path"> {
+    version: string;
+    type: string;
+    path: string;
+    location: string;
 }
 
 /** Swift Package.resolved file */
@@ -187,7 +195,8 @@ export class SwiftPackage implements PackageContents {
     private constructor(
         readonly folder: vscode.Uri,
         private contents: SwiftPackageState,
-        public resolved: PackageResolved | undefined
+        public resolved: PackageResolved | undefined,
+        private workspaceState: WorkspaceState | undefined
     ) {}
 
     /**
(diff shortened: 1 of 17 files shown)


plemarquand:
Screenshot 2025-01-27 at 3 40 23 PM

src/SwiftPackage.ts
```diff
@@ -375,6 +493,15 @@ export class SwiftPackage implements PackageContents {
         const filePath = path.relative(this.folder.fsPath, file);
         return this.targets.find(target => isPathInsidePath(filePath, target.path));
     }
+
+    private static trimStdout(stdout: string): string {
```
award999, line 497: we could look at the `jsonc-parser` npm dependency which was developed for vscode, and a lot more error resilient for stuff like this

award999 (commented):
looks good, some suggestions

src/SwiftPackage.ts
plemarquand: May be kind of heavyweight for this? All this does is trim any warnings/preamble that the swift tool may or may not emit before getting to the JSON.

test/unit-tests/ui/PackageDependencyProvider.test.ts
```diff
@@ -46,17 +48,70 @@ suite("PackageDependencyProvider Unit Test Suite", function () {
...
             const item = node.toTreeItem();
 
             expect(item.label).to.equal("SwiftMarkdown");
             expect(item.description).to.deep.equal("1.2.3");
             expect(item.command).to.be.undefined;
         });
+
+        const fsMock = mockGlobalModule(fs);
+
+        test("enumeates child dependencies and files", async () => {
```
rbenegal, line 70: should be `enumerates` I think

adam-fowler:
Is it worthwhile having a toggle to switch between displaying a flat or hierarchical view?

plemarquand:
@adam-fowler I've added a toggle button to switch between flat and nested to the panel (defaults to nested)

award999 (approved):

Issue #1340, opened 2025-01-28
Include build parameters in calls to swift package resolve and update


adam-fowler:
**Is your feature request related to a problem? Please describe.**
Currently there is no way to add command line parameters to either the resolve or update tasks

**Describe the solution you'd like**
Either 
- add an additional setting for resolve/update command line parameters (which would also have to be included in build/test) 
- Or just pass the build command line parameters to resolve and update. The help for `Build arguments` actually suggests this already happens.

**Additional context**
This is so I can pass the command line parameter `--replace-scm-with-registry` to a resolve.

plemarquand:
Thanks for this, makes sense.

I think this needs to be a separate setting from the build parameters since package resolve will often claim 'unknown argument' when passed flags from build. For instance `swift build --target MyTarget` may work while `swift package resolve --target MyTarget` fails with `Unknown option --target`.

I'll be sure to update the docs for the `swift.buildArguments` setting as well as it does imply those args will be forwarded to `swift package ...` commands.

adam-fowler:
You'll need to include the resolve command line parameters in the build parameters though (as a build can run a resolve)

award999:
Verified 77f8cf5

PR #1342, opened 2025-01-28
Add `swift.packageArguments` setting

plemarquand:
Adds a new setting to provide arguments to swift commands that can resolve packages. This applies to `swift package resolve`, `swift package update`, `swift build` and `swift test`.

This lets users set package resolution specific flags that should be used whenever resolution is performed, such as
`--replace-scm-with-registry`.

Issue: #1340


package.json
```diff
@@ -271,7 +271,15 @@
...
+            "markdownDescription": "Additional arguments to pass to `swift build` and `swift test`. Keys and values should be provided as individual entries in the list. If you have created a copy of the build task in `tasks.json` then these build arguments will not be propagated to that task."
+          },
+          "swift.packageArguments": {
+            "type": "array",
+            "default": [],
+            "items": {
+              "type": "string"
+            },
+            "markdownDescription": "Additional arguments to pass to swift commands that do package resolution, such as `swift package resolve`, `swift package update`, `swift build` and `swift test`. Keys and values should be provided as individual entries in the list. If you have created a copy of the build task in `tasks.json` then these build arguments will not be propagated to that task."
```
award999, line 282: probably can omit the task part?

src/commands/dependencies/resolve.ts
```diff
@@ -39,7 +40,7 @@ export async function resolveFolderDependencies(
     checkAlreadyRunning?: boolean
 ) {
     const task = createSwiftTask(
-        ["package", "resolve"],
+        ["package", "resolve", ...configuration.packageArguments],
```
award999, line 43: can we do something similar to ` private withSwiftSDKFlags(args: string[]): string[] {` so we don't have to remember to add this to every task creation. https://github.com/swiftlang/vscode-swift/pull/1272/files#diff-5b686aa5c29d31071cb99f6f8876d6f0e965af838aac442d5a479d6c90b7ef0fR289 was the direction I was going to try to go

plemarquand: Done

Issue #1334, opened 2025-01-24
Test Failure: Test Explorer Suite - XCTests (Run Tests (Parallel)) failed on Test Insiders / Linux (6.0 - jammy)


rbenegal:
**Failure logs**

```
1) Test Explorer Suite
       Standard
         Run Tests (Parallel)
           XCTests (Run Tests (Parallel))
             Runs failing test:

      AssertionError [ERR_ASSERTION]: Expected values to be loosely deep-equal:

{
  errored: [],
  failed: [
    {
      issues: [],
      test: 'PackageTests.FailingXCTestSuite'
    }
  ],
  passed: [
    'PackageTests.FailingXCTestSuite/testFailing'
  ],
  skipped: [],
  unknown: 0
}

should loosely deep-equal

{
  errored: [],
  failed: [
    {
      issues: [
        'failed - oh no'
      ],
      test: 'PackageTests.FailingXCTestSuite/testFailing'
    },
    {
      issues: [],
      test: 'PackageTests.FailingXCTestSuite'
    }
  ],
  passed: [],
  skipped: [],
  unknown: 0
}
      + expected - actual

       {
         "errored": []
         "failed": [
           {
      +      "issues": [
      +        "failed - oh no"
      +      ]
      +      "test": "PackageTests.FailingXCTestSuite/testFailing"
      +    }
      +    {
             "issues": []
             "test": "PackageTests.FailingXCTestSuite"
           }
         ]
      -  "passed": [
      -    "PackageTests.FailingXCTestSuite/testFailing"
      -  ]
      +  "passed": []
         "skipped": []
         "unknown": 0
       }
      
      at assertTestResults (test/integration-tests/testexplorer/utilities.ts:137:12)
      at Context.<anonymous> (test/integration-tests/testexplorer/TestExplorerIntegration.test.ts:746:42)

1 test failed.
```

Link to failed run: https://github.com/swiftlang/vscode-swift/actions/runs/12940729478/job/36095522953

matthewbastien:
Haven't seen this fail in the nightly for a while.

PR #1343, opened 2025-01-28
Fix flakey XCTest parallel test failure reporting

plemarquand:
The `ParallelXCTestRunStateProxy` in `XCTestOutputParser.ts` wraps the test run state and only allows issues to be recorded using the terminal output of the test run. The remaining test run state is filled in with the xml xUnit output. This is because SwiftPM only dumps test output to the terminal if there is a failure. Using these two sources we can (almost) fully reconstruct a regular test run.

When parsing XCTest output from the terminal failure messages might be broken up over multiple lines. In order to capture all these lines and group them in to a single failure message even if the message is broken up across terminal buffer chunks we maintain an `excess` variable on the TestRunState, which we check at the beginning of chunk parsing to know if we need to continue an error message.

However, the `ParallelXCTestRunStateProxy` that wraps the test run state was not forwarding along the setting of `excess`. Instead `excess` was set on the proxy, which is recreated for every chunk of output parsed.

This manifested as XCTests _sometimes_ not correctly reporting their error message when run in the Parallel test profile, instead reporting the message "Failed".

Issue: #1334

### DIFF
--- a/src/TestExplorer/TestParsers/XCTestOutputParser.ts
+++ b/src/TestExplorer/TestParsers/XCTestOutputParser.ts
@@ -92,8 +92,9 @@ export class ParallelXCTestOutputParser implements IXCTestOutputParser {
     public parseResult(output: string, runState: ITestRunState) {
         // From 5.7 to 5.10 running with the --parallel option dumps the test results out
         // to the console with no newlines, so it isn't possible to distinguish where errors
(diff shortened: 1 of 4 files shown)


award999 (approved):

plemarquand:
Converting those to draft so it doesn’t get merged yet, as I’m still using it to track down the source of a test failure

matthewbastien (approved):

PR #1353, opened 2025-01-31
Rename extension ID from `swiftlang.vscode-swift` to `swiftlang.swift`

plemarquand:
Because there is already another extension with `vscode-swift` as its name, and the VS Code Marketplace _name_ must be unique (not just the full ID), switch to the extension ID `swiftlang.swift`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## 2.0.0 - 2025-01-28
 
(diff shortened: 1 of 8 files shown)


0xTim:
When will this go live? Currently the README link is broken and points to a non-existent extension

award999 (approved):

PR #1362, opened 2025-02-06
Use one launch configuration type for all debug adapters

matthewbastien:
This PR changes the behaviour of the generated launch configurations to always use a type of `"swift"` instead of switching between `"swift-lldb"` and `"lldb"` based on user settings. This new launch configuration type will then delegate to the appropriate debug adapter depending on settings. That way the extension no longer has to check for debug configuration updates all the time.

I've deprecated the `swift.debugger.useDebuggerFromToolchain` setting in favour of using a new `swift.debugger.debugAdapter` setting that is an enumeration of `"auto"`, `"CodeLLDB"`, and `"lldb-dap"`. Auto mode will continue to use CodeLLDB by default until we can properly move to the LLDB DAP extension provided by LLVM. Setting this to `"lldb-dap"` will continue to use the extension provided debug adapter executable factory for the same reason.

### DIFF
--- a/package.json
+++ b/package.json
@@ -638,9 +638,25 @@
       {
         "title": "Debugger",
         "properties": {
+          "swift.debugger.debugAdapter": {
+            "type": "string",
+            "default": "auto",
+            "enum": [
+              "auto",
+              "lldb-dap",
+              "CodeLLDB"
+            ],
+            "enumDescriptions": [
+              "Automatically select which debug adapter to use based on your Swift toolchain version.",
+              "Use the LLDB debug adapter extension. Requires Swift 6 or later.",
+              "Use the CodeLLDB extension's debug adapter."
+            ],
+            "order": 1
+          },
           "swift.debugger.useDebugAdapterFromToolchain": {
             "type": "boolean",
             "default": false,
+            "markdownDeprecationMessage": "**Deprecated**: Use the `swift.debugger.debugAdapter` setting instead. This will be removed in future versions of the Swift extension.",
             "markdownDescription": "Use the LLDB debug adapter packaged with the Swift toolchain as your debug adapter. Note: this is only available starting with Swift 6. The CodeLLDB extension will be used if your Swift toolchain does not contain lldb-dap.",
             "order": 1
           },
@@ -1158,8 +1174,8 @@
     ],
     "debuggers": [
       {
-        "type": "swift-lldb",
-        "label": "Swift LLDB Debugger",
+        "type": "swift",
+        "label": "Swift Debugger",
         "configurationAttributes": {
           "launch": {
             "required": [
--- a/package.json
+++ b/package.json
@@ -648,7 +648,7 @@
             ],
             "enumDescriptions": [
               "Automatically select which debug adapter to use based on your Swift toolchain version.",
-              "Use the LLDB debug adapter extension. Requires Swift 6 or later.",
+              "Use the `lldb-dap` executable from the toolchain. Requires Swift 6 or later.",
               "Use the CodeLLDB extension's debug adapter."
             ],
             "order": 1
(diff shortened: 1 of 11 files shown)


package-lock.json
```diff
@@ -1,13 +1,13 @@
 {
-  "version": "1.11.4",
   "name": "swift-vscode",
+  "version": "2.0.0",
```
award999, line 3: rebase needed?

package.json
```diff
@@ -638,9 +638,25 @@
...
+            "type": "string",
+            "default": "auto",
+            "enum": [
+              "auto",
+              "lldb-dap",
+              "CodeLLDB"
+            ],
+            "enumDescriptions": [
+              "Automatically select which debug adapter to use based on your Swift toolchain version.",
+              "Use the LLDB debug adapter extension. Requires Swift 6 or later.",
```
award999, line 651: can we use official name in ``

src/debugger/debugAdapterFactory.ts
```diff
@@ -157,20 +151,20 @@ export class LLDBDebugConfigurationProvider implements vscode.DebugConfiguration
             launchConfig.program += ".exe";
         }
 
-        // Delegate to CodeLLDB if that's the debug adapter we have selected
-        if (DebugAdapter.getLaunchConfigType(this.swiftVersion) === LaunchConfigType.CODE_LLDB) {
-            launchConfig.type = LaunchConfigType.CODE_LLDB;
+        // Delegate to the appropriate debug adapter extension
+        launchConfig.type = DebugAdapter.getLaunchConfigType(this.toolchain.swiftVersion);
+        if (launchConfig.type === LaunchConfigType.CODE_LLDB) {
             launchConfig.sourceLanguages = ["swift"];
```
award999, line 157: codelldb also has an env object https://github.com/vadimcn/codelldb/blob/master/package.json#L792

src/debugger/debugAdapterFactory.ts
```diff
@@ -157,20 +151,20 @@ export class LLDBDebugConfigurationProvider implements vscode.DebugConfiguration
...
     }
 
-    convertEnvironmentVariables(
-        map: { [key: string]: string } | undefined
-    ): { [key: string]: string } | string[] | undefined {
-        if (map === undefined) {
-            return undefined;
-        }
+    private convertEnvironmentVariables(map: { [key: string]: string }): string[] {
         return Object.entries(map).map(([key, value]) => `${key}=${value}`);
```
award999, line 168: We were going to update LLDB DAP extension to take an object, not strings

award999 (changes requested):

src/debugger/debugAdapterFactory.ts
matthewbastien: The `env` property only needs to be translated for `lldb-dap` since older versions do not support the property being an object. For CodeLLDB we just pass it through and it works fine.

https://github.com/llvm/llvm-project/blob/main/lldb/tools/lldb-dap/package.json#L172

src/debugger/debugAdapterFactory.ts
matthewbastien: Also, this PR doesn't switch us to using the LLDB DAP extension. It only changes the launch config type.

package.json
matthewbastien: Actually, this is incorrect. This PR doesn't switch us to use the LLDB DAP extension like the description says it does. Let me fix the wording. That's going to come later.

src/debugger/debugAdapterFactory.ts
award999: lets revert this out then until DAP extension is ready, instead of reverting back later

package.json
matthewbastien: done

package-lock.json
matthewbastien: rebased

award999 (approved):

PR #1376, opened 2025-02-11
Make CommandLineTools toolchain selectable if installed

plemarquand:
If the CLT are installed add it to the list of available toolchains.

Also, if there are no Xcodes installed, omit the top level item from the list of quick pick options since it would render as an empty row.

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -363,8 +363,8 @@ export class SwiftToolchain {
             return [];
         }
 
-        const coolchainSwiftPath = path.join(commandLineToolsPath, "usr", "bin", "swift");
-        if (!(await pathExists(coolchainSwiftPath))) {
+        const toolchainSwiftPath = path.join(commandLineToolsPath, "usr", "bin", "swift");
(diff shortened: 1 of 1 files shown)


src/toolchain/toolchain.ts
```diff
@@ -349,6 +354,22 @@ export class SwiftToolchain {
...
+    /**
+     * Returns the path to the CommandLineTools toolchain if its installed.
+     */
+    public static async findCommandLineTools(): Promise<string[]> {
+        const commandLineToolsPath = "/Library/Developer/CommandLineTools";
+        if (!(await pathExists(commandLineToolsPath))) {
+            return [];
+        }
+
+        const coolchainSwiftPath = path.join(commandLineToolsPath, "usr", "bin", "swift");
```
award999, line 366:
```suggestion
        const toolchainSwiftPath = path.join(commandLineToolsPath, "usr", "bin", "swift");
```

matthewbastien (approved):

src/toolchain/toolchain.ts
plemarquand: But but but.... fine....

Issue #1364, opened 2025-02-07
Unable to test `.xctest` binaries on Windows with CMake-based build (`.xctest.exe` not found)


z2oh:
When trying to debug an XCTest executable on Windows, the extension will append `.exe` to the path, which now points to a file that doesn't exist. We have a pretty unique CMake based build for our Swift project on Windows, but I couldn't find a way to fix this issue in configuration.

Our launch configuration is as follows:
```
        {
            "type": "swift-lldb",
            "request": "launch",
            "name": "Debug Test",
            "program": "${cmake.testProgram}",
            "args": [ "${cmake.testArgs}" ],
            "cwd": "${cmake.testWorkingDirectory}",
            "preLaunchTask": "Build All",
            "internalConsoleOptions": "openOnSessionStart"
        }
```
and when running, I get a popup which says `<packagename>.xctest.exe` does not exist.

It is this code which is adding the `.exe` extension to the path: https://github.com/swiftlang/vscode-swift/blob/5cd74b4d75fc2a15fd62e509d3f9f907f056decf/src/debugger/debugAdapterFactory.ts#L151-L158

When I've debugged the extension, I've seen that `launchConfig.testType` is indeed undefined, although I suspect this is _supposed_ to be set to `TestLibrary.xctest`, probably here: https://github.com/swiftlang/vscode-swift/blob/5cd74b4d75fc2a15fd62e509d3f9f907f056decf/src/TestExplorer/TestRunner.ts#L899

I assume this pathway is hit when using the Swift extension's test explorer, but we are using the [CMake Tools extension](https://github.com/microsoft/vscode-cmake-tools)'s test explorer which is easier to hook up to our CMake-based build.

Changing the predicate to:
```
if (
    this.platform === "win32" &&
    launchConfig.testType === undefined &&
    path.extname(launchConfig.program) !== ".exe" &&
    path.extname(launchConfig.program) !== ".xctest"
)
```
is enough to fix this problem and I suspect should be the desired behavior in all cases. I'm happy to make a PR for this, but figured I should file an issue first in case there's a better way to resolve this, or if maybe I missed some other way to set `launchConfig.testType` via a configuration key or similar.

z2oh:
> As a temporary workaround does adding `"testType": "XCTest"` to your launch config fix the issue?

Oh nice, this does work!

[...]

PR #1377, opened 2025-02-11
Don't append .exe to binaries ending in .xctest

plemarquand:
Certain non standard workflows (CMake) could produce binaries that end in .xctest. Only append .exe if it doesn't already have a valid extension.

Also declare `testType` in the debug configuration schema allowing users to declare what type of test framework they're using in the launch config if they're using a non standard flow.

Issue: #1364

### DIFF
--- a/src/debugger/debugAdapterFactory.ts
+++ b/src/debugger/debugAdapterFactory.ts
@@ -152,7 +152,8 @@ export class LLDBDebugConfigurationProvider implements vscode.DebugConfiguration
         if (
             this.platform === "win32" &&
             launchConfig.testType === undefined &&
-            path.extname(launchConfig.program) !== ".exe"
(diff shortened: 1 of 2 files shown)


src/debugger/debugAdapterFactory.ts
```diff
@@ -152,7 +152,8 @@ export class LLDBDebugConfigurationProvider implements vscode.DebugConfiguration
         if (
             this.platform === "win32" &&
             launchConfig.testType === undefined &&
-            path.extname(launchConfig.program) !== ".exe"
```
z2oh, line 155: Is there not still value in this check? Or does something else prevent us from getting `.exe.exe`?

award999 (approved):

matthewbastien (approved):

Issue #1320, opened 2025-01-17
Background Compilation: configurable build task


MahdiBM:
I don't always want to build to whole thing when background compilation is on.
We should be able to configure what build task to use when a background compilation is to be done.
Somewhat similar to how you'd choose a specific scheme in Xcode, when working on different things.

Per usual, you can use [vapor/penny-bot](https://github.com/vapor/penny-bot) as an example. There are 9 different executables there, and usually i'd work on the 2 bigger executables, `GHHooksLambda` _or_ `Penny`, and sometimes building tests is not needed either.
One can always manually trigger a more complete build if needed when changing the configured background-compilation build task is not desired.

award999:
`swift.backgroundCompilation` will now accept and object notation to give more control:
```
"swift.backgroundCompilation": {
    "enabled": boolean,
    "useDefaultTask": boolean,
    "release": boolean
}
```


Setting `"swift.backgroundCompilation": true` functions with the equivalent of these defaults:
```
"swift.backgroundCompilation": {
    "enabled": true,
    "useDefaultTask": true,
    "release": false
}
```

when `"useDefaultTask"` is `true`, then whichever task the user has configured as the default build group task will be used for background compilation.

`"release"` and `"useDefaultTask"` are mutually exclusive. The `"useDefaultTask"` property takes precedence. When `"release": true` is set then the release variant of the Build All task is used

matthewbastien:
Verified with 6c9733e1

PR #1378, opened 2025-02-11
add editor run actions for Swift executables

matthewbastien:
This PR adds two new actions in the editor for running and debugging the active Swift file. These actions use existing commands and are only shown when the active Swift file belongs to an executable target.

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,13 +136,15 @@
       },
       {
         "command": "swift.run",
-        "title": "Run Build",
-        "category": "Swift"
+        "title": "Run active Swift file",
+        "category": "Swift",
+        "icon": "$(play)"
       },
       {
         "command": "swift.debug",
-        "title": "Debug Build",
-        "category": "Swift"
+        "title": "Debug active Swift file",
+        "category": "Swift",
+        "icon": "$(debug)"
       },
       {
         "command": "swift.resetPackage",
@@ -860,6 +862,18 @@
           "command": "swift.previewDocumentation",
           "when": "swift.supportsDocumentationLivePreview && (editorLangId == markdown || editorLangId == tutorial || editorLangId == swift)",
           "group": "navigation"
+        }
+      ],
+      "editor/title/run": [
+        {
+          "command": "swift.run",
+          "group": "navigation@0",
+          "when": "resourceLangId == swift && !isInDiffEditor && !virtualWorkspace && swift.currentTargetType == 'executable'"
+        },
+        {
+          "command": "swift.debug",
+          "group": "navigation@0",
+          "when": "resourceLangId == swift && !isInDiffEditor && !virtualWorkspace && swift.currentTargetType == 'executable'"
         }
       ],
       "swift.editor": [
--- a/package.json
+++ b/package.json
@@ -136,13 +136,13 @@
       },
       {
         "command": "swift.run",
-        "title": "Run active Swift file",
+        "title": "Run Swift executable",
         "category": "Swift",
         "icon": "$(play)"
       },
       {
         "command": "swift.debug",
-        "title": "Debug active Swift file",
+        "title": "Debug Swift executable",
         "category": "Swift",
         "icon": "$(debug)"
       },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -136,13 +136,15 @@
       },
       {
         "command": "swift.run",
-        "title": "Run Build",
-        "category": "Swift"
+        "title": "Run active Swift file",
```
MahdiBM, line 139: I can see the wording being confusing. It won't actually run that specific single file, will it?

Maybe "Run recent Swift executable"? So also if you switch to another file for a sec it won't just go away since it says "recent"?

I could also get behind just having it for the "current Swift executable" since I'd guess it will have a more predictable algorithms and users won't need to remember what executable they were working on last.
But I'd still slightly prefer "recent" since I feel like that's what will be more useful to me based on what I remember from working on stuff.

Also would be nice to have build actions for current active target as well, perhaps in another PR. Related to https://github.com/swiftlang/vscode-swift/issues/1320

matthewbastien: Yeah, that's a good point. The command looks at the most recently opened Swift file and runs the executable target that builds it (if any). Though wording it as `Run recent Swift executable` would look weird in the editor action.

I think what I'd rather do is rename the two commands `Run Swift executable`/`Debug Swift executable` and hide them from the command palette. As far as I'm aware these commands are only used by SourceKit-LSP to launch targets via CodeLens. In this case (and the case of clicking the run button in the editor actions) then there is definitely an active file that can be run/debug in a Swift executable. I wouldn't expect a user to run this command themselves from the command palette.

matthewbastien: Actually, taking a closer look at this we already limit the run and debug commands to show up in the command palette only when a Swift file built by an executable target is active. I've just renamed the commands because of this. Let me know what you think.

MahdiBM: hmm I never noticed these commands but I can see them now. This is actually pretty cool, I would've used these commands if I knew they existed. I probably wished for something like this a few times. Though most of the times I probably want just a "Build" not a "Run".

[vapor/penny-bot](https://github.com/vapor/penny-bot) has IIIRC 8 executables, so it would be nice to have these commands so I don't even need to pay attention to "what lambda executable am I in?" then do CMD+B and choose that specific executable to build. Swift's build times are relatively slow so I usually would like to only build what I'm actually focusing on.

So can we have these same commands but only for _building_ the active target? Should I create a new issue?

Also for visibility, should we not conditionally hide these commands and maybe just error out with "current file is not executable" or just run the most recent executable? Not 100% sure about this recommendation but sounds like an improvement to me right now. WDYT?

matthewbastien: > [vapor/penny-bot](https://github.com/vapor/penny-bot) has IIIRC 8 executables, so it would be nice to have these commands so I don't even need to pay attention to "what lambda executable am I in?" then do CMD+B and choose that specific executable to build. Swift's build times are relatively slow so I usually would like to only build what I'm actually focusing on.

You should be able to build a specific target by running the appropriate build task. If you use the VS Code shortcut `CMD + Shift + B` it'll pop up build tasks for all of your targets. Is that what you're looking for?

> Also for visibility, should we not conditionally hide these commands and maybe just error out with "current file is not executable" or just run the most recent executable? Not 100% sure about this recommendation but sounds like an improvement to me right now. WDYT?

I'll have to think about that one more. Perhaps it makes more sense to add a new command that gives you a dropdown of all available targets that can be debugged? Could be another entry point into debugging.

award999: Maybe a palette command for letting the user select a target, but I wouldn't think we'd add it as an editor action which are intended to be provided for the context of the current editor

award999 (approved):

package.json
MahdiBM: @matthewbastien 

> If you use the VS Code shortcut CMD + Shift + B

That's actually what I meant by "CMD+B" in my comment. That's fine and it's doing what it's supposed to, but no that's not what I want considering that's what I already have/use.

I prefer an easier way to trigger the right build. I guess It's more complicated than what I thought considering users might want to have custom build settings and a simple command pallete command is probably not the best place for that 🤔.

In the ideal world the extension would just know what build task I want to run (first build task that is related to the current most recent file perhaps?) and I can just use that. But maybe I'm getting a bit impractical. Maybe "Xcode schemes" is the best thing we could have without getting too messy, considering other stuff like the "background compilation" feature can also use that to not build the whole app when a specific "scheme" is selected, like mentioned in #1320 🤔

matthewbastien: Ah, I think I see what you mean now. Yeah that's something that needs some design work for sure to make it practical. Thanks for raising that issue!

plemarquand:

PR #1384, opened 2025-02-12
Use LLDB DAP extension to launch debug sessions

matthewbastien:
Add an extension dependency on the LLDB DAP extension by the LLVM project. I've also removed the debug adapter executable factory and instead launch debug sessions for lldb-dap via the LLDB DAP extension.

A PR will come later for removing the dependency on CodeLLDB.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1459,7 +1459,8 @@
     ]
   },
   "extensionDependencies": [
-    "vadimcn.vscode-lldb"
+    "vadimcn.vscode-lldb",
+    "llvm-vs-code-extensions.lldb-dap"
   ],
   "scripts": {
     "vscode:prepublish": "npm run bundle",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1417,7 +1417,8 @@
     ]
   },
   "extensionDependencies": [
-    "vadimcn.vscode-lldb"
+    "vadimcn.vscode-lldb",
```
award999, line 1420: wouldn't the user manually install CodeLLDB? Otherwise still would get the warning notification for unverified publisher

matthewbastien: I'm going to put up a separate PR for removing our dependency on CodeLLDB since it also involves adding error messages.

matthewbastien: Remove dependency on CodeLLDB: #1385

plemarquand (approved):

plemarquand:

PR #1386, opened 2025-02-13
Allow disabling SwiftPM sandboxing

award999:
To get our tests to run in a sandbox, we need to disable sandboxing because you cannot create a new sandbox when you're already running under a sandbox

- Add new `swift.disableSandbox` setting
- Disable sandboxing for tasks and commands run by the extension
- Disable hardware acceleration since ci.swift.org will run on x64
- Fix failing xcode watcher unit test
- Increase some timeouts as build times seem slower on these nodes
- Skip any LSP dependent tests for 6.0 or earlier. The LSP will only allow disabling sandboxing in 6.1+
- Disable debugging tests since need shareport permission

### DIFF
--- a/src/toolchain/BuildFlags.ts
+++ b/src/toolchain/BuildFlags.ts
@@ -202,14 +202,15 @@ export class BuildFlags {
         if (!configuration.disableSandbox) {
             return args;
         }
+        const disableSandboxFlags = ["--disable-sandbox", "-Xswiftc", "-disable-sandbox"];
         switch (args[0]) {
             case "package": {
-                return [args[0], ...BuildFlags.disableSandboxFlags(), ...args.slice(1)];
+                return [args[0], ...disableSandboxFlags, ...args.slice(1)];
             }
             case "build":
             case "run":
             case "test": {
-                return [...args, ...BuildFlags.disableSandboxFlags()];
+                return [...args, ...disableSandboxFlags];
             }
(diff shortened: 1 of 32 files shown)


award999:
@swift-ci please test

award999:
@swift-ci please test

award999:
@swift-ci please test

award999:
@swift-ci please test

src/toolchain/BuildFlags.ts
```diff
@@ -192,6 +192,44 @@ export class BuildFlags {
...
+    withAdditionalFlags(args: string[]): string[] {
+        return this.withSwiftPackageFlags(
+            this.withDisableSandboxFlags(this.withSwiftSDKFlags(args))
+        );
+    }
+
+    /**
+     * Get flags for disabling sandboxing when running SwiftPM
+     */
+    static disableSandboxFlags(): string[] {
```
plemarquand, line 229: nit: could just be an array declared in `withDisableSandboxFlags`

plemarquand (approved):
Looks good to me. CI failure on 6.0 linux should hopefully be resolved by https://github.com/swiftlang/vscode-swift/pull/1387

src/toolchain/BuildFlags.ts
award999: done

PR #1388, opened 2025-02-13
Fix nightly debugging test failures

matthewbastien:
Recent changes to launch configuration options and prompts for CodeLLDB broke the nightly tests. Fix those issues here by:

- Updating tests to use the `swift` launch config type instead of `swift-lldb`
- Use the new setting `swift.debugger.debugAdapter` to set the debug adapter that the tests use
- Add a new setting called `swift.debugger.setupCodeLLDB` that controls how CodeLLDB settings are updated. Tests now set this to `"alwaysUpdateGlobal"` so that they don't have to deal with the prompt at all.

I also had to disable the use of `liblldb` in tests with Swift versions <5.10 since those versions of `liblldb` don't seem to be compatible with CodeLLDB. The tests fail to start due to the debug adapter crashing.

### DIFF
--- a/assets/test/.vscode/launch.json
+++ b/assets/test/.vscode/launch.json
@@ -1,7 +1,7 @@
 {
     "configurations": [
         {
-            "type": "swift-lldb",
+            "type": "swift",
             "request": "launch",
             "name": "Debug PackageExe (defaultPackage)",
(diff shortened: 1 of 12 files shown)


award999:
@matthewbastien pre-6.0 you may have to disable these tests or use codelldb

award999 (approved):

Issue #1335, opened 2025-01-24
Wrong SDK updated when Select Target Platform is used


rbenegal:
**Describe the bug**
Support for Swift SDKs was added in #1191 however it seems like the wrong SDK may be updated.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a new `Library` project with `>Swift:  Create new project` such as `MyLibrary`
2. Put some code in MyLibrary.swift such as:

``` swift
import SwiftUI

public struct MyLibraryView: View {
  public init() {}  // Add an initializer if needed

    public var body: some View {
    VStack {
        Text("Hello")
    }
  }
}
```

4. Use `Select Target Platform` to switch to `iOS`
5. Reload Extensions using the prompt popup
6. Try to use code completion by changing Text("Hello") to something like

```swift
Text("Hello").frame(width: 100, height: 100, alignment: .center)
```

Code will not work, as it seems like the wrong SDK is being updated:

![Image](https://github.com/user-attachments/assets/faa7e639-a201-496c-97e0-2c8e5f4ca221)

Also, it would be good to reduce to just one SDK setting to avoid confusion in the future.

.**Expected behavior**
Code completion should work correctly.

**Environment**
 - OS: macOS 15.1
 - Swift version (output of `swift --version`)
 - swift-driver version: 1.116.1 Apple Swift version 6.1 (swiftlang-6.1.0.101.37 clang-1700.0.4.30)
Target: arm64-apple-macosx15.0
 - Visual Studio Code version: 1.96.4
 - vscode-swift version: 2.0.0

rbenegal:
In order to get things to work on iOS, I had to set `swift.swiftSDK` to `arm64-apple-ios` and click `Reload the extensions` on the prompt that pops up:

<img width="779" alt="Image" src="https://github.com/user-attachments/assets/bdf484e8-2c71-4342-9040-08f4829912bd" />

As well, once the settings are modified manually it seems like doing a `Select Target Platform` may no longer work to switch things back to another platform.

Once the settings are set correctly code completion and availability warnings should show up.

<img width="802" alt="Image" src="https://github.com/user-attachments/assets/d49b562e-ba0e-4435-b267-e6e7c7f44ec5" />

matthewbastien:
Verified with a1a6b0f

PR #1390, opened 2025-02-14
Set swift.swiftSDK for target platforms

rbenegal:
SourceKit-LSP now provides code editing support for non-macOS Darwin platforms starting Swift 6.1 using the `--swift-sdk` flag. Set this flag to the appropriate target-triple when using the `Select Target Platform` feature on macOS.

Issue: #1335

### DIFF
--- a/src/commands/switchPlatform.ts
+++ b/src/commands/switchPlatform.ts
@@ -19,7 +19,6 @@ import {
     getDarwinTargetTriple,
 } from "../toolchain/toolchain";
 import configuration from "../configuration";
-import { Version } from "../utilities/version";
 import { WorkspaceContext } from "../WorkspaceContext";
 
 /**
@@ -39,12 +38,7 @@ export async function switchPlatform(ctx: WorkspaceContext) {
         }
     );
     if (picked) {
-        if (ctx.toolchain.swiftVersion.isLessThan(new Version(6, 1, 0))) {
-            vscode.window.showWarningMessage(
-                "Code editing support for non-macOS platforms is only available starting Swift 6.1"
-            );
-        }
-        // show a status item as getSDKForTarget can sometimes take a long while to xcrun for the SDK
+        // show a status item as getSDKForTarget can sometimes take a long while to run xcrun to find the SDK
         const statusItemText = `Setting target platform to ${picked.label}`;
         ctx.statusItem.start(statusItemText);
         try {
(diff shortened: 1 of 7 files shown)


src/commands/switchPlatform.ts
```diff
@@ -29,28 +35,39 @@ export async function switchPlatform() {
...
+        // show a status item as getSDKForTarget can sometimes take a long while to xcrun for the SDK
+        const statusItemText = `Setting target platform to ${picked.label}`;
+        ctx.statusItem.start(statusItemText);
         try {
-            const sdkForTarget = picked.value
-                ? await SwiftToolchain.getSDKForTarget(picked.value)
-                : "";
-            if (sdkForTarget !== undefined) {
-                if (sdkForTarget !== "") {
-                    configuration.sdk = sdkForTarget;
```
rbenegal, line 42: Not sure if we should keep `configuration.sdk` for toolchains less than 6.1 or if we should just avoid setting it entirely since it doesn't provide any code editing support from SourceKit-LSP. Also it seems like we'll need either `configuration.sdk` or `configuration.swiftSDK` set, but not both together.

matthewbastien: > Not sure if we should keep `configuration.sdk` for toolchains less than 6.1 or if we should just avoid setting it entirely since it doesn't provide any code editing support from SourceKit-LSP.

I'm leaning towards just hiding this command altogether on unsupported toolchain and OS combinations instead of showing a warning notification. The setting description in the `package.json` should probably have a warning about this as well.

> Also it seems like we'll need either `configuration.sdk` or `configuration.swiftSDK` set, but not both together.

I'm not entirely sure of the difference between the two, but I'm guessing `configuration.sdk `is for older Swift versions that don't support the `--swift-sdk` option? In which case we could even change the behaviour of this command depending on the Swift version.

rbenegal: Thank you, yes good points, seems like it only makes sense to really show the option for supported combinations (currently only Darwin-based platforms). Potentially in the future it would make sense to have a list of common target-triples for other platforms as well to enable convenient switching.

It doesn't look like setting `configuration.sdk` makes much difference for any code editing/SourceKit support functionality. Perhaps it did something for much older toolchains, but it doesn't seem to do much in terms of code editing support for 5.10 or 6 at least. Probably better to avoid setting it to something automatically to prevent confusion.

matthewbastien (approved):

PR #1396, opened 2025-02-19
Add PrivateFrameworks as additional Swift Testing framework path

plemarquand:
This is essentially the same change made in:
https://github.com/swiftlang/swift-package-manager/pull/8199

VS Code would be able to leverage it for regular runs with `swift test` but because when debugging we launch the xctest executable directly to run tests, we must include the same framework search paths as SPM.


stmontgomery (approved):

award999:
Verified 55504137c17a004af8299429bda74b2c19498e65

Issue #1393, opened 2025-02-16
Missing Test Result output on Linux when using `print`


RandomHashTags:
**Describe the bug**
Title describes it.

**To Reproduce**
1. Copy paste the following code
2. Run the test by clicking the run button for `yup` provided by vscode-swift

NOT reproducible:
- on macOS
- executing `swift test` from terminal

```swift
struct League3Tests {
    @Test func yup() {
        var string:String = ""
        for i in 1...2000 {
            string += "\(i)\n"
        }
        print(string)
    }
}
```

Which always produces:

<img src='https://global.discourse-cdn.com/swift/original/3X/5/e/5e259320838f6e4be3e3d650c8928224cdc77b83.png'>

Without line breaks:

![Image](https://github.com/user-attachments/assets/4637ae4c-aaa7-4f42-a805-937025447710)

**Expected behavior**
Printing of the whole string and test result output with no missing content.

**Environment**
 - OS: Arch Linux (6.13.2-arch1-1)
 - Swift version (swift-6.0.3-RELEASE)
 - Visual Studio Code version: 1.96.4-1
 - vscode-swift version: 2.0.0

**Additional context**
Forum post: https://forums.swift.org/t/77808

plemarquand:
Thanks for the report, with the simple reproduction case! I've tracked this back to a longstanding issue in node-pty, https://github.com/microsoft/node-pty/issues/72. We'll have to see if we can work around it. VSCode uses the same module to drive its built in terminal, so maybe they've already worked around it in https://github.com/microsoft/vscode/blob/3ccdd1e63d4819349903e17019264b954e56db44/src/vs/workbench/contrib/terminal/node/terminalProcess.ts

rbenegal:
Verified with the latest pre-release version `2.1.20250321` (commit 8a0a18c) that the test outputs data past 826 now:

<img width="517" alt="Image" src="https://github.com/user-attachments/assets/d9cc2630-8c33-4a04-a250-0e3ccd5d0db3" />

Ran on Ubuntu 24.04.

PR #1401, opened 2025-02-23
Fix missing Test Result output on Linux when using `print`

plemarquand:
node-pty on Linux suffers from a long standing issue where the last chunk of output before a program exits is sometimes dropped, especially if that program produces a lot of output immediately before exiting. See https://github.com/microsoft/node-pty/issues/72

For swift processes that don't require input and may produce a lot of output that could potentially be truncated we can add a new `ReadOnlySwiftProcess` that spawns a child process using node's built in `child_process`. These spawned child processes emit their full output before exiting without the need to resort to flakey workarounds on the node-pty process.

When creating new `SwiftExecution`s in future `readOnlyTerminal` should be set to `true` unless the process may require user input.

Issue: #1393

### DIFF
--- a/assets/test/defaultPackage/Tests/PackageTests/PackageTests.swift
+++ b/assets/test/defaultPackage/Tests/PackageTests/PackageTests.swift
@@ -102,6 +102,14 @@ struct MixedSwiftTestingSuite {
   }
   #expect(2 == 3)
 }
+
+@Test func testLotsOfOutput() {
+  var string = ""
+  for i in 1...100_000 {
+    string += "\(i)\n"
+  }
+  print(string)
+}
 #endif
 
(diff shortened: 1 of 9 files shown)


plemarquand:
I initially tried to work around the issue in node-pty using suggestions in related issues and [VS Code itself](https://github.com/microsoft/vscode/blob/3ccdd1e63d4819349903e17019264b954e56db44/src/vs/workbench/contrib/terminal/node/terminalProcess.ts#L135-L140) but because we were using node-pty as a wrapper around a single execution and not as an actual psudoterminal these techniques didn't work. 

I like the final approach better anyway since wrapping one off executions like this in a pty seems like overkill.

matthewbastien (approved):

PR #1406, opened 2025-02-25
Use `lldb-dap` by default when `swift.debugger.debugAdapter` is set to `"auto"`

matthewbastien:
Now that we've removed the dependency on CodeLLDB to avoid VS Code warnings about untrusted extensions, we should use `lldb-dap` by default on Swift 6.0 and later.


plemarquand (approved):

plemarquand:

Issue #1365, opened 2025-02-09
Configurable Command Plugin Build Configuration


MahdiBM:
[Configurable Command Plugin Permissions](https://github.com/swiftlang/vscode-swift/pull/1297) changes are working very nicely and have resolved the [other issue](https://github.com/swiftlang/vscode-swift/issues/1277) I had filed.

We should add build configuration to the list of configurable settings too, for heavier operations.

Example:

```bash 
time swift package -c release --allow-writing-to-package-directory generate-code-from-openapi
...
________________________________________________________
Executed in   32.80 secs    fish           external
   usr time   49.43 secs    0.11 millis   49.43 secs
   sys time    0.81 secs    1.46 millis    0.80 secs
```

```bash 
time swift package --allow-writing-to-package-directory generate-code-from-openapi
...
________________________________________________________
Executed in   68.68 secs    fish           external
   usr time  115.34 secs    0.12 millis  115.34 secs
   sys time    1.44 secs    1.74 millis    1.44 secs
```

The non-release version takes more than twice as much to finish, wasting ~36 seconds.

This is with [hello-world-async-http-client-example](https://github.com/apple/swift-openapi-generator/tree/main/Examples/hello-world-async-http-client-example) and replacing openapi.yaml with [GitHub's](https://raw.githubusercontent.com/github/rest-api-description/main/descriptions-next/api.github.com/api.github.com.yaml) on my M1 Macbook Pro.

MahdiBM:
Hmm I'd say yes for this one? It should also be fine if you were to implement this in a more free-form way. Like accepting `package-command-args`. But it should be separate for each plugin as most plugins are not worth release-building.

award999:
we can add a setting for default package command args similar to what is done for package resolve

award999:
Verified with 82fb69c

PR #1409, opened 2025-02-26
Configurable Command Plugin Build Configuration

plemarquand:
Packages can define their own plugins either directly or through their dependencies. These plugins define commands, and the extension exposes a list of these when you use `> Swift: Run Command Plugin`.

It may be desirable to build and run these plugins with specific arguments. This patch introduces a new setting that can be specified globally or on a per workspace folder basis that allows users to configure arguments to pass to plugin command invocations.

The setting is defined under `swift.pluginArguments`, and is specified as an object in the following form:

```json
{
  "PluginCommandName:intent-name": ["--some", "--args"]
}
```

- The top level string key is the command id in the form `PluginCommandName:intent-name`. For instance, swift-format's format-source-code command would be specified as `swift-format:format-source-code`
- Specifying `PluginCommandName` will apply the arguments to all intents in the command plugin
- Specifying `*` will apply the arguments to all commands.

This patch also adds this wildcard functionality to the `swift.pluginPermissions` setting.

Issue: #1365

### DIFF
--- a/package.json
+++ b/package.json
@@ -418,6 +418,30 @@
             "default": true,
             "markdownDescription": "Controls whether or not the extension will contribute environment variables defined in `Swift: Environment Variables` to the integrated terminal. If this is set to `true` and a custom `Swift: Path` is also set then the swift path is appended to the terminal's `PATH`."
           },
+          "swift.pluginArguments": {
+            "default": [],
+            "markdownDescription": "Configure a list of arguments to pass to command invocations. This can either be an array of arguments, which will apply to all command invocations, or an object with an with command names as the key where the value is an array of arguments.",
+            "scope": "machine-overridable",
+            "oneOf": [
+              {
+                "type": "array",
+                "items": {
+                  "type": "string"
+                }
+              },
+              {
+                "type": "object",
+                "patternProperties": {
+                  "^([a-zA-Z0-9_-]+(:[a-zA-Z0-9_-]+)?)$": {
+                    "type": "array",
+                    "items": {
+                      "type": "string"
+                    }
+                  }
+                }
+              }
+            ]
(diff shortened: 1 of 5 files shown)


plemarquand:
@MahdiBM please take a look and let me know if this addresses all your concerns from https://github.com/swiftlang/vscode-swift/issues/1365

plemarquand:
@MahdiBM I think because for backwards compatibility sake, because we've already released with pluginPermissions, we shouldn't roll it in to the more generic pluginArguments. It would also make the [settings validation](https://github.com/swiftlang/vscode-swift/blob/main/package.json#L430) more tricky since it could be either one of the permissions values, or a freeform argument.

matthewbastien (approved):

package.json
```diff
@@ -418,6 +418,30 @@
             "default": true,
             "markdownDescription": "Controls whether or not the extension will contribute environment variables defined in `Swift: Environment Variables` to the integrated terminal. If this is set to `true` and a custom `Swift: Path` is also set then the swift path is appended to the terminal's `PATH`."
           },
+          "swift.pluginArguments": {
+            "default": [],
+            "markdownDescription": "Configure a list of arguments to pass to command invocations. This can either be an array of arguments, which will apply to all command invocations, or an object with an with command names as the key where the value is an array of arguments.",
```
rbenegal, line 423:
```suggestion
            "markdownDescription": "Configure a list of arguments to pass to command invocations. This can either be an array of arguments, which will apply to all command invocations, or an object with command names as the key where the value is an array of arguments.",
```

rbenegal (approved):

award999:
Verified with 82fb69c

PR #1411, opened 2025-02-26
Add snippets for debug configurations

matthewbastien:
Add three new Swift debug configuration snippets:

- Launch process
- Attach by process name
- Attach by process ID

Also, added a process picker command that can be used to select a process to attach to. The logic is largely the same as the process picker in the `vscode-js-debug` extension created by Microsoft. It will use available executables based on the current platform to find the list of available processes:

- **Linux**: uses the `ps` executable to list processes
- **macOS**: nearly identical to Linux except that the command line options passed to `ps` are different
- **Windows**: uses the PowerShell cmdlet `Get-CimInstance` to query for processes

I manually tested this on a MacBook Pro running macOS Sequoia, a Windows 11 VM, and an Ubuntu 22.04 VM.


plemarquand (approved):

src/process-tree/index.ts
```diff
@@ -0,0 +1,50 @@
...
+    command: string;
+
+    /** The full command including arguments that was used to start the process */
+    arguments: string;
+
+    /** The date when the process was started */
+    start: number;
+}
+
+export interface ProcessTree {
```
plemarquand, line 36: nit: this isn't really a ProcessTree so much as its a ProcessList

plemarquand (approved):

src/process-tree/index.ts
matthewbastien: changed to `ProcessList`

plemarquand:
Verification depends on https://github.com/swiftlang/vscode-swift/issues/1464

plemarquand:
Verified in a1a6b0fc69b229280cbef82ca13bf410a6e03571

Issue #1419, opened 2025-03-03
Setting to specify the Visual Studio tool directory


DougGregor:
**Is your feature request related to a problem? Please describe.**
If there are multiple versions of Visual Studio installed, Swift (via Clang) will pick one of them based on a complicated series of checks for environment variables / path / config files / registry. If it picks the wrong one, it's hard to know how to select the one you wanted.

**Describe the solution you'd like**
Can we add a setting to the extension that lets us directly specify the path to the Visual Studio tools that Swift should use? The most straightforward way seems to be to set the `VCToolsInstallDir` environment variable.

**Describe alternatives you've considered**
Setting the environment outside of VS Code, in `tasks.json`, or with unsafe flags in SwiftPM. None of them are as discoverable as a setting would be, especially given that we have an SDK path setting already.

plemarquand:
Would using the `swift.swiftEnvironmentVariables` setting to set `VCToolsInstallDir` be enough here? If so, we can more formally document this as an approach to select which version of Visual Studio is used.

DougGregor:
That's probably fine if we document it somewhere that folks can find.

plemarquand:
This is now documented in the [Extension Settings](https://github.com/swiftlang/vscode-swift/blob/main/docs/settings.md#windows-development) documentation.

PR #1421, opened 2025-03-03
Document how to use a different Visual Studio

plemarquand:
Add some documentation to the settings docs that outlines how to specify which Visual Studio installation Swift should pull its dependencies from on Windows.

Also annotates existing setting names in the document with links that jump straight to the relevent setting in VS Code.

Issue: #1419

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -6,9 +6,9 @@ This document outlines useful configuration options not covered by the settings 
 
 ## Command Plugins
 
-Swift packages can define [command plugins](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/Plugins.md) that can perform arbitrary tasks. For example, the [swift-format](https://github.com/swiftlang/swift-format) package exposes a `format-source-code` command which will use swift-format to format source code in a folder. These plugin commands can be invoked from VS Code using `> Swift: Run Command Plugin`. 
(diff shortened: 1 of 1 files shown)


docs/settings.md
```diff
@@ -51,17 +51,25 @@ Alternatively, you can define a task in your tasks.json and define permissions d
...
 ]
 ```
+
+## Windows Development
+
+### Specifying a Visual Studio installation
+
+Swift depends on a number of developer tools when running on Windows, including the C++ toolchain and the Windows SDK. Typically these are installed with [Visual Studio](https://visualstudio.microsoft.com/).
+
+If there are multiple versions of Visual Studio installed you can specify which one should be used by setting a new the `VCToolsInstallDir` envrionment variable to you desired path using the [`swift.swiftEnvironmentVariables`](vscode://settings/swift.swiftEnvironmentVariables) setting.
```
rbenegal, line 75:
```suggestion
If you have multiple versions of Visual Studio installed you can specify the path to the desired version by setting a `VCToolsInstallDir` environment variable using the [`swift.swiftEnvironmentVariables`](vscode://settings/swift.swiftEnvironmentVariables) setting.
```

(Just rewriting with some suggestions to fix typos, feel free to change it to something else)

rbenegal (approved):

Issue #1250, opened 2024-12-08
Slow startup


MahdiBM:
When opening VS Code, it takes 10-20s for VS Code to allow me to run a build task, or run a command plugin.
I'd like it to optimally be under 5s.

When opening the build tasks list (CMD+SHIFT+B) it does not list the tasks before the loading is complete, and it won't update the tasks list when the loading is complete, so I have to keep opening the build tasks list until it finally shows the build tasks.

I'd be fine with being able to immediately queueing a build task, but having to wait 10-20s before the build task starts so the extension startup and the package loading the complete.

The same goes for testing. The testing tab does now show up until the package loading is complete (I think it also requires a build? or at least it used to).

adam-fowler:
Can you check to see how long `swift package describe` takes to run in your project.

MahdiBM:
@adam-fowler to be clear I've noticed this on both big and small projects so I don't think it's related to project size.

First run on a bunch of different projects, it takes 1-1.5s. Next runs 0.5s or less, maybe 0.3s on avg.

adam-fowler:
I ask because for a while Soto was taking 15 seconds to run that command. The other thing to check is how long all your other extensions take to load. If you disable everything but the swift extension does it still take as long? 

When you bring up the build task list VSCode will attempt to activate any extension that might include a build task.

plemarquand:
`> Developer: Show Running Extensions` will show each extensions startup time.

I did notice that the initial `swift -print-target-info` takes upwards of 4.5s during extension activation, which is odd to me since running that command on the command line takes only ~150ms. Not sure how to account for the difference yet.

MahdiBM:
See these 2 screen recordings about the slownesses and me trying to access plugins (watch at 2x+):

This one is my normal set up in Penny (background build on, resolve packages not disabled, sourcekit-lsp backgroundPreparationMode set to 'build'): 
[Screen Recording 1 Compressed.mp4.zip](https://github.com/user-attachments/files/18429492/Screen.Recording.1.Compressed.mp4.zip)

This one I think I tried 2 things in the video. Both with background build off:
[Screen Recording 2 Compressed.mp4.zip](https://github.com/user-attachments/files/18429549/Screen.Recording.2.Compressed.mp4.zip)

MahdiBM:
The extension loading times are reasonable (~100ms).

PR #1422, opened 2025-03-04
Avoid blocking folder addition on package loading

plemarquand:
When a folder is added to a VSCode workspace the extension performs a `swift package describe` to load information about the package. This operation must complete before the folder is added to the `workspaceContext`. This prevents the user from taking many actions in the extension that could be made available right away. A `swift package describe` operation can potentially take a good amount of time if package resolution has not yet been performed.

Work around this limitation by wrapping the data pulled from this `package describe` behind promises, allowing clients to wait for package resolution to complete only if they need information that is gated by it.

This was tested with a large project (https://github.com/vapor/penny-bot) and I noticed the time until defined build tasks were made available improved by about ~7-10 seconds on a fresh open with an up to date .build folder. This is even faster on fresh checkout.

Issue: #1250

### DIFF
--- a/src/FolderContext.ts
+++ b/src/FolderContext.ts
@@ -69,14 +69,12 @@ export class FolderContext implements vscode.Disposable {
     ): Promise<FolderContext> {
         const statusItemText = `Loading Package (${FolderContext.uriName(folder)})`;
         workspaceContext.statusItem.start(statusItemText);
-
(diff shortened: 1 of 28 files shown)


rbenegal (approved):
Thanks for this! Seems to be working well for me!

Issue #1395, opened 2025-02-17
Render educational notes from the compiler


daveyc123:
The Swift compiler exposes educational notes for certain diagnostics. These notes are markdown files that describe the diagnostic along with examples.

For example, the following code:

```
protocol P {}

/// A description
struct S {
  let p: some P
}
```

Results in a diagnostic ```Property declares an opaque return type, but has no initializer expression from which to infer an underlying type``` which includes an educational note detailed https://github.com/swiftlang/swift/blob/main/userdocs/diagnostics/opaque-type-inference.md. These markdown files are included in the toolchain under usr/share/doc/swift/diagnostics

Sourcekit-lsp supports sending the path to the markdown file along with the diagnostic, but the VSCode behavior is to open the educational note for editing.

Instead, it should show it as rendered markdown.

rbenegal:
Confirmed that the educational notes now show up as rendered markdown instead of markdown for editing on a recent pre-release version (version 2.1.20250327 commit 882a038) with the following swift versions on different platforms (as provided by `swiftc -v`:

```
Apple Swift version 6.1 (swiftlang-6.1.0.110.21 clang-1700.0.13.3)
Target: arm64-apple-macosx15.0
```

```
Swift version 6.1-dev (LLVM fc8af60ca909f22, Swift fc66ec88fb2737b)
Target: x86_64-unknown-windows-msvc
```

```
Swift version 6.2-dev (LLVM 162ee50b401fff2, Swift 57288d13c9f3c02)
Target: aarch64-unknown-linux-gnu
```

Note: Swift 6.1 or higher will be required to see educational notes.

PR #1423, opened 2025-03-07
Improve UX around educational notes

daveyc123:
The Swift compiler contains educational notes to further describe diagnostics [1]. These educational notes are documented in markdown files that are contained within the toolchain.

Sourcekit LSP includes a link to the local markdown file when returning diagnostics that have an associated educational note (as part of the diagnostic code). The default behaviour in VSCode is to present these as a link in the diagnostic hover, and open the editor to the markdown file when the link is clicked.

This PR updates the behaviour for educational notes to instead open the link using the markdown preview, which shows nicely rendered content. It also updates the link in the hover to show "More
 Information" instead of the code.

Issue: #1395

[1] https://github.com/swiftlang/swift/tree/main/userdocs/diagnostics

### DIFF
--- a/src/commands/openEducationalNote.ts
+++ b/src/commands/openEducationalNote.ts
@@ -1,0 +1,24 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2021-2025 the VS Code Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of VS Code Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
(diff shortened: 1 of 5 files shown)


src/commands/openEducationalNote.ts
```diff
@@ -0,0 +1,24 @@
...
+//===----------------------------------------------------------------------===//
+
+import * as vscode from "vscode";
+
+/**
+ * Handle the user requesting to show an educational note.
+ *
+ * The default behaviour is to open it in a markdown preview to the side.
+ */
+export async function openEducationalNote(markdownFile: vscode.Uri | undefined): Promise<void> {
```
award999, line 22: since it is a context command, we should hide it from the command palette

daveyc123: Fixed

award999 (approved):

Issue #1274, opened 2024-12-13
Not auto-build after command plugin modifies files


MahdiBM:
**Describe the bug**
While having the "auto-build project" feature on, I noticed a command plugin changing the files doesn't trigger a new build.

**To Reproduce**
Trigger a command plugin which modifies Swift files.
Observe no new build is triggered.

**Expected behavior**
A new build is triggered.

**Environment**
VS Code Version: 1.93.1

**Additional context**

MahdiBM:
The same thing is true for "Swift: Clean Build Folder" as well.

plemarquand:
@MahdiBM Apologies as I know this on has been languishing for a while. What VS Code Swift extension settings do you have turned on in your configuration that is performing auto build? (building on save?)

PR #1431, opened 2025-03-12
Use FileWatcher for background compilation

plemarquand:
This `swift.backgroundCompilation` setting listened for file save events from VS Code and triggered the build task. This meant that only modifications made by saving in a VS Code editor would trigger background compilation.

Switch to using a `vscode.FileWatcher` so background compilation still gets kicked off even if a file is modified from outside VS Code. This is useful if you have a command plugin that modifies source files like swift-format.

Issue: #1274

### DIFF
--- a/src/BackgroundCompilation.ts
+++ b/src/BackgroundCompilation.ts
@@ -17,6 +17,8 @@ import { getBuildAllTask } from "./tasks/SwiftTaskProvider";
 import configuration from "./configuration";
 import { FolderContext } from "./FolderContext";
 import { TaskOperation } from "./tasks/TaskQueue";
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+import throttle = require("lodash.throttle");
 
 export class BackgroundCompilation implements vscode.Disposable {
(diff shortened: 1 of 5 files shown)


src/BackgroundCompilation.ts
```diff
@@ -13,63 +13,67 @@
...
+        );
 
-            if (!folderContext) {
-                return;
-            }
+        // Throttle events since many change events can be recieved in a short time if the user
+        // does a "Save All" or a process writes several files in quick succession.
+        this.disposables.push(
+            this.workspaceFileWatcher.onDidChange(
+                throttle(
```
matthewbastien, line 59: Are you sure you want `throttle` here and not `debounce`? I would imagine that you only want to trigger background compilation once for a given group of changes that occur in quick succession.

plemarquand: Good call, done

plemarquand: Good call, done

matthewbastien (approved):

Issue #1173, opened 2024-10-28
Migrate to Node.js 20 and vsce 3


award999:
To get Dependabot back on we had to disable major updates of vsce: https://github.com/swiftlang/vscode-swift/pull/1172

We should migrate to the next major version Node.js that VSCode now ships with at the same time we migrate to vsce 3 which depends on the Node 20 engine. Should discuss if we would like to remove the Dependabot rule or if we believe we always suspect a major update would cause this issue.

roulpriya:
Can I work on this?

matthewbastien:
@roulpriya feel free to take a stab at this! The problem holding this back right now is that the `swift-docc-render` build (which happens during `npm install`) fails on NodeJS versions >18:

```
❯ npm i

npm error code 1
...
npm error ../src/parser.c:9992:3: warning: variable 'eof' set but not used [-Wunused-but-set-variable]
npm error   9992 |   START_LEXER();
npm error        |   ^
npm error ../src/tree_sitter/parser.h:135:8: note: expanded from macro 'START_LEXER'
npm error   135 |   bool eof = false;             \
npm error       |        ^
npm error 1 warning generated.
npm error In file included from ../bindings/node/binding.cc:3:
npm error ../../nan/nan.h:2548:8: error: no matching member function for call to 'SetAccessor'
npm error  2548 |   tpl->SetAccessor(
npm error       |   ~~~~~^~~~~~~~~~~
npm error /Users/<REDACTED>/Library/Caches/node-gyp/22.4.0/include/node/v8-template.h:1055:8: note: candidate function not viable: no known conversion from 'v8::AccessControl' to 'PropertyAttribute' for 5th argument
npm error  1055 |   void SetAccessor(
npm error       |        ^
npm error  1056 |       Local<Name> name, AccessorNameGetterCallback getter,
npm error  1057 |       AccessorNameSetterCallback setter = nullptr,
npm error  1058 |       Local<Value> data = Local<Value>(), PropertyAttribute attribute = None,
npm error       |                                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
...
npm error gyp ERR! cwd /Users/<REDACTED>/Documents/swiftlang/swift-docc-render/node_modules/tree-sitter-javascript
npm error gyp ERR! node -v v22.4.0
npm error gyp ERR! node-gyp -v v10.2.0
npm error gyp ERR! not ok
```

[...]

PR #1436, opened 2025-03-14
Update: Migrate to Node.js 20 and vsce 3

roulpriya:
Fixes Issue #1173

## Summary
Updating NodeJs version - 20 and VSCE version - 3

## Testing
Run `npm i`

## Checklist

- [ ] Added tests (No code change involved, hence didn't add any tests)
- [X] Ran `npm test`, and it succeeded


matthewbastien (changes requested):
Thanks for looking into this! Just a few more things that need to be changed before this can be merged:
- Update the GitHub workflows scripts in `.github/workflows` to use NodeJS 20.18.2 as that seems to be the current version of NodeJS in VS Code.
- The `scripts/update_swift_docc_render.ts` also has a check for Node 18 that can be removed.

roulpriya:
In addition to the above-given items, I modified the docker file to use node version 20.

roulpriya:
> @roulpriya thanks for the PR. Can you also update the `.nvmrc` file? On `nvm use` the correct version will automatically be installed

updated the file

scripts/update_swift_docc_render.ts
```diff
@@ -27,9 +27,9 @@ function checkNodeVersion() {
             "Unable to determine the version of NodeJS that this script is running under."
         );
     }
-    if (!semver.satisfies(nodeVersion, "18")) {
+    if (!semver.satisfies(nodeVersion, "20")) {
```
matthewbastien, line 30: Have you tried building this with node versions >20? E.g. the LTS which I believe is 22? I'm hoping we can just get rid of this whole check and allow people to build with whatever NodeJS version they have installed. The builds will continue to use 20.18.

roulpriya: Yes, this will break at node version 21, as the `tree-sitter-vue` is not maintained, we will not be able to build for node version >20.

To enable users to run it on any node version, we need to fork the tree-sitter-vue dependencies and update them.

matthewbastien (approved):
LGTM. Thank you!

award999 (approved):

Issue #1434, opened 2025-03-13
Test Failure: DiagnosticsManager Test Suite - default diagnosticsStyle failed on Linux (nightly-main - jammy)


rbenegal:
**Failure logs**

```
  DiagnosticsManager Test Suite
    Parse diagnostics
      Parse from task output
        1) default diagnosticsStyle
Captured logs during DiagnosticsManager Test Suite → Parse diagnostics → Parse from task output → default diagnosticsStyle:
Starting test: DiagnosticsManager Test Suite → Parse diagnostics → Parse from task output → default diagnosticsStyle
======== END OF LOGS ========


DiagnosticsManager Test Suite Parse diagnostics Parse from task output default diagnosticsStyle - Not all diagnostics were fulfilled. Remaining: [
 [
  "/tmp/code/assets/test/diagnostics/Sources/main.swift",
  [
   {
    "severity": "Warning",
    "message": "Initialization of variable 'unused' was never used; consider replacing with assignment to '_' or removing it",
    "range": [
     {
      "line": 1,
      "character": 8
     },
     {
      "line": 1,
      "character": 8
     }
    ],
    "source": "swiftc"
   }
  ]
 ]
]
        2) swift diagnosticsStyle
Captured logs during DiagnosticsManager Test Suite → Parse diagnostics → Parse from task output → swift diagnosticsStyle:
Starting test: DiagnosticsManager Test Suite → Parse diagnostics → Parse from task output → swift diagnosticsStyle
======== END OF LOGS ========


DiagnosticsManager Test Suite Parse diagnostics Parse from task output swift diagnosticsStyle - Not all diagnostics were fulfilled. Remaining: [
 [
  "/tmp/code/assets/test/diagnostics/Sources/main.swift",
  [
   {
    "severity": "Warning",
    "message": "Initialization of variable 'unused' was never used; consider replacing with assignment to '_' or removing it",
    "range": [
     {
      "line": 1,
      "character": 8
     },
     {
      "line": 1,
      "character": 8
     }
    ],
    "source": "swiftc"
   }
  ]
 ]
]

```

Link to failed run: https://github.com/swiftlang/vscode-swift/actions/runs/13824274954/job/38676245991

**Additional context**
Seems like this test may be flaky now, have seen it fail occasionally.

rbenegal:
[...]

PR #1440, opened 2025-03-18
Update diagnostics parsing to handle new warnings

rbenegal:
Starting with Swift 6.2 some warnings may end with a tag such as `[#no-usage]`. Update parsing logic to remove these tags from the warnings to ensure we can combine them with the warnings from SourceKit.

Issue: #1434


award999 (approved):
Awesome @rbenegal!

PR #1450, opened 2025-03-20
Build pre-release VSIX in CI

matthewbastien:
Update the `preview-package` script to generate a pre-release package with one minor version greater than the current version from the `package.json` and a patch number set to today's date formatted as `YYYYMMddhhmm`. The nightly CI will now generate both a release and pre-release extension VSIX. This enables us to create a pre-release version of the Swift extension.

Going forward, we'll have to abide by the following rules in order for this to work:

1. The version in the `package.json` should always be the current release version of the extension as published on the VS Code marketplace.
2. Always use even numbered minor versions for releases. Pre-release builds will then always have an odd minor version.

I've added a check in CI to enforce even numbered minor versions in the `package.json`.

### DIFF
--- a/scripts/lib/utilities.ts
+++ b/scripts/lib/utilities.ts
@@ -1,0 +1,42 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2024 the VS Code Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of VS Code Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+/* eslint-disable no-console */
+
+import * as child_process from "child_process";
+
+export async function exec(
+    command: string,
+    args: string[],
+    options: child_process.SpawnOptionsWithoutStdio = {}
+): Promise<void> {
+    let logMessage = "> " + command;
+    if (args.length > 0) {
+        logMessage += " " + args.join(" ");
+    }
+    console.log(logMessage + "\n");
+    return new Promise<void>((resolve, reject) => {
+        const childProcess = child_process.spawn(command, args, { stdio: "inherit", ...options });
+        childProcess.once("error", reject);
+        childProcess.once("close", (code, signal) => {
+            if (signal !== null) {
+                reject(new Error(`Process exited due to signal '${signal}'`));
+            } else if (code !== 0) {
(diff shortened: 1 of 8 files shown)


scripts/lib/utilities.ts
```diff
@@ -0,0 +1,42 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2024 the VS Code Swift project authors
```
award999, line 5: 2025

scripts/preview_package.ts
```diff
@@ -0,0 +1,61 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2024 the VS Code Swift project authors
```
award999, line 5: 2025

scripts/lib/utilities.ts
```diff
@@ -31,6 +34,30 @@ export async function main(mainFn: () => Promise<void>): Promise<void> {
     }
 }
 
+/**
+ * Returns the root directory of the repository.
+ */
+export function getRootDirectory(): string {
+    return path.join(__dirname, "..", "..");
```
rbenegal, line 41: Is it always guaranteed that root directory will be two levels above from whatever` __dirname` is?

scripts/preview_package.ts
```diff
@@ -0,0 +1,45 @@
...
+    const hour = date.getUTCHours().toString().padStart(2, "0");
+    const minutes = date.getUTCMinutes().toString().padStart(2, "0");
+    return year + month + day + hour + minutes;
+}
+
+main(async () => {
+    const rootDirectory = getRootDirectory();
+    const version = await getExtensionVersion();
+    // Increment the minor version and set the patch version to today's date
+    const minor = version.minor + 1;
```
rbenegal, line 37: Wondering if there needs to be another check here to be sure that minor version will be odd, just in case. Checking the release version is definitely more important though, so probably doesn't matter much here?

scripts/lib/utilities.ts
matthewbastien: The `utilities.ts` file is located in `scripts/lib/` which means that this will always return the root directory (unless someone moves the file of course).

scripts/preview_package.ts
matthewbastien: It definitely doesn't hurt to add a check here. Done!

rbenegal (approved):
Looks great to me!

award999 (approved):

PR #1453, opened 2025-03-20
Show error message in project panel when manifest can't be built

plemarquand:
Because we use `swift package describe` to extract metadata about the package we need the Package.swift to be buildable. If it isn't, create a header node in the project panel that calls out the fact that the Package.swift can't be built. Clicking it will take the user to the Package.swift.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -12,6 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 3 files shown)


plemarquand:
Preview of error UI:
Screenshot 2025-03-20 at 3 25 25 PM

award999 (approved):

PR #1454, opened 2025-03-21
Show all task types in project panel

plemarquand:
If users want to have non-Swift task types in their package they should be able to.


rbenegal (approved):
Yes good idea, thanks!

PR #1470, opened 2025-03-30
Use Swiftly's toolchain path if active toolchain is managed by Swiftly

plemarquand:
On macOS the toolchain path is determined by `xcrun --find swift`. `xcrun` will return paths inside the active Xcode. If the active toolchain is being managed by the newly released Swiftly 1.0 on macOS then the swift path and the toolchain path will be mismatched.

Check to see if the active toolchain is managed by Swiftly, and if so point the toolchain path to the Swiftly installed toolchain. The name of the `.xctoolchain` folder needs to be reverse engineered from Swiftly's in use toolchain name using the ToolchainVersion.parse method which was ported from Swiftly's codebase. A nice feature in the future would be to be able to ask Swiftly where its managed toolchains are stored on disk.

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -24,6 +24,7 @@ import { expandFilePathTilde, pathExists } from "../utilities/filesystem";
 import { Version } from "../utilities/version";
 import { BuildFlags } from "./BuildFlags";
 import { Sanitizer } from "./Sanitizer";
+import { SwiftlyConfig, ToolchainVersion } from "./ToolchainVersion";
 
 /**
  * Contents of **Info.plist** on Windows.
@@ -229,12 +230,13 @@ export class SwiftToolchain {
     }
 
     /**
-     * Reads the swiftly configuration file to find a list of installed toolchains.
+     * Finds the list of toolchains managed by Swiftly.
      *
      * @returns an array of toolchain paths
      */
     public static async getSwiftlyToolchainInstalls(): Promise<string[]> {
         // Swiftly is only available on Linux right now
+        // TODO: Add support for macOS
         if (process.platform !== "linux") {
             return [];
         }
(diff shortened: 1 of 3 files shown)


award999 (approved):

src/toolchain/toolchain.ts
```diff
@@ -617,6 +639,31 @@ export class SwiftToolchain {
...
+            const { stdout: swiftLocation } = await execFile("which", ["swift"]);
+            if (swiftLocation.indexOf(swiftlyHomeDir) === 0) {
+                const swiftlyConfig = await SwiftToolchain.getSwiftlyConfig();
+                if (swiftlyConfig) {
+                    const version = ToolchainVersion.parse(swiftlyConfig.inUse);
+                    return path.join(
+                        os.homedir(),
+                        "Library/Developer/Toolchains/",
+                        `${version.identifier}.xctoolchain`,
+                        "usr"
```
cmcgee1024, line 659: question: Should this be a path to the `<toolchain>/usr/bin` where the binaries like `swift` are located?

plemarquand: For consistency with other toolchain paths in VS Code they're specified as `name.xctoolchain/usr`

src/toolchain/toolchain.ts
```diff
@@ -617,6 +639,31 @@ export class SwiftToolchain {
...
+     * Determine if Swiftly is being used to manage the active toolchain and if so, return
+     * the path to the active toolchain.
+     * @returns The location of the active toolchain if swiftly is being used to manage it.
+     */
+    private static async swiftlyToolchainLocation(): Promise<string | undefined> {
+        const swiftlyHomeDir: string | undefined = process.env["SWIFTLY_HOME_DIR"];
+        if (swiftlyHomeDir) {
+            const { stdout: swiftLocation } = await execFile("which", ["swift"]);
+            if (swiftLocation.indexOf(swiftlyHomeDir) === 0) {
+                const swiftlyConfig = await SwiftToolchain.getSwiftlyConfig();
```
cmcgee1024, line 652: thought: For the in use toolchain, swiftly has a `swiftly use --print-location` command that will give the filesystem path of the one that's in-use. It can be more expensive than reading the config.json, but it also takes into consideration toolchains that are in-use because of the `.swift-version` file.

Issue #1471, opened 2025-03-31
2 warning popups when changing toolchain


award999:
**Describe the bug**
When changing the selected version of Xcode using the `Swift: Select Toolchain...` command, you get prompted with 2 warning about needing to reload the extension. Little annoying

<img width="548" alt="Image" src="https://github.com/user-attachments/assets/96424c49-16ed-43ce-a4af-9178418c90d1" />

**To Reproduce**
1. Run the command and select a different Xcode installation

**Expected behavior**
Should ideally only have one warning popup

**Environment**
 - OS: macOS
 - Swift version: N/A
 - Visual Studio Code version: N/A
 - vscode-swift version: N/A

**Additional context**
N/A

award999:
Verified with 38db866

PR #1473, opened 2025-03-31
Ensure only one reload extension dialog can be shown at a time

plemarquand:
If `showReloadExtensionNotification` is in flight and being awaited then return the promise being awaited instead of showing another dialog.

Also, debounce `showReloadExtensionNotification` with a 5 second window because its possible for users to close the first dialog very fast and have another one appear immediately if we update several settings at a time that require an extension restart.

Issue: #1471


rbenegal (approved):
Seems to work great thanks!

PR #1476, opened 2025-04-02
feat: secondary quick pick for selecting Swift Version with `runSwiftScript` command

### LINKED ISSUE #1427: `Swift: Run Swift Script` always runs in Swift 5 language mode
**Describe the bug**
When running `Swift: Run Swift Script` on single files, the extension runs effectively this command:

```bash
swift path/to/your-file.swift
```

Since Swift's default language mode is 5, this will be run in Swift 5 language mode.

To fix, the extension should provide a second "step" when you choose `Swift: Run Swift Script` that also asks about what language mode a user would like to choose.

Then use that in the command that runs the Swift file. For example the command for Swift 6 language mode would be:

```bash
swift -swift-version 6 path/to/your-file.swift
```

**Environment**
 - OS: [e.g. Ubuntu 20.04] -
 - Swift version (output of `swift --version`): 6
 - Visual Studio Code version: -
 - vscode-swift version: 2.0.2

louisunlimited:
This PR tries to implement a secondary quick pick for selecting desired Swift Version for `Run Swift Script` command, and also enabling global settings for it in vscode workspace.
- Fixes  #1427

### DIFF
--- a/package.json
+++ b/package.json
@@ -324,6 +324,19 @@
             },
             "markdownDescription": "Additional arguments to pass to `swift build` and `swift test`. Keys and values should be provided as individual entries in the list. If you have created a copy of the build task in `tasks.json` then these build arguments will not be propagated to that task."
           },
+          "swift.script.defaultSwiftVersion": {
+            "type": "number",
+            "default": null,
+            "description": "The default Swift version to use when running Swift scripts. If set, skips the version selection dialog.",
+            "enum": [
+              5,
+              6
+            ],
+            "enumDescriptions": [
+              "Swift 5",
+              "Swift 6"
+            ]
+          },
           "swift.packageArguments": {
             "type": "array",
             "default": [],
--- a/package.json
+++ b/package.json
@@ -324,7 +324,7 @@
             },
             "markdownDescription": "Additional arguments to pass to `swift build` and `swift test`. Keys and values should be provided as individual entries in the list. If you have created a copy of the build task in `tasks.json` then these build arguments will not be propagated to that task."
           },
-          "swift.script.defaultSwiftVersion": {
+          "swift.defaultSwiftVersion": {
             "type": "number",
             "default": null,
             "description": "The default Swift version to use when running Swift scripts. If set, skips the version selection dialog.",
--- a/package.json
+++ b/package.json
@@ -324,7 +324,7 @@
             },
             "markdownDescription": "Additional arguments to pass to `swift build` and `swift test`. Keys and values should be provided as individual entries in the list. If you have created a copy of the build task in `tasks.json` then these build arguments will not be propagated to that task."
           },
-          "swift.defaultSwiftVersion": {
+          "swift.scriptSwiftLanguageVersion": {
             "type": [
               "number",
               null
--- a/package.json
+++ b/package.json
@@ -337,7 +337,7 @@
               "Prompt to select the Swift version each time a script is run."
             ],
             "default": "6",
-            "markdownDescription": "The default Swift version to use when running Swift scripts. If set, skips the version selection dialog.",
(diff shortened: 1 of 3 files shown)


src/commands/runSwiftScript.ts
```diff
@@ -40,33 +40,50 @@ export async function runSwiftScript(ctx: WorkspaceContext) {
...
-        // otherwise save document
-        await document.save();
-    }
-
-    const runTask = createSwiftTask(
-        [filename],
-        `Run ${filename}`,
+    const picked = await vscode.window.showQuickPick(
+        [
+            // Potnetially add more versions here
```
plemarquand, line 45: nit: spelling
```suggestion
            // Potentially add more versions here
```

plemarquand (commented):
Thanks for the contribution! I think this is looking good so far. The only change I'd like to see it the addition of a setting that would allow users to bypass the quickPick dialog and always use the same language mode. 

The setting should be defined in the package.json of the extension, and exposed through configuration.ts, probably at the [workspace folder level](https://github.com/swiftlang/vscode-swift/blob/main/src/configuration.ts#L152) so it can be defined per package as well as globally

louisunlimited:
Hi @plemarquand , thanks for the feedback! I have included some changes to enable setting something like
` "swift.defaultSwiftVersion": 6` in settings.json to automatically take the version. However I do see that by implementing it this way we can't edit this in the UI part of setting.json only in the json file itself, do you have any recommendations on what should we give the users? Also I was wondering would it be nice to have a popup? or a line somewhere saying something along the lines of `you are using settings.json to automatically use Swift 5/6, if you want to change this behavior simply remove the line in settings.json or change it to a new value`.
Screenshot 2025-04-02 at 20 03 14

package.json
```diff
@@ -324,6 +324,15 @@
             },
             "markdownDescription": "Additional arguments to pass to `swift build` and `swift test`. Keys and values should be provided as individual entries in the list. If you have created a copy of the build task in `tasks.json` then these build arguments will not be propagated to that task."
           },
+          "swift.defaultSwiftVersion": {
```
plemarquand, line 327: This setting should be an `enum` type, with the options "6", "5", and "Ask Every Run". `enumDescriptions` should be provided for each option as well. I think the default value for the setting should be `6`.

You can see an example of an `enum` style setting here: https://github.com/swiftlang/vscode-swift/blob/main/package.json#L557. This way the user can't pick an unsupported value, they'll get a warning in the JSON editor and they'll get a nice picker UI in the settings UI page.

Also I think the name `defaultSwiftVersion` is too generic and may be confusing. This setting only applies when running Swift scripts but could be interpreted as some kind of global. I'd rename it to `scriptSwiftLanguageVersion`.

src/configuration.ts
```diff
@@ -328,6 +328,9 @@ const configuration = {
             .get<string[]>("buildArguments", [])
             .map(substituteVariablesInString);
     },
+    get defaultSwiftVersion(): number | undefined {
```
plemarquand, line 331: The type of this setting would also have to change to `string` since the valid options are `"6"`, `"5"` and `"Ask Every Run"

louisunlimited:
Hi @plemarquand thanks for the review! I have made corresponding changes, let me know if you need anything else!

plemarquand:
@louisunlimited thanks for taking a look, but I don't see any new commits. Did the changes get pushed up?

plemarquand:
@louisunlimited this is looking great! I've kicked off a test run in CI

louisunlimited:
I also removed the last sentence in setting's description a bit because now we are not showing picker based on whether this is set or not, but rather which selection the user choses.

plemarquand (approved):
Looks good! Tests are green (the Windows CI failures were unrelated to your patch and should be resolved now).

PR #1478, opened 2025-04-02
Support different toolchains per folder

plemarquand:
With the introduction of swiftly's `.swift-version` file the active toolchain can now be per-folder instead of global. If the user has multiple different packages open they may each be using a different toolchain as defined by their `.swift-version` file.

In order to support this new paradigm the `toolchain` has been moved from `WorkspaceContext` to `FolderContext`. Each time a folder (package) is added to the workspace a new toolchain is created, as it might be different from folder to folder.

The toolchain created respects the `.swift-version` file. If the toolchain specified in the `.swift-version` file is not installed an error message is shown prompting the user to install the version with swiftly.

There is still a `globalToolchain` on the `WorkspaceContext` which refers to the globally available toolchain. This would be the toolchain used when you run `swift` outside of a workspace folder. This is mainly used as a fallback toolchain for when there are no workspace folders.

It is generally advisable to use the toolchain provided on the `FolderContext` to ensure you don't end up using mismatched versions.

This PR also refactors the `LanguageClientManager` so that one instance of sourcekit-lsp is started per-toolchain, coordinating startup so that the server from a given toolchain starts up when a folder using that toolchain is added to the workspace.

While this PR adds support for .swift-version files, there is still quite a bit of work to do to make using swiftly with the VS Code Swift extension a nicer experience including: Installing swiftly directly from the extension, downloading missing toolchains automatically, listing/picking/downloading toolchains via `swiftly list`, a smoother toolchain switching experience that would optionally write the `.swift-version` file, and more.


cmcgee1024 (commented):
This all looks good from my perspective. I'm not sure about the vscode specifics, and the failing test.

src/PackageWatcher.ts
```diff
@@ -87,6 +94,16 @@ export class PackageWatcher {
         watcher.onDidCreate(async () => await this.handleWorkspaceStateChange());
         watcher.onDidChange(async () => await this.handleWorkspaceStateChange());
         watcher.onDidDelete(async () => await this.handleWorkspaceStateChange());
+
+        const fileExists = await fs
```
matthewbastien, line 98: We have an implementation of `fileExists()` already in `src/utilities/fileSystem.ts`.

src/PackageWatcher.ts
```diff
@@ -99,6 +116,43 @@ export class PackageWatcher {
...
+            const version = await this.readSwiftVersionFile();
+            if (version && version.toString() !== this.currentVersion?.toString()) {
+                this.workspaceContext.fireEvent(
+                    this.folderContext,
+                    FolderOperation.swiftVersionUpdated
+                );
+            }
+            this.currentVersion = version ?? this.folderContext.toolchain.swiftVersion;
+        } catch {
+            // do nothing
```
matthewbastien, line 142: Does it make sense to be logging errors from this in the Output Channel at the very least?

src/PackageWatcher.ts
```diff
@@ -99,6 +116,43 @@ export class PackageWatcher {
...
+        }
+    }
+
+    private async readSwiftVersionFile() {
+        const versionFile = path.join(this.folderContext.folder.fsPath, ".swift-version");
+        try {
+            const contents = await fs.readFile(versionFile);
+            return Version.fromString(contents.toString().trim());
+        } catch {
+            return undefined;
```
matthewbastien, line 152: Same here, except we should probably check `fileExists()` beforehand.

src/commands.ts
```diff
@@ -146,9 +146,13 @@ export function register(ctx: WorkspaceContext): vscode.Disposable[] {
         ),
         vscode.commands.registerCommand(Commands.RUN_PLUGIN_TASK, () => runPluginTask()),
         vscode.commands.registerCommand(Commands.RUN_TASK, name => runTask(ctx, name)),
-        vscode.commands.registerCommand("swift.restartLSPServer", () =>
-            ctx.languageClientManager.restart()
-        ),
+        vscode.commands.registerCommand("swift.restartLSPServer", async () => {
+            if (!ctx.currentFolder) {
```
matthewbastien, line 150: If there's more than one SourceKit-LSP instance, should we be prompting the user to select which SourceKit-LSP they want to restart? Otherwise they'd have to know to open a file within a folder before running this command which could lead to some confusion.

src/sourcekit-lsp/LanguageClientToolchainCoordinator.ts
```diff
@@ -0,0 +1,131 @@
...
+import { FolderContext } from "../FolderContext";
+import { LanguageClientFactory } from "./LanguageClientFactory";
+import { LanguageClientManager } from "./LanguageClientManager";
+import { FolderOperation, WorkspaceContext } from "../WorkspaceContext";
+
+/**
+ * Manages the creation of LanguageClient instances for workspace folders.
+ *
+ * A LanguageClient will be created for each unique toolchain version. If two
+ * folders share the same toolchain version then they will share the same LanaugeClient.
```
matthewbastien, line 26:
```suggestion
 * folders share the same toolchain version then they will share the same LanguageClient.
```

src/sourcekit-lsp/LanguageClientToolchainCoordinator.ts
```diff
@@ -0,0 +1,131 @@
...
+ * This ensures that a folder always uses the LanaugeClient bundled with its desired toolchain.
```
matthewbastien, line 27:
```suggestion
 * This ensures that a folder always uses the LanguageClient bundled with its desired toolchain.
```

src/sourcekit-lsp/LanguageClientToolchainCoordinator.ts
```diff
@@ -0,0 +1,131 @@
...
+            throw new Error(
+                "LanguageClientManager has not yet been created. This is a bug, please file an issue at https://github.com/swiftlang/vscode-swift/issues"
+            );
+        }
+        return client;
+    }
+
+    /**
+     * Stops all LanguageClient instances.
+     * This should b called when the extension is deactivated.
```
matthewbastien, line 101:
```suggestion
     * This should be called when the extension is deactivated.
```

src/WorkspaceContext.ts
```diff
@@ -248,19 +248,21 @@ export class WorkspaceContext implements vscode.Disposable {
...
-                return;
-            }
-            contextKeys.supportsReindexing =
-                experimentalCaps[ReIndexProjectRequest.method] !== undefined;
-            contextKeys.supportsDocumentationLivePreview =
-                experimentalCaps[DocCDocumentationRequest.method] !== undefined;
-        });
+        if (this.currentFolder) {
+            const languageClient = this.languageClientManager.get(this.currentFolder);
+            languageClient.useLanguageClient(async client => {
```
matthewbastien, line 253: Not an issue introduced by your patch, but I'm pretty sure this just does nothing when SourceKit-LSP is disabled in settings. `useLanguageClient()` should indicate when there's no language client to run in this case so that we can set `supportsReindexing` and `supportsDocumentationLivePreview` to false. This would only really be an issue the moment after the user disables SourceKit-LSP in their settings since the default is false.

src/commands/dependencies/resolve.ts
```diff
@@ -47,7 +47,7 @@ export async function resolveFolderDependencies(
             prefix: folderContext.name,
             presentationOptions: { reveal: vscode.TaskRevealKind.Silent },
         },
-        folderContext.workspaceContext.toolchain
+        folderContext.toolchain
```
matthewbastien, line 50: Another existing issue prior to this PR: `resolveDependencies()` only accepts the workspace context as a parameter. However, what happens if there are multiple folders available? It probably makes sense to prompt the user which folder here, but have things like the Project Panel provide a specific folder as the argument when triggered.

I imagine this is a more widespread issue with the repo that might warrant its own issue/PR.

src/debugger/debugAdapterFactory.ts
```diff
@@ -91,14 +91,19 @@ function registerLLDBDebugAdapter(
 export class LLDBDebugConfigurationProvider implements vscode.DebugConfigurationProvider {
     constructor(
         private platform: NodeJS.Platform,
-        private toolchain: SwiftToolchain,
+        private workspaceContext: WorkspaceContext,
         private outputChannel: SwiftOutputChannel
     ) {}
 
     async resolveDebugConfigurationWithSubstitutedVariables(
         _folder: vscode.WorkspaceFolder | undefined,
```
matthewbastien, line 99: Should probably remove the underscore here since it was only there to silence the unused argument warning.
```suggestion
        folder: vscode.WorkspaceFolder | undefined,
```

src/debugger/debugAdapterFactory.ts
```diff
@@ -91,14 +91,19 @@ function registerLLDBDebugAdapter(
...
         launchConfig: vscode.DebugConfiguration
     ): Promise<vscode.DebugConfiguration | undefined | null> {
+        const folder = this.workspaceContext.folders.find(
```
matthewbastien, line 102: Related to above: you would also have to rename this variable to something like `folderContext`.

src/documentation/DocumentationPreviewEditor.ts
```diff
@@ -203,8 +203,17 @@ export class DocumentationPreviewEditor implements vscode.Disposable {
                 return;
             }
 
+            const folderContext = this.context.folders.find(folderContext =>
+                document.uri.fsPath.startsWith(folderContext.folder.fsPath)
+            );
+
+            if (!folderContext) {
+                return;
```
matthewbastien, line 211: Should probably log this as an error or show an error to the user. Otherwise, the button/command that they ran to get to this point would appear to do nothing.

matthewbastien (changes requested):
Overall it looks great, thanks! Some minor issues here and there as well as a general question about prompting when there are multiple folders/SourceKit-LSP clients.

src/PackageWatcher.ts
plemarquand: General guidance tends to be try a file operation and handle the failure instead of checking for existence first, because in the time between the `exists()` call returning and the file operation beginning the file might be removed. But logging, yes.

src/commands.ts
plemarquand: This is a really good point. Do we want to use the current folder for these commands if there is one, and show a picker if there isn't a current folder? Or do we always show a picker? I'm leaning towards option 1.

plemarquand: I added a [folder quick picker](https://github.com/swiftlang/vscode-swift/pull/1478/files#diff-06fafcfb4ede7cbf03a0cb5962d2ce5779e579bf307253731d4b41771e752c2b) as a fallback for when there is no swift currentFolder. You can test with no current folder by opening a second workspace folder that isn't a swift project and clicking in it, then running Swift > Restart LSP Server. 

There isn't a way today to run `resolveDependencies` without a current folder (its only able to be triggered from the project panel, and that disappears when there is no swift current folder) but I've updated that code anyway in case we add the ability to resolve dependencies via the command palette in future.

matthewbastien: > Do we want to use the current folder for these commands if there is one, and show a picker if there isn't a current folder? Or do we always show a picker?

I think we should always show the quick pick, but have the current folder be the default selected option with an additional option to restart everything. Might be worthwhile to add a setting to configure the default behaviour to avoid frustration (though I hope people aren't needing to use this command a lot).

> There isn't a way today to run resolveDependencies without a current folder (its only able to be triggered from the project panel, and that disappears when there is no swift current folder) but I've updated that code anyway in case we add the ability to resolve dependencies via the command palette in future.

Something to consider for the future, yeah. When running from the command palette I would definitely expect a quick pick to be shown with the available options if there are more than one. Same as above for SourceKit-LSP.

src/commands/restartLSPServer.ts
```diff
@@ -0,0 +1,59 @@
...
+        description: toolchainLookup[key][0].fullToolchainName,
+        detail: toolchainLookup[key].map(({ folder }) => folder.name).join(", "),
+    }));
+
+    // Skip picking a toolchain if there is only one option to pick
+    if (toolchains.length === 1) {
+        return restartLSP(ctx, toolchains[0].label);
+    }
+
+    const selected = await vscode.window.showQuickPick(toolchains, {
```
matthewbastien, line 43: Should probably also add an option to restart all of them.

plemarquand: Done

PR #1486, opened 2025-04-07
Support all supported swift language versions in swift.scriptSwiftLanguageVersion

plemarquand:
For completeness, include all supported language versions

### DIFF
--- a/package.json
+++ b/package.json
@@ -329,8 +329,8 @@
             "enum": [
               "6",
               "5",
-              "4",
               "4.2",
+              "4",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -329,6 +329,8 @@
             "enum": [
               "6",
               "5",
+              "4",
+              "4.2",
```
MahdiBM, line 333: nit: is this order fine or should 4.2 be higher?

award999 (approved):

PR #1498, opened 2025-04-11
Handle Info.plist encoded as `utf16le`

plemarquand:
The March 10th nightly toolchain has the Info.plist file encoded using `utf16le`, which we fail to parse as we expect it to be in `utf8`. Try `utf8`  first, and fallback to `utf16le` if we can't parse the file.

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -830,12 +830,24 @@ export class SwiftToolchain {
             );
             return undefined;
         }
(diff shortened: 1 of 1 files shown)


matthewbastien (commented):
Is there any way to add an integration test that would be able to verify this? Or is it actually failing in nightly?

plemarquand:
Closing in favour of the upstream fix: https://github.com/swiftlang/swift/pull/80775

PR #1513, opened 2025-04-22
Add system-target and binary target icons to Project Panel

plemarquand:
Missed a few target types when adding icons to the Project Panel. Also adds the target type description to the tooltip to make it more clear what icons correspond to what target type.

### DIFF
--- a/src/ui/ProjectPanelProvider.ts
+++ b/src/ui/ProjectPanelProvider.ts
@@ -239,6 +239,7 @@ class TargetNode {
         item.iconPath = new vscode.ThemeIcon(this.icon());
         item.contextValue = this.contextValue();
         item.accessibilityInformation = { label: name };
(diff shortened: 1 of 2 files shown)


src/ui/ProjectPanelProvider.ts
```diff
@@ -252,6 +253,12 @@ class TargetNode {
                 return "output";
             case "library":
                 return "library";
+            case "system-target":
+                return "server";
+            case "binary":
+                return "file-binary";
+            case "plugin":
+                return "plug";
```
award999, line 261: not "plugin"?

award999 (approved):

src/ui/ProjectPanelProvider.ts
matthewbastien: The VS Code icon is called `plug`: https://code.visualstudio.com/api/references/icons-in-labels

PR #1516, opened 2025-04-23
Poll for updated tasks in project panel

plemarquand:
When a task is updated, added or deleted it is not reflected in the project panel. There are no VS Code APIs that notify on task modification/update/deletion, so poll every 2 seconds and check if the tasks are updated. If they are, refresh the tasks list in the project panel.


matthewbastien:
Have you tried adding a file watcher on the `tasks.json`? Though I suppose other extensions can add or remove tasks at will outside of the `tasks.json`.

plemarquand:
That was my initial idea, but the tasks the Swift extension provides (Build All, etc...) are liable to change if we want to support changing toolchains/settings on the fly without restarting.

award999 (approved):

PR #606, opened 2023-09-12
Toolchain path is one folder up from swift bin path

adam-fowler:
Previously it was two folders up and swift executables were assumed to be in a `usr/bin` folder relative to the toolchain folder. It looks like this cannot be assumed. The arm64 toolchains generated by @futurejones are installed at `/usr/libexec/swift` and have their binaries folder in `/usr/libexec/swift/bin`. So the toolchain path is now set to one folder up instead and the binaries folder is set to be `bin` relative to the toolchain folder.

I have checked the sourcekit-lsp code to see if it is happy with this as a valid toolchain path. It appears to be, but will verify with developers

PR #1521, opened 2025-04-24
Update DYLD_LIBRARY_PATH to fix debugging tests in nightly toolchains

plemarquand:
A change in SwiftPM removed the rpath that pointed to `usr/lib/swift/macosx/testing`
(https://github.com/swiftlang/swift-package-manager/pull/8295).

Consequently when debugging tests in VS Code `libTesting.dylib` could not be found and the tests would not start. This is only happening in nightly toolchains that contain the SwiftPM change noted above.

Fix this by adding the search path to the DYLD_LIBRARY_PATH.

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -492,7 +492,7 @@ export class SwiftToolchain {
         if (!base) {
             return undefined;
         }
-        return path.join(base, "usr/lib");
+        return `${path.join(base, "usr/lib")}:${path.join(base, "usr/lib/swift/macosx/testing")}`;
     }
 
     /**
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -488,11 +488,12 @@ export class SwiftToolchain {
      * Library path for swift-testing executables
      */
     public swiftTestingLibraryPath(): string | undefined {
+        let result = "";
         const base = this.basePlatformDeveloperPath();
-        if (!base) {
-            return undefined;
+        if (base) {
+            result = `${path.join(base, "usr/lib")}:`;
         }
-        return `${path.join(base, "usr/lib")}:${path.join(base, "usr/lib/swift/macosx/testing")}`;
+        return `${result}${path.join(this.toolchainPath, "lib/swift/macosx/testing")}`;
     }
 
     /**


src/toolchain/toolchain.ts
```diff
@@ -492,7 +492,7 @@ export class SwiftToolchain {
         if (!base) {
             return undefined;
         }
-        return path.join(base, "usr/lib");
+        return `${path.join(base, "usr/lib")}:${path.join(base, "usr/lib/swift/macosx/testing")}`;
```
stmontgomery, line 495: `base` appears to be the _platform_ directory, but in this scenario the path should be relative to the toolchain

src/toolchain/toolchain.ts
```diff
@@ -488,11 +488,12 @@ export class SwiftToolchain {
...
     public swiftTestingLibraryPath(): string | undefined {
+        let result = "";
         const base = this.basePlatformDeveloperPath();
-        if (!base) {
-            return undefined;
+        if (base) {
+            result = `${path.join(base, "usr/lib")}:`;
         }
-        return path.join(base, "usr/lib");
+        return `${result}${path.join(this.toolchainPath, "lib/swift/macosx/testing")}`;
```
stmontgomery, line 496: I think it's just missing `usr/` prefix, now
```suggestion
        return `${result}${path.join(this.toolchainPath, "usr/lib/swift/macosx/testing")}`;
```

plemarquand: `this.toolchainPath` already has `/usr` prefixed in the path, the comment on it says "one folder up from the swift bin path": https://github.com/swiftlang/vscode-swift/pull/1521/files#diff-219ebe526be52811261023900cb320d9e51723308f63891304bfe53dcab033f1R107

[The commit message has a bit of history](https://github.com/swiftlang/vscode-swift/pull/606) there that predates my time, it may not be true anymore, however it would be a bigger change to update the toolchainPath everywhere.

rbenegal (approved):

src/toolchain/toolchain.ts
stmontgomery: Ah, ok if it already has that prefix then great! 👍

stmontgomery (approved):

Issue #1512, opened 2025-04-22
Add setting to turn inlay hints (inferred variable types) on or off


rbenegal:
**Is your feature request related to a problem? Please describe.**
Currently the default behaviour of the extension is to turn on inlay hints. This adds inferred types with grey text, and the user can double click on the text to insert it into the code. For example the inlay hint for the `Package` variable type:

<img width="349" alt="Image" src="https://github.com/user-attachments/assets/ff9bc34c-2b4d-4070-a849-2376b2c00645" />

While many users like this feature, some may not appreciate it as much and feel like it affects the readability of the code. Additionally, they may not know where this text is coming from if they are new and what they should do to turn it on/off.

**Describe the solution you'd like**
We should add a setting to turn inlay hints on and off. For now we can turn it off by default and when we add a walkthrough mode we should give the user the option to turn inlay hints on or off by default for swift projects. As well, we can update the documentation to tell users how to turn this feature on/off.

**Describe alternatives you've considered**
It is possible to turn off inlay hints on or off for swift with the following setting in the settings.json file:
```
    "[swift]": {
        "editor.inlayHints.enabled": "off"
    },
```

But users may not know that the inferred types are coming from "inlay hints".

**Additional context**
We can possibly add a setting that will modify the settings.json to set the right editor.inlayHints.enabled values.

adam-fowler:
There is already a setting, it just isn't in the Swift settings. Search the settings for inlay hints.

rbenegal:
Thank you! Yes, can definitely turn off inlay hints through the global settings as well (which would set `"editor.inlayHints.enabled": "off"` globally or for `settings.json` if done for the workspace I believe) but was thinking the Swift specific setting would be good when the extension is run for the first time and there is some sort of a walkthrough mode to help users configure preferences. New users may not know the setting name to search for and they could be new to VS Code in general as well.

Possibly inlay hints should have been called "variable type hints" or "variable type suggestions" or something to make it easier to find. Also instead of just "Double-click to insert" on hover perhaps it could say "Double-click to insert inlay hint" or show a link to the setting to hide it, but I'm not sure if that text can be overwritten.

jv-np:
I think the way F# shows type hints is one of the best ways of doing this, it puts them in virtual comments on the right side of the *code

![Image](https://github.com/user-attachments/assets/a49b9f83-3976-4e92-8531-77415fd088f1)

Is something like this particularly easy to implement as an alternative mode of exhibition?

[...]

PR #1526, opened 2025-04-28
Add an extension setting to control inlay hints

rbenegal:
Add an extension setting that controls `[swift]` language inlay hints so that hints can be disabled by default for now. Most other language extensions have inlay hints turned off by default (particularly for inferred variable types), likely because it is tricky to figure out how to turn them off/on otherwise.

Later on, when there is a walkthrough mode, we can provide users with an option to set inlay hints on or off when the extension runs for the first time.

Issue: #1512

### DIFF
--- a/src/commands/toggleInlayHints.ts
+++ b/src/commands/toggleInlayHints.ts
@@ -1,0 +1,35 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2025 the VS Code Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of VS Code Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import * as vscode from "vscode";
+import configuration from "../configuration";
+
+/**
+ * Configures editor.inlayHints.enabled settings based on swift.inlayHints.enabled settings
+ */
+export async function toggleInlayHints() {
+    let settingValue = undefined;
+
+    if (!configuration.inlayHintsEnabled) {
+        settingValue = "off";
+    }
+
+    const config = vscode.workspace.getConfiguration("", { languageId: "swift" });
+    await config.update(
+        "editor.inlayHints.enabled",
+        settingValue,
+        vscode.ConfigurationTarget.Workspace,
+        true
+    );
+}
--- a/package.json
+++ b/package.json
@@ -311,6 +311,11 @@
       {
         "title": "Swift",
         "properties": {
+          "swift.inlayHints.enabled": {
+            "type": "boolean",
+            "default": false,
+            "markdownDescription": "Display inlay hints. Inlay hints are variable annotations indicating their inferred type."
+          },
           "swift.path": {
             "type": "string",
             "default": "",
@@ -683,9 +688,9 @@
           },
           "sourcekit-lsp.inlayHints.enabled": {
             "type": "boolean",
-            "default": true,
+            "default": false,
             "markdownDescription": "Display Inlay Hints. Inlay Hints are variable annotations indicating their inferred type. They are only available if you are using Swift 5.6 or later.",
-            "markdownDeprecationMessage": "**Deprecated**: Please use `#editor.inlayHints.enabled#` instead."
+            "markdownDeprecationMessage": "**Deprecated**: Please use `#swift.inlayHints.enabled#` instead."
           },
           "sourcekit-lsp.support-c-cpp": {
             "type": "string",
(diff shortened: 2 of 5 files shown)


src/commands/toggleInlayHints.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+export async function toggleInlayHints() {
+    let settingValue = undefined;
+
+    if (!configuration.inlayHintsEnabled) {
+        settingValue = "off";
+    }
+
+    const config = vscode.workspace.getConfiguration("", { languageId: "swift" });
+    await config.update(
```
rbenegal, line 29: I think this mostly works ok, except in the special case where a user manually goes to their `settings.json` and sets:
```
    "[swift]": {
        "editor.inlayHints.enabled": "on"
    },
```
but doesn't set `swift.inlayHints.enabled` to `true`. This will cause the extension to just set `"editor.inlayHints.enabled": "off"` the next time the extension loads, which may be annoying if folks don't know about this and want inlayHints to be enabled all the time.

plemarquand, line 29: This is going to automatically add a .vscode/settings.json file for every package opened in VS Code and add the following to it:

```json
    "[swift]": {
        "editor.inlayHints.enabled": "off"
    }
```

When I update .vscode/settings.json to set this value to `"on"`, save, and reload the window it gets re-written to "off".

Automatically resetting the setting aside, I don't think we should be generating a settings.json for the user at all as these are meant to be for the user to control fully, not an extension.

plemarquand (changes requested):

src/commands/toggleInlayHints.ts
plemarquand: This is going to automatically add a .vscode/settings.json file for every package opened in VS Code and add the following to it:

```json
    "[swift]": {
        "editor.inlayHints.enabled": "off"
    }
```

When I update .vscode/settings.json to set this value to `"on"`, save, and reload the window it gets re-written to "off".

Automatically resetting the setting aside, I don't think we should be generating a settings.json for the user at all as these are meant to be for the user to control fully, not an extension.

rbenegal: Thanks, yes agreed that touching settings.json automatically is probably not welcome (although we do touch the settings.json for other things like swift environment variables, for example, but the user would have to set them manually I think). Any other suggestions on how the inlay hints could be disabled by default so that new users who don't know how to turn this setting off are not annoyed?

Could send a setting to sourcekit-lsp separately to turn off inlay hints possibly, and then if the user says: 
```
  "[swift]": {
        "editor.inlayHints.enabled": "on"
    }
```
or sets `swift.inlayHints.enabled` to `true` then it would turn on inlay hints?

rbenegal: Good ideas thanks, it seems like most do it with passing in flags to lsp (Go does it with gopls and rust with rust-analyser), although it looks like C/C++ use vscode api's to override inlayHint behaviour directly.

You are correct that it's not possible to tell whether `editor.inlayHints.enabled` is explicitly set to `on`. So it's looking like I'll just have to add functionality like the `sourcekit-lsp.inlayHints.enabled` setting had at one point.

MahdiBM: I just saw this PR (https://github.com/swiftlang/sourcekit-lsp/pull/2156), and i tracked back the discussion to here.

@plemarquand 
> How do other extensions handle this?

rust-analyzer (which is basically our sourcekit-lsp) has a bunch of settings for different kinds of inlay hints which are exposed through the VSCode extension as well. There are a lot of settings so I could have missed something but there doesn't appear to be a rust-specific settings to _completely_ turn off inlay hints.
Instead there are 27 different inlay hints settings that can granularly turn specific inlay hints on or off, and a lot of them are off by default.
I think this will be a decent amount of work, but should be what we aim for.

<kbd> Rust vscode extension inlay hints settings, scrshot 1 </kbd>
<kbd> Rust vscode extension inlay hints settings, scrshot 2 </kbd>
<kbd> Rust vscode extension inlay hints settings, scrshot 3 </kbd>

rbenegal: Thanks for this @MahdiBM. I think most other LSPs have several options for inlay hints, but as far as I understand sourcekit-lsp only provides them for inferred variable type names. So, I think for now it'll just be one inlay hint setting that will control this. For sure more options can be added once sourcekit-lsp provides inlay hints with the same level of granularity as other LSPs.

Also, I'm trying out some options on the LSP side to control this here - currently just the flag at startup, but need to look into using capabilities directly so the server doesn't need to be restarted every time (thanks to Mahdi linking this as well): https://github.com/swiftlang/sourcekit-lsp/pull/2156

rbenegal:
Closing this pull request, we'll just add documentation to the walkthrough to make people aware of the VS Code settings that can be used to turn off this feature.

PR #1534, opened 2025-05-02
Changes for release process

award999:
## Allow renaming the prerelease VSIX
- Since pre-release is always 1 minor version higher, we know it's the second binary

## Archive the commit SHA used to package the VSIX
- Eventually will be used downstream to tag releases
- Also setup PR job to package the VSIX and run tests against it so we can test against that in verifier jobs as well

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -31,6 +31,7 @@ jobs:
             echo "Created bundle $name"
             mv "$file" "$name"
           done
+          git config --global --add safe.directory $PWD
           git rev-parse HEAD > vscode-swift-sha.txt
       - name: Archive production artifacts
         id: archive
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -58,8 +58,6 @@ jobs:
         NVM_DIR=/usr/local/nvm
         CI=1
         FAST_TEST_RUN=${{ contains(github.event.pull_request.labels.*.name, 'full-test-run') && '0' || '1'}}
-        VSCODE_SWIFT_VSIX_ID=${{needs.package.outputs.artifact-id}}
-        GITHUB_REPOSITORY=${{github.repository}}
       linux_pre_build_command: . .github/workflows/scripts/setup-linux.sh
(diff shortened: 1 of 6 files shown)


scripts/download_vsix.ts
```diff
@@ -59,5 +59,9 @@ const repo = repository.split("/")[1];
     const newName = process.env["VSCODE_SWIFT_VSIX"] || "vscode-swift.vsix";
     await rename(files[0].path, newName);
     console.log(`Renamed artifact: ${files[0].path} => ${newName}`);
+    const preNewName =
+        process.env["VSCODE_SWIFT_PRERELEASE_VSIX"] || "vscode-swift-prerelease.vsix";
+    await rename(files[1].path, preNewName);
```
matthewbastien, line 64: I'm a little wary of just assuming that it'll be the second path. Why not extract the pre-release naming and version info into a helper method and use that in both the download and package scripts? That way it's more obvious where we're actually using the pre-release VSIX in the future if ever things need to change.

award999: I ended up using a regex approach in case there was a mismatch between the package.json numbers and the artifact version we pull down

.github/workflows/pull_request.yml
```diff
@@ -5,10 +5,51 @@ on:
...
+        uses: actions/checkout@v4
+      - name: Build Extension
+        run: |
+          export NODE_VERSION=v20.18.2
+          export NODE_PATH=/usr/local/nvm/versions/node/v20.18.2/bin
+          export NVM_DIR=/usr/local/nvm
+          . .github/workflows/scripts/setup-linux.sh
+          [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+          npm ci
+          npm run compile
```
matthewbastien, line 26: This compile step is not necessary: `npm run package` already does a compile.

matthewbastien (approved):
LGTM. Just one extremely minor comment.

.github/workflows/pull_request.yml
award999: Will fix

PR #1543, opened 2025-05-07
Don't colorize swift testing issue messages

plemarquand:
The inline issue label in VS Code doesn't support ANSI colour codes, and so renders them verbatim. Remove these so the messages are easier to read.


plemarquand:
Before: 
Screenshot 2025-05-07 at 1 30 52 PM

After: 
Screenshot 2025-05-07 at 1 30 20 PM

matthewbastien (approved):

Issue #1072, opened 2024-09-16
Support code-workspace files


award999:
Right now the vscode-swift extension is not working well with [multi-root workspaces](https://code.visualstudio.com/docs/editor/workspaces#_multiroot-workspaces). Using the asset/test projects as an example

user should be able to define multiple folders
```
"folders": [
  {
	  "name": "defaultPackage",
	  "path": "./defaultPackage"
  },
  {
	  "name": "diagnostics",
	  "path": "./diagnostics"
  },
],
```

and multiple task definitions
```
"tasks": [
	{
		"type": "swift",
		"args": [
			"build",
			"--build-tests",
			"--verbose"
		],
		"cwd": "./defaultPackage",
		"group": "build",
		"label": "swift: Build All (defaultPackage)",
		"detail": "swift build --build-tests --verbose"
	}
]
```

particularly "cwd" does not resolve properly whether it is relative ```./defaultPackage``` or ```${workspaceFolder:defaultPackage}```. SourceKit-LSP was not starting up correctly. Have not tried debugging yet but could see ```"program"``` not resolving properly as well.

rbenegal:
I can take a look at this.

rbenegal:
Verified with 7d54deff. Confirmed using the workspace from the test workspace file in `assets/test.code-workspace` that the directories, settings, tasks and launch configurations now work as expected (previously it would just error out).

<img width="622" alt="Image" src="https://github.com/user-attachments/assets/65aeaff1-6495-459e-a08a-c77fd4cb76f9" />

For example, running the `--show-bin-path` command now works:
```
*  Executing task: swift: Build All from code workspace 

> /Users/user/Library/Developer/Toolchains/swift-6.1-RELEASE.xctoolchain/usr/bin/swift build --show-bin-path

/Users/user/src/vscode-swift/assets/test/defaultPackage/.build/arm64-apple-macosx/debug
```

<img width="558" alt="Image" src="https://github.com/user-attachments/assets/e15fa7a5-1d5a-40e0-88f0-2c62ff6438b1" />

PR #1566, opened 2025-05-20
Support for multi-root .code-workspace workspaces

award999:
- Support workspace folder variable subsitution, ex. `${workspaceFolder:proj1}`
- When a code-workspace add the configured `name` instead of the relative path as a postfix to the task name
- Add same postfixes to plugin tasks
- For Run and Debug Build commands, all fetching launch configurations from outside the workspace
- Add a new test target to test a handful of currated suites to make sure that .code-workspace support does not break

Issue: #1072

### DIFF
--- a/src/commands/build.ts
+++ b/src/commands/build.ts
@@ -19,6 +19,7 @@ import { debugLaunchConfig, getLaunchConfiguration } from "../debugger/launch";
 import { executeTaskWithUI } from "./utilities";
 import { FolderContext } from "../FolderContext";
 import { Target } from "../SwiftPackage";
+import { packageName } from "../utilities/tasks";
 
 /**
  * Executes a {@link vscode.Task task} to run swift target.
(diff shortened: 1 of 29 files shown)


src/commands/build.ts
```diff
@@ -111,7 +112,11 @@ export async function debugBuildWithOptions(
     const launchConfig = getLaunchConfiguration(target.name, current);
     if (launchConfig) {
         ctx.buildStarted(target.name, launchConfig, options);
-        const result = await debugLaunchConfig(current.workspaceFolder, launchConfig, options);
+        const result = await debugLaunchConfig(
+            vscode.workspace.workspaceFile ? undefined : current.workspaceFolder,
```
plemarquand, line 116: Often I'll have a workspace but just with one folder in it. Does this check its truly multi-root? It could be a bit confusing because a workspace with one folder and a single workspace folder look the same (but maybe they are the same thing when it comes to this check?)

award999: Ya this is expected. The folder parameter docstring: `The workspace folder for looking up named configurations and resolving variables or undefined for a non-folder setup.`

Basically we don't want to resolve the launch configuration in the context of the singular workspace folder, we want to resolve it in the context of the full workspace

plemarquand (approved):

Issue #1363, opened 2025-02-07
enable "select target platform" for remote ssh sessions


RDMurray:
**Is your feature request related to a problem? Please describe.**

I want to be able to use VSCode on windows to develop for iOS via a remote ssh session. Currently select target platform doesn't work.

**Describe the solution you'd like**

The select target platform command to work for remote sessions when the remote is macOS.

**Describe alternatives you've considered**

I tried live share, but don't get any language services. It looks for a swift toolchain on the client instead.

**Additional context**

The reason I would like this feature is that I am blind and use a screen reader. While macOS has voiceover which does work with VSCode, it is much better on Windows with the NVDA screen reader.

matthewbastien:
@RDMurray can you confirm whether or not this is still happening for you? I was able to reproduce this issue a while ago, but with the latest version of the Swift extension (v2.2.0) it seems to be fixed.

RDMurray:
I didn't realise that it now requires Swift 6.1. After updating Xcode, the command is available when running locally. It still doesn't work from a remote ssh session on a Windows client.

The LSP does work correctly in a remote ssh session if I use the local VSCode to select the target platform and save it in the workspace settings first. It's just the command that doesn't appear in the command pallet.

matthewbastien:
Okay I can reproduce this again. I was SSHing from `macOS -> Windows` instead of the other way around. Sorry about that. Investigating further...

matthewbastien:
Fixed by #1567. Going to try to put out a pre-release tomorrow that contains the fix.

matthewbastien:
Pre-release version `2.3.20250521` contains the fix for this.

award999:
Verified with rdar://136229497&148576141&146712855&145335915&149873324&147774241&146317274&151867003&147774323&150786278&148576146

award999:
https://github.com/swiftlang/vscode-swift/actions/runs/15480110148/job/43584267614

PR #1567, opened 2025-05-20
Only use the context key for enabling the "swift.switchPlatform" command

matthewbastien:
It appears that the `isMac` context key provided by VS Code does not work the way we expect when running in a remote extension host. Update the logic for the `swift.switchPlatformAvailable` context key to check for the active platform and use that instead.

Issue: #1363

### DIFF
--- a/package.json
+++ b/package.json
@@ -908,7 +908,7 @@
         },
         {
           "command": "swift.switchPlatform",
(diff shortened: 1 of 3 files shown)


plemarquand (approved):

rbenegal (approved):
Great find thanks!

PR #1572, opened 2025-05-21
Add an extension walkthrough

plemarquand:
Adds a walkthrough to the package.json that guides users through:

- Installing Swift
- Creating a New Project
- Extension Commands
- Running Tests
- Selecting a Toolchain


package.json
```diff
@@ -33,6 +33,76 @@
...
+            "description": "🎨 Explore the commands available in the command palette prefixed with \"> Swift:\". You can use the command palette to run any of the commands provided by the Swift extension.\n[Open Command Palette](command:workbench.action.showCommands)",
+            "media": {
+              "svg": "assets/walkthrough/images/swiftCommands.svg",
+              "altText": "Swift Commands"
+            }
+          },
+          {
+            "id": "testing",
+            "title": "Running Tests",
+            "description": "✅ Tests are automatically discovered in your project and added to the [Testing View](https://code.visualstudio.com/docs/debugtest/testing#_automatic-test-discovery-in-testing-view).\n🧪 Both [XCTest](https://developer.apple.com/documentation/xctest) and [Swift Testing](https://developer.apple.com/xcode/swift-testing/) tests are supported.",
```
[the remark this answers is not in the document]
rbenegal: Ah right thanks, good point, sorry I guess I got confused between XCTests requiring Xcode on the Mac (I think if you just install the tool chain with Swiftly it may not work?) and it not working on other platforms.

assets/walkthrough/createNewProject.md
```diff
@@ -0,0 +1,12 @@
+There are several project templates to choose from:
```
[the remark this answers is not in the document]
plemarquand: I agree, but these descriptions aren't going to 1:1 match the ones from SwiftPM's help documentation so we'd need to craft a description for each one by hand anyway.

assets/walkthrough/swiftToolchains.md
```diff
@@ -0,0 +1,10 @@
+The Swift extension automatically detects your installed Swift toolchain. However, it also provides a command called `Swift: Select Toolchain...` which can be used to select between toolchains if you have multiple installed.
+
+You may be prompted to select where to configure this new path. Your options are to:
+
+- Save it in User Settings
```
award999, line 5: similar to what is done for command names, we should show button text as `Save it in User Settings` and `Save in Workspace Setting`

assets/walkthrough/swiftToolchains.md
```diff
@@ -0,0 +1,10 @@
...
+- Save it in Workspace Settings
+
+Keep in mind that Workspace Settings take precedence over User Settings.
```
award999, line 8: Maybe for the `User Settings` tab name too?

award999 (commented):
amazing start, here are some thoughts

package.json
```diff
@@ -33,6 +33,76 @@
...
+            "when": "isLinux"
+          },
+          {
+            "id": "installing-mac",
+            "title": "Installing Swift",
+            "description": "💻 Swift is cross-platform! See the instructions to [install on Windows](https://www.swift.org/install).\n🧑‍🤝‍🧑 Swift is [open source](https://github.com/swiftlang) and community driven!\n💬 Questions? Visit the [Swift forums](https://forums.swift.org/) for help.",
+            "media": {
+              "markdown": "./assets/walkthrough/welcome.md"
+            },
+            "when": "isWindows"
```
matthewbastien, line 67: I'm curious to know if this works when using Remote SSH to go from Mac -> Windows or vice versa. The isWindows/isMac/isLinux clauses only check the platform that the UI is running on. However, installing Swift depends on where the extension host is running.

Issue #1533, opened 2025-05-01
Swift: Generate Launch Configurations


MahdiBM:
In the Swift extension settings, there is this option: 'Swift: Auto Generate Launch Configurations'.
You already know what it does. What I'm asking is that there should be a "Command Palette" button to manually trigger this on-demand. It should probably be named "Swift: Generate Launch Configurations" (notice no 'Auto').

matthewbastien:
Fixed by #1577

award999:
Verified with 705a3da1fa794d3f8ded90ed25d28ed41553ceb7

PR #1577, opened 2025-05-22
Add command to generate launch configurations

matthewbastien:
Adds a new command called `Swift: Generate Launch Configurations` that can be used to manually trigger the generation of launch configurations for the workspace.

I've also simplified the logic for generating launch configurations by moving some of the generated keys into the debugAdapterFactory so they can be applied at runtime without needing to be set in the launch config. This will reduce the need for updating launch configurations and, by extension, reduce the number of notifications we push to the user.

Issue: #1533

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,6 +101,11 @@
     ],
     "commands": [
       {
+        "command": "swift.generateLaunchConfigurations",
+        "title": "Generate Launch Configurations",
+        "category": "Swift"
+      },
+      {
         "command": "swift.previewDocumentation",
         "title": "Preview Documentation",
         "category": "Swift",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -886,6 +891,10 @@
         }
       ],
       "commandPalette": [
+        {
+          "command": "swift.generateLaunchConfigurations",
+          "when": "swift.hasPackage"
```
plemarquand, line 896: I think this should be based off a new context key called `hasExecutableTarget`. If you create a basic library package (i.e. with `swift package init`) there are no launch configs to generate because there is nothing executable to launch.

plemarquand (changes requested):

plemarquand (approved):

Issue #1610, opened 2025-06-12
Test Failure: ` ProjectPanelProvider Test Suite` - setup hook failed on `Windows 6.0`


award999:
**Failure logs**

```
1) ProjectPanelProvider Test Suite
       "before all" hook for "Includes top level nodes":
     Error: Timeout of 300000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (d:\a\vscode-swift\vscode-swift\dist\test\integration-tests\ui\ProjectPanelProvider.test.js)
      at listOnTimeout (node:internal/timers:581:17)
      at processTimers (node:internal/timers:519:7)
```

Link to failed run: [GitHub Actions Link](https://github.com/swiftlang/vscode-swift/actions/runs/15449355842/job/43487369711?pr=1600)

**Additional context**
N/A

**Related test failures**
N/A

plemarquand:
Verified with [a5d7037](https://github.com/swiftlang/vscode-swift/commit/a5d703713108deb87f16410fe8f2dcb1d626113e)

PR #1623, opened 2025-06-16
Try to fix project panel tests

award999:
Issue: #1610


award999:
@plemarquand I moved the build step to after the tree provider getting initialized since you had logic in there to grab focus when a task runs

rbenegal (approved):
Thank you!

PR #1625, opened 2025-06-18
Avoid use of instanceof

award999:
instanceof causes issues in nightly because VSIX version and test version of the PackageNode class are different

Also allow passing folder context to package commands as are some random focus issues

### DIFF
--- a/src/ui/ProjectPanelProvider.ts
+++ b/src/ui/ProjectPanelProvider.ts
@@ -69,6 +69,23 @@ async function getChildren(directoryPath: string, parentId?: string): Promise<Fi
 export class PackageNode {
     private id: string;
 
+    /**
+     * "instanceof" has a bad effect in our nightly tests when the VSIX
+     * bundled source is used. For example:
+     *
+     * ```
+     * vscode.commands.registerCommand(Commands.UNEDIT_DEPENDENCY, async (item, folder) => {
+     *  if (item instanceof PackageNode) {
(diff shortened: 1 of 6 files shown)


src/ui/ProjectPanelProvider.ts
```diff
@@ -69,6 +69,23 @@ async function getChildren(directoryPath: string, parentId?: string): Promise<Fi
...
+     *  if (item instanceof PackageNode) {
+     *      return await uneditDependency(item.name, ctx, folder);
+     *  }
+     * }),
+     * ```
+     *
+     * So instead we'll check for this set boolean property. Even if the implementation of the
+     * {@link PackageNode} class changes, this property should not need to change
+     */
+    __isDependency = true;
```
matthewbastien: Super nitpicky, but I feel like this property should just match the name of the class. E.g. `__isPackageNode` or `__isProjectPanelPackageNode`.

matthewbastien (approved):
LGTM. Just one incredibly minor nitpick about naming.

rbenegal (approved):

src/ui/ProjectPanelProvider.ts
award999: @matthewbastien changed, but also just made it a static check to make this a little easier to use

matthewbastien: Awesome! That's even better.

PR #1626, opened 2025-06-18
Hide files excluded with `files.exclude` from Project Panel

plemarquand:
The project panel should respect the `files.exclude` VS Code setting. This setting is combined with the existing
`swift.excludePathsFromPackageDependencies`.

The tests required an update to `updateSettings`, as the existing global `files.exclude` settings are combined with the ones set in the test. When setting the settings, verify that all expected settings are present before continuing, but don't require the settings object to be an exact match of what we set.

### DIFF
--- a/test/integration-tests/ui/ProjectPanelProvider.test.ts
+++ b/test/integration-tests/ui/ProjectPanelProvider.test.ts
@@ -356,6 +356,38 @@ suite("ProjectPanelProvider Test Suite", function () {
             const errorNode = children.find(n => n.name === "Error Parsing Package.swift");
             expect(errorNode).to.not.be.undefined;
         });
+
+        suite("", () => {
+            let resetSettings: (() => Promise<void>) | undefined;
+            beforeEach(async function () {
+                resetSettings = await updateSettings({
+                    "files.exclude": { "**/*.swift": true },
+                    "swift.excludePathsFromPackageDependencies": ["**/*.md"],
+                });
+            });
+
+            test("Excludes files based on settings", async () => {
+                contextKeys.flatDependenciesList = false;
+                const children = await getHeaderChildren("Dependencies");
+                const dep = children.find(n => n.name === "swift-markdown") as PackageNode;
+                expect(dep, `${JSON.stringify(children, null, 2)}`).to.not.be.undefined;
+
+                const folders = await treeProvider.getChildren(dep);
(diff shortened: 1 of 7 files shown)


test/integration-tests/ui/ProjectPanelProvider.test.ts
```diff
@@ -356,6 +356,38 @@ suite("ProjectPanelProvider Test Suite", function () {
             const errorNode = children.find(n => n.name === "Error Parsing Package.swift");
             expect(errorNode).to.not.be.undefined;
         });
+
+        suite("", () => {
```
award999: name

plemarquand: done

test/integration-tests/utilities/testutilities.ts
```diff
@@ -453,3 +453,74 @@ function decomposeSettingName(setting: string): { section: string; name: string
...
+    if (
+        typeof configValue === "object" &&
+        typeof expected === "object" &&
+        configValue !== null &&
+        expected !== null &&
+        !Array.isArray(configValue) &&
+        !Array.isArray(expected)
+    ) {
+        // Ensure we're working with plain objects
+        const configObj = configValue as Record<string, unknown>;
```
matthewbastien, line 509: This won't actually perform a runtime type check if that's what you were going for. Using `as` only tells the TypeScript compiler to treat the value as the provided type. This will most likely just be optimized out of the final JavaScript code. You'd need to use a `typeof` check (and even then it's a little finicky for objects).

But really, we could probably just use `Record<string, unknown>` as the type for the arguments instead of just `unknown`.

matthewbastien: Or I suppose it would also need to allow an array of records.

plemarquand: This is mainly to satisfy the TS compiler

matthewbastien: Makes sense. The comment just made it seem like it was intended to be a check. So, I wanted to make sure. ;)

rbenegal (approved):

award999:
@plemarquand can we save owning some of this logic by using `vscode.workspace.findFiles` API?

plemarquand:
@award999 `vscode.workspace.findFiles` was only finding files, not directories, so had to go back to `glob`

award999:
Verified with 7d54def

PR #1629, opened 2025-06-19
Properly scope all settings

plemarquand:
Be more explicit about settings scopes, marking appropriate ones as `application` (user settings applied across all swift projects) and `machine-overridable` (for per-project settings).


tracymiranda (approved):

package.json
```diff
@@ -782,7 +803,8 @@
               "Never automatically update CodeLLDB settings when they are incorrect."
             ],
             "markdownDescription": "Choose how CodeLLDB settings are updated when debugging Swift executables.",
-            "order": 3
+            "order": 3,
+            "scope": "machine-overridable"
```
award999, line 808: maybe application scope?

package.json
```diff
@@ -782,7 +803,8 @@
...
           },
           "swift.debugger.useDebugAdapterFromToolchain": {
```
award999, line 810: maybe application?

award999:
@plemarquand is their any repercussion if one of the settings we're changing to `application` is set for the workspace already?

package.json
plemarquand: I think the CodeLLDB settings should be per project because its possible some projects may work better with one debugger over the other.

plemarquand:
@award999 Good point, there is a risk that some of the ones being set to "application" will no longer take effect if they're set in a settings.json for a project. 

A list of the ones that are now application are:
- `swift.actionAfterBuildError`
- `swift.enableTerminalEnvironment`
- `swift.showBuildStatus`
- `swift.showCreateSwiftProjectInWelcomePage`
- `swift.openAfterCreateNewProject`

None of these would break a project if they were ignored on a per-project basis, but it might be a bit confusing if they stopped working as they did before.

award999 (approved):

award999:
@plemarquand I think those are low risk settings. Will point it out in release notes and message but I'm not concerned

matthewbastien:
Verified with 323d996

Issue #1619, opened 2025-06-15
`runInTerminal` not working when using lldb-dap from swiftly


svanimpe:
**Context**
I teach programming to novices with no prior programming experience. My programming examples rely heavily on `print` and `readLine` as that is the only way novices can do input/output with code they can understand.

**Problem**
When debugging using the CodeLLDB extension, input and output goes through the integrated terminal, so `print` and `readLine` work fine. However, the new debug adapter uses the debug console, where I can see the output of `print`, but cannot provide any input to `readLine`.
I was told to use `runInTerminal` in **launch.json**, but unfortunately that's not working for me.

My launch configuration is as follows:

```json
{
    "configurations": [
        {
            "type": "swift",
            "request": "launch",
            "args": [],
            "cwd": "${workspaceFolder:Greeter}",
            "name": "Debug Greeter",
            "program": "${workspaceFolder:Greeter}/.build/debug/Greeter",
            "preLaunchTask": "swift: Build Debug Greeter",
            "runInTerminal": true
        },
    ]
}
```

Launching this outputs the following error in the integrated terminal:

```
❯  /usr/bin/env /Users/steven/Developer/Greeter/lldb-dap --comm-file /var/folders/j2/lvq8vj0s6r5bzykn9_k7zp
q00000gn/T/lldb-dap-run-in-terminal-comm-4796cb --debugger-pid 1980 --launch-target /Users/steven/Developer
/Greeter/.build/debug/Greeter 
env: /Users/steven/Developer/Greeter/lldb-dap: No such file or directory
```

I get the same error on Ubuntu. On Windows, an error window pops up with the message "Unimplemented".

**Environment**
 - OS: macOS 15.5 and Ubuntu 24.04.2 and Windows 11 24H2
 - Swift verson: 6.1.2
 - Visual Studio Code version: 1.100.3
 - vscode-swift version: 2.4.0
 - lldb-dap version: 0.2.15

award999:
thanks for the report @svanimpe 👍 
The stdout and stdin issues on Windows are a little separate and tracked by #309 which I have been looking at

Now the error message including `/Users/steven/Developer/Greeter/lldb-dap: No such file or directory` explains why it doesn't launch on macOS.

To figure why it's not found, can you confirm some things?
- Is this an Xcode or swiftly toolchain?
- In a terminal outside vscode does `xcrun -find lldb-dap` return a path?
- Is your `swift.path` setting explicitly set in VSCode settings?

[...]

PR #1632, opened 2025-06-20
Resolve the swiftly bin path

award999:
Was failing to compute the correct lldb-dap path because the user had pointed `swift.path` to `~/.swiftly/bin`. When resolving the path, lldb-dap path ends up being computed as `$PWD/lldb-dap`. To get around this, if we know the provided path is a swiftly bin directory, then resolve the correct swift path and toolchain path using built-in swiftly commands

Issue: #1619

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -692,19 +692,9 @@ export class SwiftToolchain {
     }
 
     private static async swiftlyInUseLocation(swiftlyPath: string, cwd?: vscode.Uri) {
-        const env: Record<string, string> = {};
-        if (path.isAbsolute(swiftlyPath)) {
-            env["SWIFTLY_HOME_DIR"] = path.dirname(path.dirname(swiftlyPath));
-            env["SWIFTLY_BIN_DIR"] = path.dirname(swiftlyPath);
-        }
-        const { stdout: inUse } = await execFile(
-            swiftlyPath,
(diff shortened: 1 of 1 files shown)


src/toolchain/toolchain.ts
```diff
@@ -605,7 +605,15 @@ export class SwiftToolchain {
                 }
             }
             // swift may be a symbolic link
-            const realSwift = await fs.realpath(swift);
+            let realSwift = await fs.realpath(swift);
+            if (path.basename(realSwift) === "swiftly") {
+                try {
+                    const { stdout } = await execFile(realSwift, ["run", "which", "swift"]);
```
cmcgee1024: suggestion: If you know that this is swiftly based on the symlink, then you can use `swiftly use --print-location` and append `usr/bin/swift` to the end of that path. This will work on systems that don't have the `which` command, such as Windows, or a minimal Linux system.

award999: @cmcgee1024 done. For the new solution, is setting those env variables a bad idea?

cmcgee1024: Yes, I think that you should try to avoid setting the environment variables. You can't easily deduce where the user might have put the swiftly home directory from the binary location. This can set it to be anywhere. If they installed swiftly in the default location, then swiftly itself should be able to figure that out.

plemarquand (approved):

Issue #1472, opened 2025-03-31
No longer notified to reload after `xcode-select`


award999:
**Describe the bug**
We used to notify the user to reload the extension after detecting and `xcode-select`

**To Reproduce**
1. Run the `Swift: Select Toolchain...` command and select a version of Xcode
2. When prompted, reload the extension
3. Outside VSCode, `xcode-select -s` a different Xcode
4. Nothing happens in VSCode

**Expected behavior**

**Environment**
 - OS: macOS
 - Swift version: 6.1
 - Visual Studio Code version: 1.98.2
 - vscode-swift version: main

**Additional context**
N/A

plemarquand:
I think this is intended because you explicitly set a toolchain path in your settings using the first command. If a toolchain is set in settings the prompt won't appear because the user has explicitly overridden the toolchain and the system Xcode has no effect. See: https://github.com/swiftlang/vscode-swift/blob/main/src/toolchain/SelectedXcodeWatcher.ts#L75

plemarquand:
I'm notified to reload with [a5d7037](https://github.com/swiftlang/vscode-swift/commit/a5d703713108deb87f16410fe8f2dcb1d626113e)

PR #1640, opened 2025-06-23
Smarter decisions on if user should be prompted for Xcode change

award999:
If you use a swiftly path then you are prompted everytime VSCode opens
even if the xcode-select -p is the same. Example settings:
```
"swift.path": "${userHome}/.swiftly/bin",
"swift.swiftEnvironmentVariables": {
    "DEVELOPER_DIR": "/Applications/Xcode.app/Contents/Developer"
},
```

So consider the DEVELOPER_DIR as well to know if this changed.

Issue: #1472

### DIFF
--- a/src/toolchain/SelectedXcodeWatcher.ts
+++ b/src/toolchain/SelectedXcodeWatcher.ts
@@ -67,10 +67,14 @@ export class SelectedXcodeWatcher implements vscode.Disposable {
      */
     private async setup() {
         this.xcodePath = await this.xcodeSymlink();
+        const env = configuration.swiftEnvironmentVariables;
+        const developerDir = env["DEVELOPER_DIR"];
+        const matchesPath = (xcodePath: string) => configuration.path && configuration.path.startsWith(xcodePath);
+        const matchesDeveloperDir = (xcodePath: string) => developerDir && developerDir.startsWith(xcodePath);
         if (
             this.xcodePath &&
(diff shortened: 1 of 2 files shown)


plemarquand (approved):

src/toolchain/SelectedXcodeWatcher.ts
```diff
@@ -89,7 +93,18 @@ export class SelectedXcodeWatcher implements vscode.Disposable {
                     await showReloadExtensionNotification(
                         "The Swift Extension has detected a change in the selected Xcode. Please reload the extension to apply the changes."
                     );
-                } else {
+                } else if (developerDir() && !matchesDeveloperDir(this.xcodePath)) {
+                    const selected = await vscode.window.showWarningMessage(
+                        'The Swift Extension has detected a change in the selected Xcode which does not match the value of your DEVELOPER_DIR in the "swift.swiftEnvironmentVariables" setting. Would you like to update your configured "swift.swiftEnvironmentVariables" setting?',
+                        "Remove From Settings",
```
rbenegal, line 99: Hmm, thinking a bit more about `Remove From Settings` and I'm not sure if I could figure out what this actually does if I just saw the pop-up directly. I think this is fine for now, but maybe in the future it could something like `Clear swiftEnvironmentVariables settings` (or maybe "reset")?

rbenegal (approved):

Issue #1648, opened 2025-06-25
Inserting newline in doc comment with text after cursor inserts `//` instead of `///` on next line


award999:
**Describe the bug**
When write a func doc comment and splitting a line, the new line is prepended with // instead of ///

**To Reproduce**
See: https://github.com/user-attachments/assets/f783dbaf-992f-41f4-bfbe-f62cb1b675d9

**Expected behavior**
New line starts with "///" not "//"

**Environment**
 - OS: [e.g. Ubuntu 20.04]
 - Swift version: 6.1.2
 - Visual Studio Code version: 1.100
 - vscode-swift version: 2.4

**Additional context**
N/A

rbenegal:
Verified with 3a615aad. Confirmed that breaking up a `///` comment now results in another `///` comment, and that adding a return at the end of a `///` comment results in a suggestion to begin a new `///` comment.

PR #1651, opened 2025-06-25
Make sure newline starts with /// when splitting doc comment

award999:
The existing behaviour where splitting a /// line results in the next line starting with // is a vscode issue. Other examples follow a similar approach as this PR to accomplish the desired behaviour.

Issue: #1648

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - Cleanup Swift diagnostics when the source file is moved or deleted ([#1653](https://github.com/swiftlang/vscode-swift/pull/1653))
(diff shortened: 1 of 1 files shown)


rbenegal (approved):
Good catch! Thanks for the fix!

src/editor/CommentCompletion.ts
```diff
@@ -15,6 +15,14 @@
 import * as vscode from "vscode";
 import { DocumentParser } from "./DocumentParser";
 
+function isLineComment(document: vscode.TextDocument, line: number): boolean {
+    // test if line consists of just '///'
+    if (/^\s*\/\/\//.test(document.lineAt(line).text)) {
```
plemarquand: nit: just `return /^\s*\/\/\//.test(document.lineAt(line).text)`

award999: ya good call, I just cut and past :p

Issue #1652, opened 2025-06-25
Diagnostics are not cleaned up after a file is deleted or moved


award999:
**Describe the bug**
Diagnostics linger when they should be cleaned up when files 

**To Reproduce**
1. Write an error to a file
2. Delete or rename the file
3. Notice the diagnostic still exists in the collection
4. Clicking the link brings up an error editor

**Expected behavior**
Diagostics are cleaned up. SourceKit-LSP may send a new diagnostic for the new file if it was moved.

**Environment**
All

**Additional context**
N/A

rbenegal:
Verified with 3a615aad. The issues no longer show up in the "Problems" panel if the file is deleted, but updated issues do show up if the file is renamed.

PR #1653, opened 2025-06-25
Cleanup diagnostics for deleted or renamed files

award999:
When the file is deleted or renamed, remove it from the diagnostic collection. Extract the swift file types to a constant.

Issue: #1652


plemarquand (approved):

plemarquand:
Would be nice to add a test for this

award999:
> Would be nice to add a test for this

I figured this added a bit of a flake factor as we would have to delete or rename the source file and hope we can undo the action later in the suite. I guess I could add a `removeFile` method that calls the update directly to get around the file watcher, just not sure if there is much value then in that

PR #1663, opened 2025-06-27
Fix test explorer tests not updating on document modification

plemarquand:
The LSP client was being created too early, before the imperative set of the `documentSymbolWatcher` on the client. Consequently we weren't making the `textDocument/tests` request when symbols updated in a test file.


award999:
Verified with c4e3c03

PR #1664, opened 2025-06-28
Fix improper parenting of tests w/ identical names in explorer

plemarquand:
Two tests in swift-testing can share the same identifier, for instance if you have two identically named tests with the same name but that take arguments of different types, i.e:

```swift
@Test(arguments: [1]) func a(_ x: Int) {}
@Test(arguments: ["a"]) func a(_ x: String) {}
```

In this case sourcekit-lsp provides a disambiguated id with the filename/line number at the end, i.e. `TestTarget/a:File.swift:2:2`.

The code to synthesize parents in the test explorer when provided with test items that had none didn't take in to account this form of disambiguated id, which lead to these identically named tests being incorrectly synthesized and parented to the wrong test.

### DIFF
--- a/src/TestExplorer/TestDiscovery.ts
+++ b/src/TestExplorer/TestDiscovery.ts
@@ -73,6 +73,12 @@ export async function updateTestsFromClasses(
     updateTests(testController, results);
 }
 
+function isFileDisambiguated(id: string): boolean {
+    // a regex to check if the id ends with a string like "filename.swift:line:column"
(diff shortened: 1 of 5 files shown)


rbenegal (approved):

award999:
Changelog entry 🙏

award999:
Verified with c3abc77

award999:
Verified with c3abc77

PR #1668, opened 2025-06-30
Ensure document symbols are provided for folders in multi root workspaces

plemarquand:
If the user is working in a multi root workspace document symbols would not be provided if they were for any folder except the first one added. This also prevented the test explorer from properly initializing.

Check to see if the folder being added is in the list of VS Code workspace folders and associate the folder with the onDocumentSymbols request from the LSP. 

Issue: #1669

### DIFF
--- a/src/sourcekit-lsp/LanguageClientManager.ts
+++ b/src/sourcekit-lsp/LanguageClientManager.ts
@@ -97,7 +97,8 @@ export class LanguageClientManager implements vscode.Disposable {
     private singleServerSupport: boolean;
     // used by single server support to keep a record of the project folders
     // that are not at the root of their workspace
-    public subFolderWorkspaces: vscode.Uri[];
+    public subFolderWorkspaces: FolderContext[] = [];
+    private addedFolders: FolderContext[] = [];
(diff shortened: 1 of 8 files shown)


src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -425,7 +429,7 @@ export class LanguageClientManager implements vscode.Disposable {
         const serverOptions: ServerOptions = sourcekit;
         let workspaceFolder = undefined;
         if (folder && !this.singleServerSupport) {
-            workspaceFolder = {
+            workspaceFolder = workspaceFolder = {
```
rbenegal: Seems like workspaceFolder is here twice, is that intentional?

plemarquand: copy paste error, thanks!

award999:
Is there a reason the code-workspace test run missed this, i.e. need for new/modified tests?

plemarquand:
There were no tests covering modification (textDocument/tests LSP req), only finding tests on workspace open (workspace/tests). I've added some new tests in this patch.

award999 (commented):
Missing CHNAGELOG entry

Issue #636, opened 2023-11-10
Allow exclusion of sourcekit-lsp workspaces.


GeorgeLyon:
**Is your feature request related to a problem? Please describe.**
A VSCode workspace may contain multiple sourcekit-lsp workspaces. To be specific, by "VSCode workspace" I mean the folder you are opening in a VSCode window and by "sourcekit-lsp workspace" I mean the workspaces managed by sourcekit-lsp (which effectively tells the LSP server where to search for Package.swift or compile_commands.json). As I understand it, this VSCode extension automatically tries to find a workspace when a Swift file is opened in the VSCode UI, and there is no way to influence this behavior. The one customization option provided is "search subfolders for packages", which tries to identify workspaces when the extension starts up.
This is fine for most projects, but there are times when this is problematic. One example is a project with a top-level Package.swift file (the "user-facing package") may have a number of other packages in its directory tree (packages testing the user-facing package, local packages that can be referenced by the user-facing package, or just one-off examples). I've noticed a number of weird issues in such a project, including some failures with package resolution (which can be disabled via an option) and LSP functionality being unavailable for the nested packages.

**Describe the solution you'd like**
For my example above, just deleting the top-level Package.swift fixed all my issues (i.e. LSP began working well in the nested packages). It would be great if there was a setting like the following:
```
"swift.sourckit-lsp.workspaces.exclude": [
  ".": true,
  "my/other/problematic/package": true
]
```

tamird:
Possibly related to this feature request: opening a workspace that contains non-swift `compile_commands.json` breaks in the presence of this extension. I don't have a small example, but if you [download Fuchsia](https://fuchsia.dev/fuchsia-src/get-started/learn/build/source-tree) and run `fx set` you'll have generated a `compile_commands.json`; if you open that workspace in the presence of this extension you'll see that just about all C++ tooling is broken.

rbenegal:
I can take a look a this.

rbenegal:
Not currently looking into this, but when someone does look into this: Another use case mentioned in #1545 is that folks may want to use their `compile_commands.json` file instead of `Package.swift`. Currently it's possible to use sourceKit settings to do this, but would be nice if we are smarter/provide more options for `compile_commands.json` files, non-swift `compile_commands.json` files and `Package.swift` files.

award999:
[...]

PR #1693, opened 2025-07-04
Introduce new `swift.excludePathsFromActivation` setting

award999:
## Description
This will make sure the specified folders are not initialized and added to the `WorkspaceContext`. This means no tasks, tests, project panel, etc for these excluded folders, and they are not added to the sourcekit-lsp instance. SourceKit-LSP will still initialize if the an excluded file is open in the editor, an issue against the LS will be raised separately.

Issue: #636

## Tasks

### DIFF
--- a/package.json
+++ b/package.json
@@ -797,7 +797,7 @@
             "markdownDescription": "Disable the running of SourceKit-LSP.",
             "markdownDeprecationMessage": "**Deprecated**: Please use `#swift.sourcekit-lsp.disable#` instead."
           },
-          "swift.exclude": {
+          "swift.excludePathsFromActivation": {
             "type": "object",
             "additionalProperties": {
               "type": "boolean"
--- a/package.json
+++ b/package.json
@@ -796,6 +796,13 @@
             "default": false,
             "markdownDescription": "Disable the running of SourceKit-LSP.",
             "markdownDeprecationMessage": "**Deprecated**: Please use `#swift.sourcekit-lsp.disable#` instead."
+          },
+          "swift.exclude": {
+            "type": "object",
+            "additionalProperties": {
+              "type": "boolean"
+            },
+            "markdownDescription": "Configure glob patterns for excluding Swift package folders from getting activated. This will take precedence over the glob patterns provided to `files.exclude`."
           }
         }
       },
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -740,6 +740,13 @@
             "default": false,
             "markdownDescription": "Disable the running of SourceKit-LSP.",
             "markdownDeprecationMessage": "**Deprecated**: Please use `#swift.sourcekit-lsp.disable#` instead."
+          },
+          "swift.exclude": {
```
matthewbastien: I'm leaning towards calling this `swift.packages.exclude` or `swift.excludePackages` given that it applies specifically to folders/packages. Gives a better indication of what's being excluded. We may also want to add exclusion for other things in the future and this will help differentiate.

matthewbastien (changes requested):
Looks great! Just one issue with naming.

package.json
award999: it is not just package though, it could be subfolders and files too, enabling and disabling different parts of the same tree

award999: I did notice a couple other exclusion settings I missed before, so going to use `excludePathsFromActivation` to give more granularity. There is the aspect of auto resolving packages but also individual files / sub-folders, etc can trigger activation when the files come into focus

test/integration-tests/utilities/workspace.test.ts
```diff
@@ -0,0 +1,32 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2024 the VS Code Swift project authors
```
rbenegal, line 5:
```suggestion
// Copyright (c) 2025 the VS Code Swift project authors
```

userdocs/userdocs.docc/Articles/Reference/settings.md
```diff
@@ -8,6 +8,64 @@ The Swift extension comes with a number of settings you can use to control how i
...
+    /proj1
+        /Package.swift
+    /proj2
+        /Package.swift
+    /aSubfolder
+        /proj3
+            /Package.swift
+```
+
+You can enable the `searchSubfoldersForPackages` setting so the Swift extension can initializing all these projects.
```
rbenegal, line 27:
```suggestion
You can enable the `searchSubfoldersForPackages` setting so the Swift extension can initialize all the projects.
```

rbenegal (approved):

plemarquand (approved):

PR #1696, opened 2025-07-04
Respect `files.exclude` values set to `false`

plemarquand:
The project panel would ignore all files of a given type regardless of if they their value was `true` or `false` in the setting. Only values of `true` should result in files being excluded.


award999 (approved):

plemarquand:
Verified with 323d996c4f32350896b7f0126b72de974100a8a1

Issue #1695, opened 2025-07-04
Better test buttons


MahdiBM:
**Is your feature request related to a problem? Please describe.**
I find it somewhat annoying that everytime i want to debug a test, i need to double click then choose the "debug" button.

**Describe the solution you'd like**

Something like rust has:
<kbd> <img width="561" height="89" alt="Image" src="https://github.com/user-attachments/assets/b8b3aa2b-3c8e-488e-b788-c07f8b6f71aa" /> </kbd>

**Describe alternatives you've considered**
Accepting the current implementation and moving on.

**Additional context**

plemarquand:
Does the rust extension also have the test run button (and test result icon after a run) in the gutter?

MahdiBM:
You mean behind the line numbers? You need to enable it (it == the whole test explorer) via a setting, then the answer would be a yes.

<kbd> <img width="653" height="97" alt="Image" src="https://github.com/user-attachments/assets/02320b79-6ba1-424f-b813-a72e40899055" /> </kbd>

plemarquand:
I think this would be nice to have, but available through a setting and turned off by default.

MahdiBM:
Should it be turned off by default though? I feel like it'll help those that come from Xcode a lot.
It took me a bit to realize how i need to debug a test with VSCode. Those lenses will certainly help.

MahdiBM:
(Considering Xcode does debug-test by default)

award999:
Verified with 58ad058e6d135dfb17514c475b4068721364944d

PR #1698, opened 2025-07-07
Code Lenses for Tests

plemarquand:
When the `swift.showTestCodeLenses` setting is `true`, show code lenses inline in the editor above suites and tests. This setting defaults to `false` and must be enabled first.

The three lenses are Run, Debug, and Run w/ Coverage.

Issue: #1695

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - Added code lenses to run suites/tests, configurable with the `swift.showTestCodeLenses` setting ([#1698](https://github.com/swiftlang/vscode-swift/pull/1698))
 
-## 2.6.2 - 2025-07-02
+## 2.8.0 - 2025-07-14
 
 ### Added
 


MahdiBM:
It's fine with me either way but just asking that are we sure we want the "Run w/ coverage" button?
I personally have never used it in any project, regardless if I have code coverage enabled in CI or not (I usually do).
If it's a low-usage button we can just not add it. It's still accessible behind the test indicator button if anyone needs it once in a while.

Us having some telemetry about how much users use these buttons would have been nice, but that's a big can of worms ...

plemarquand:
@MahdiBM I do use the "run with coverage" option sometimes on a single test as I want to make sure my test is exercising what I think it is. Maybe the setting shouldn't be a binary on/off but a tri-state of on/off/specific buttons I want to see

MahdiBM:
Hmm I think on/off is fine, I dont think having an extra button will be annoying. Though having the ability to choose specific buttons will be an improvement even if little, so that's also fine.

plemarquand:
@MahdiBM After some discussion I've turned these on by default. If you want fine grained control over which lenses appear you can set some permutation of `"swift.showTestCodeLenses": ["run", "debug", "coverage"]` in your application level VS Code settings

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 ### Added
 
+- Added code lenses to run suites/tests, configurable with the `swift.showTestCodeLenses` setting ([#1698](https://github.com/swiftlang/vscode-swift/pull/1698))
+
+## 2.6.2 - 2025-07-02
```
award999: 2.8.0 in the end right?

plemarquand:
```suggestion
## 2.8.0 - 2025-07-14
```

award999 (approved):

PR #1701, opened 2025-07-07
Cleanup: Remove log on missing .swift-version file

plemarquand:
In most cases this file won't exist, and that is OK. Omit the log pollution that prints the error when the .swift-version file can't be found.

### DIFF
--- a/src/PackageWatcher.ts
+++ b/src/PackageWatcher.ts
@@ -154,9 +154,6 @@ export class PackageWatcher {
             const contents = await fs.readFile(versionFile);
             return Version.fromString(contents.toString().trim());
         } catch (error) {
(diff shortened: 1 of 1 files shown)


src/PackageWatcher.ts
```diff
@@ -154,9 +154,6 @@ export class PackageWatcher {
             const contents = await fs.readFile(versionFile);
             return Version.fromString(contents.toString().trim());
         } catch (error) {
-            this.workspaceContext.outputChannel.appendLine(
```
matthewbastien, line 157: I think it would make more sense to check the `code` property of the error. If it's not ENOENT then we probably do want to log it.

plemarquand: Good call, fixed

PR #1717, opened 2025-07-14
Add Swiftly toolchain management

roulpriya:
## Description
- Refactored Swiftly code from toolchain.ts to swiftly.ts
- Added list-available with json output

Issue: To be created

## Tasks

### DIFF
--- a/src/toolchain/swiftly.ts
+++ b/src/toolchain/swiftly.ts
@@ -1,3 +1,17 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2025 the VS Code Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of VS Code Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
 import * as path from "node:path";
 import { SwiftlyConfig } from "./ToolchainVersion";
 import * as fs from "node:fs/promises";
@@ -71,7 +85,7 @@ export class Swiftly {
             const response = ListAvailableResult.parse(JSON.parse(stdout));
             return response.toolchains.map(t => t.name);
         } catch (error) {
-            throw new Error("Failed to retrieve Swiftly installations from disk.");
+            throw new Error("Failed to retrieve Swiftly installations from disk: ${error.message}");
         }
     }
 
@@ -93,7 +107,7 @@ export class Swiftly {
                 .filter((toolchain): toolchain is string => typeof toolchain === "string")
                 .map(toolchain => path.join(swiftlyHomeDir, "toolchains", toolchain));
         } catch (error) {
-            throw new Error("Failed to retrieve Swiftly installations from disk.");
+            throw new Error("Failed to retrieve Swiftly installations from disk: ${error.message}");
         }
     }
 
@@ -129,7 +143,7 @@ export class Swiftly {
                 } catch (err: unknown) {
                     const error = err as ExecFileError;
                     // Its possible the toolchain in .swift-version is misconfigured or doesn't exist.
-                    void vscode.window.showErrorMessage(`${error.stderr}`);
+                    void vscode.window.showErrorMessage(`Failed to load toolchain from Swiftly: ${error.stderr}`);
                 }
             }
         }
--- a/src/toolchain/swiftly.ts
+++ b/src/toolchain/swiftly.ts
@@ -1,0 +1,157 @@
+import * as path from "node:path";
+import { SwiftlyConfig } from "./ToolchainVersion";
+import * as fs from "node:fs/promises";
+import { execFile, ExecFileError } from "../utilities/utilities";
+import * as vscode from "vscode";
+import { Version } from "../utilities/version";
+import { z } from "zod";
+
+const ListAvailableResult = z.object({
+    toolchains: z.array(
+        z.object({
+            inUse: z.boolean(),
+            installed: z.boolean(),
+            isDefault: z.boolean(),
+            name: z.string(),
+            version: z.discriminatedUnion("type", [
+                z.object({
+                    major: z.number(),
+                    minor: z.number(),
+                    patch: z.number().optional(),
+                    type: z.literal("stable"),
+                }),
+                z.object({
+                    major: z.number(),
+                    minor: z.number(),
+                    branch: z.string(),
+                    date: z.string(),
+
+                    type: z.literal("snapshot"),
+                }),
+            ]),
+        })
+    ),
+});
+
+export class Swiftly {
+    /**
+     * Finds the version of Swiftly installed on the system.
+     *
+     * @returns the version of Swiftly as a `Version` object, or `undefined`
+     * if Swiftly is not installed or not supported.
+     */
+    public async getSwiftlyVersion(): Promise<Version | undefined> {
+        if (!this.isSupported()) {
+            return undefined;
+        }
+        const { stdout } = await execFile("swiftly", ["--version"]);
+        return Version.fromString(stdout.trim());
+    }
+
+    /**
+     * Finds the list of toolchains managed by Swiftly.
+     *
+     * @returns an array of toolchain paths
+     */
+    public async getSwiftlyToolchainInstalls(): Promise<string[]> {
+        if (!this.isSupported()) {
+            return [];
+        }
+        const version = await swiftly.getSwiftlyVersion();
+        if (version?.isLessThan(new Version(1, 1, 0))) {
+            return await this.getToolchainInstallLegacy();
+        }
+
+        return await this.getListAvailableToolchains();
+    }
+
+    private async getListAvailableToolchains(): Promise<string[]> {
+        try {
+            const { stdout } = await execFile("swiftly", ["list-available", "--format=json"]);
+            const response = ListAvailableResult.parse(JSON.parse(stdout));
+            return response.toolchains.map(t => t.name);
+        } catch (error) {
+            throw new Error("Failed to retrieve Swiftly installations from disk.");
+        }
+    }
+
+    private async getToolchainInstallLegacy() {
+        try {
+            const swiftlyHomeDir: string | undefined = process.env["SWIFTLY_HOME_DIR"];
+            if (!swiftlyHomeDir) {
+                return [];
+            }
+            const swiftlyConfig = await swiftly.getSwiftlyConfig();
+            if (!swiftlyConfig || !("installedToolchains" in swiftlyConfig)) {
+                return [];
+            }
+            const installedToolchains = swiftlyConfig.installedToolchains;
+            if (!Array.isArray(installedToolchains)) {
+                return [];
+            }
+            return installedToolchains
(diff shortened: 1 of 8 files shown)


src/toolchain/swiftly.ts
```diff
@@ -0,0 +1,157 @@
...
+        }
+        const swiftlyConfigRaw = await fs.readFile(
+            path.join(swiftlyHomeDir, "config.json"),
+            "utf-8"
+        );
+        return JSON.parse(swiftlyConfigRaw);
+    }
+}
+
+export const swiftly = new Swiftly();
```
matthewbastien: Having global state like this will make it harder to test future integrations with Swiftly. This should probably be constructed in `WorkspaceContext`. Things like `ToolchainSelection.ts` can then either have it passed in as a dependency or pull it out of the workspace.

test/unit-tests/toolchain/swiftly.test.ts
```diff
@@ -0,0 +1,119 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2021 the VS Code Swift project authors
```
matthewbastien:
```suggestion
// Copyright (c) 2025 the VS Code Swift project authors
```

src/toolchain/swiftly.ts
```diff
@@ -0,0 +1,157 @@
+import * as path from "node:path";
```
matthewbastien: Missing copyright header.

src/toolchain/swiftly.ts
```diff
@@ -0,0 +1,157 @@
...
+                // it respects the .swift-version file in the cwd and resolves using that.
+                try {
+                    const inUse = await swiftly.swiftlyInUseLocation("swiftly", cwd);
+                    if (inUse.length > 0) {
+                        return path.join(inUse, "usr");
+                    }
+                } catch (err: unknown) {
+                    const error = err as ExecFileError;
+                    // Its possible the toolchain in .swift-version is misconfigured or doesn't exist.
+                    void vscode.window.showErrorMessage(`${error.stderr}`);
```
matthewbastien: This seems like something we should be logging and then show a better error message to the user. Something like `"Failed to load toolchain from Swiftly: ${error.stderr}"`.

What are the possible error messages from Swiftly here? Asking because I'm not sure how well VS Code notifications deal with newlines and such.

src/toolchain/swiftly.ts
```diff
@@ -0,0 +1,157 @@
...
+            }
+            const installedToolchains = swiftlyConfig.installedToolchains;
+            if (!Array.isArray(installedToolchains)) {
+                return [];
+            }
+            return installedToolchains
+                .filter((toolchain): toolchain is string => typeof toolchain === "string")
+                .map(toolchain => path.join(swiftlyHomeDir, "toolchains", toolchain));
+        } catch (error) {
+            throw new Error("Failed to retrieve Swiftly installations from disk.");
```
matthewbastien: Should probably be logging the original error here rather than swallowing it entirely. Either that or append the message here with `${error.message}`.

test/unit-tests/toolchain/swiftly.test.ts
```diff
@@ -0,0 +1,119 @@
...
+import * as sinon from "sinon";
+import { Swiftly } from "../../../src/toolchain/swiftly";
+import * as utilities from "../../../src/utilities/utilities";
+
+suite("Swiftly Unit Tests", () => {
+    let execFileStub: sinon.SinonStub;
+    let swiftly: Swiftly;
+
+    setup(() => {
+        execFileStub = sinon.stub(utilities, "execFile");
```
matthewbastien: We actually have some utilities that will mock and do the teardown automatically:

```typescript
import { mockGlobalObject } from "../MockUtils.ts";
import * as utilities from "../../../src/utilities/utilities";

suite("Swiftly Unit Tests", () => {
    const mockExecFile =  mockGlobalObject(utilities, "execFile");
    let swiftly: Swiftly;

    ...
}
```

This way you don't need the `teardown()` block at all.

test/unit-tests/toolchain/swiftly.test.ts
```diff
@@ -0,0 +1,119 @@
...
+            const originalPlatform = process.platform;
+            Object.defineProperty(process, "platform", {
+                value: "win32",
+                writable: true
+            });
+
+            const result = await swiftly.getSwiftlyToolchainInstalls();
+
+            expect(result).to.deep.equal([]);
+            expect(execFileStub.called).to.be.false;
```
matthewbastien:
```suggestion
            expect(execFileStub).to.not.have.been.called;
```

test/unit-tests/toolchain/swiftly.test.ts
```diff
@@ -0,0 +1,119 @@
...
+
+            const result = await swiftly.getSwiftlyToolchainInstalls();
+
+            expect(result).to.deep.equal([
+                "swift-5.9.0-RELEASE",
+                "swift-5.8.0-RELEASE",
+                "swift-DEVELOPMENT-SNAPSHOT-2023-10-15-a"
+            ]);
+
+            expect(execFileStub.calledWith("swiftly", ["--version"])).to.be.true;
```
matthewbastien:
```suggestion
            expect(execFileStub).to.have.been.calledWith("swiftly", ["--version"]);
```

test/unit-tests/toolchain/swiftly.test.ts
```diff
@@ -0,0 +1,119 @@
...
+            expect(execFileStub.calledWith("swiftly", ["list-available", "--format=json"])).to.be.true;
```
matthewbastien:
```suggestion
            expect(execFileStub).to.have.been.calledWith("swiftly", ["list-available", "--format=json"]);
```

matthewbastien (approved):
Thanks! Just a few things to clean up and a question about the failure paths for `swiftlyToolchain()`.

matthewbastien (changes requested):
Whoops. Meant to request changes, not approve.

src/toolchain/swiftly.ts
roulpriya: Sure, will add the suggested line.

> What are the possible error messages from Swiftly here? Asking because I'm not sure how well VS Code notifications deal with newlines and such.

I think `toolchain not found`

src/toolchain/swiftly.ts
roulpriya: I am trying to use it as a Singleton. Let me try moving it to workspace context

roulpriya: We use `SwiftToolchain` in the extention startup, at this point we don't have a `WorkspaceContext`
These functions are effectively static functions, so it will be possible for me to just define them at the module level

matthewbastien: Good point. Perhaps it just makes sense to instantiate a `Swiftly` instance in the extension activation and then pass it into things that need it? As far as I'm aware, it should only be used by some commands.

matthewbastien: Actually yeah I like your idea. Just make it all static. There's no internal state here anyway.

test/unit-tests/toolchain/swiftly.test.ts
```diff
@@ -0,0 +1,109 @@
...
+                "swift-5.8.0-RELEASE",
+                "swift-DEVELOPMENT-SNAPSHOT-2023-10-15-a"
+            ]);
+
+            expect(mockUtilities.execFile).to.have.been.calledWith("swiftly", ["--version"]);
+            expect(mockUtilities.execFile).to.have.been.calledWith("swiftly", ["list-available", "--format=json"]);
+        });
+
+        test("should return empty array when platform is not supported", async () => {
+            const originalPlatform = process.platform;
```
matthewbastien: We have another utility for this one:
```typescript
import { mockGlobalValue } from "../MockUtils";

suite("Test Suite", () => {
    const platformMock = mockGlobalValue(process, "platform");
 
    test("test case", () => {
        platformMock.setValue("linux");
    });
});
```

src/toolchain/swiftly.ts
matthewbastien: Can we also log it to the output channel? You'll need to pass it in kind of like how some of the static functions in SwiftToolchain do it.

src/toolchain/swiftly.ts
matthewbastien: Can we also log it to the output channel? You'll need to pass it in kind of like how some of the static functions in SwiftToolchain do it.

src/toolchain/swiftly.ts
```diff
@@ -0,0 +1,175 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of VS Code Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import * as path from "node:path";
```
award999: nit: you don't need the "node:" prefixes, "path" works

award999 (commented):
looks good to me

test/unit-tests/toolchain/swiftly.test.ts
```diff
@@ -0,0 +1,106 @@
...
+//===----------------------------------------------------------------------===//
+
+import { expect } from "chai";
+import { Swiftly } from "../../../src/toolchain/swiftly";
+import * as utilities from "../../../src/utilities/utilities";
+import { mockGlobalModule, mockGlobalValue } from "../../MockUtils";
+
+suite("Swiftly Unit Tests", () => {
+    const mockUtilities = mockGlobalModule(utilities);
+    const mockedPlatform = mockGlobalValue(process, "platform");
```
matthewbastien, line 22: The Windows tests are failing because the default platform there will be `win32`. You'll want to add a setup() block and set the platform to "darwin" by default if you want them to work on Windows.

```typescript
setup(() => {
    mockedPlatform.setValue("darwin");
});
```

test/unit-tests/toolchain/toolchain.test.ts
```diff
@@ -351,8 +353,8 @@ suite("SwiftToolchain Unit Test Suite", () => {
 
             mockFS({});
 
-            await expect(SwiftToolchain.getSwiftlyToolchainInstalls()).to.be.rejectedWith(
-                "Failed to retrieve Swiftly installations from disk."
+            await expect(Swiftly.listAvailableToolchains()).to.be.rejectedWith(
+                "Failed to retrieve Swiftly installations from disk: ENOENT, no such file or directory '/home/user/.swiftly/config.json'"
```
matthewbastien: This test is also failing on Windows because the path has back slashes instead of forward slashes. You'll either want to use `path.sep` or `path.join()` to convert the path in the error message to a platform-compatible one.

matthewbastien (changes requested):

test/unit-tests/toolchain/toolchain.test.ts
matthewbastien: Or you can just match on the start of the error and ignore the platform-specific portion:

```typescript
await expect(Swiftly.listAvailableToolchains()).to.be.rejected.then((error) => {
    expect(error.message).to.include("Failed to retrieve Swiftly installations from disk");
});
```

matthewbastien (approved):
Looks great. Thank you!

Issue #1208, opened 2024-11-16
Provide JSON validation for config files


MahdiBM:
**Is your feature request related to a problem? Please describe.**
This extension should provide [JSON schema](https://json-schema.org/draft-07)s for validation, for JSON config files related to this extension.
For example for `.sourcekit-lsp/config.json`, [when sourcekit-lsp supports that](https://github.com/swiftlang/sourcekit-lsp/issues/1829).
Then users will have an easier time writing config files.

Please Note that VS Code does not fully support newer version of JSON schema specification and we'll probably need to stick to version `draft-07`.

For the JSON validations to be enabled, this extension can use the `package.json` file. See [rust-analyzer](https://github.com/rust-lang/rust-analyzer/blob/9224ec44978b6817510b0b2999c7cf25c1ab837f/editors/code/package.json#L3212C10-L3212C24) as an example.

**Describe alternatives you've considered**
Users can manually use [`"$schema": "link/to/schema"`](https://code.visualstudio.com/Docs/languages/json#_json-schemas-and-settings) in the config JSON files to manually enable schema validations. See the link for a full of the options.

**Additional context**

award999:
With the work done by https://github.com/swiftlang/sourcekit-lsp/pull/1849, this should be straightforward and helpful. The unfortunate thing is that you can only statically set this file as a manifest contribution.
```
 "jsonValidation": [
      {
        "fileMatch": "**/.sourcekit-lsp/config.json",
        "url": "https://raw.githubusercontent.com/swiftlang/sourcekit-lsp/refs/heads/main/config.schema.json"
      }
    ],
```

This means we cannot set it based on the sourcekit-LSP version that is actually being used. Dynamically setting this has been requested issue against VSCode, but doesn't have a lot of traction by the looks of it: https://github.com/microsoft/vscode/issues/230136

So since we're stuck with static, there is then a couple possible issues:

1. Do we pin to "main" or to the latest stable release version of the schema? The pro for main is we don't have to update it for each release. We also have no idea which version they may be using (could be old) so we could let them see all the new configurations as a slight incentive to upgrade toolchain versions. The pro for latest release version is maybe we don't want to push people to main, although it has only been updated once since 6.1.
2. Do we bundle the schema statically in the extension, or specify a Github URL like above? Static is always there but we'd then have to update the file manually for new schema versions. Using the URL you dynamically you get updates for free, but may hit a GH outage.

[...]

PR #1726, opened 2025-07-18
Add a new `Generate SourceKit-LSP Configuration` command

award999:
## Description
- Add a new command that will generate the `.sourcekit-lsp/config.json` file for one or all the folders in the workspace
- Separate the select folder quick pick used to generate launch configurations so it could be reused
- Add `configurationBranch` setting so the user can hardcode
- Add a `jsonValidation` contribution to the manifest, the `$schema` property added by command takes precedence
- We'll always fall back to "main" if the expected branch does not exist
- Add new logic where if "-dev" suffix in version, we assume it is a "main" build

*Note* in a follow up PR I'm going to add logic to watch this file and prompt them to update the schema if the toolchain version changes

Issue: #1208

## Tasks

### DIFF
--- a/src/commands/generateSourcekitConfiguration.ts
+++ b/src/commands/generateSourcekitConfiguration.ts
@@ -26,7 +26,7 @@ export async function generateSourcekitConfiguration(ctx: WorkspaceContext): Pro
 
     if (ctx.folders.length === 1) {
         const folder = ctx.folders[0];
-        const success = await createSourcekitConfiguration(folder);
+        const success = await createSourcekitConfiguration(ctx, folder);
         void vscode.window.showTextDocument(vscode.Uri.file(sourcekitConfigFilePath(folder)));
         return success;
     }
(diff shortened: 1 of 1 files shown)


src/commands/generateSourcekitConfiguration.ts
```diff
@@ -0,0 +1,106 @@
...
+
+async function createSourcekitConfiguration(folderContext: FolderContext): Promise<boolean> {
+    const sourcekitFolder = vscode.Uri.file(sourcekitFolderPath(folderContext));
+    const sourcekitConfigFile = vscode.Uri.file(sourcekitConfigFilePath(folderContext));
+
+    try {
+        await vscode.workspace.fs.stat(sourcekitConfigFile);
+        return true;
+    } catch {
+        // Ignore, don't care if the file doesn't exist yet
```
matthewbastien, line 69: **nit:** check the error's `code` property to see if it's `ENOENT`. Otherwise log it.

matthewbastien (approved):

src/commands/generateSourcekitConfiguration.ts
award999: will do

Issue #1725, opened 2025-07-17
Swift Environment Variables setting is not used when running swift command plugins


adam-fowler:
**Describe the bug**
The Swift Environment Variables setting is not used when running swift command plugins. 

Given the extension uses the swift environment variables to set DEVELOPER_DIR it is probably a good idea we do include them in the environment while running swift command plugins

rbenegal:
Verified with `9f874ecd` using a CommandPlugin to check that the env variable is shown when run.

PR #1727, opened 2025-07-18
Include environment when resolving plugin task

award999:
## Description
Provide environment variables when `resolveTask` is called on a `swift-plugin` type task.

Issue: #1725

## Tasks


award999:
@adam-fowler it was being set when creating the plugin task: https://github.com/swiftlang/vscode-swift/blob/main/src/tasks/SwiftPluginTaskProvider.ts#L149

but not when resolveTask was called if it was in your tasks.json. Even if not in your tasks.json, there are cases where resolveTask may still get called

Issue #1728, opened 2025-07-19
Test-run struggles under a lot of tests


MahdiBM:
**Describe the bug**
Test-run struggles under a lot of tests, and can take much longer time than it would if you used CLI to run the tests.
I'm guessing the swift-testing output parsing is the bottleneck. 

**To Reproduce**
* Clone https://github.com/MahdiBM/swift-dns.
* Run `IDNATests` ([here](https://github.com/MahdiBM/swift-dns/blob/main/Tests/DNSTests/IDNATests.swift)).

There are 2 tests, each with 6400 swift-testing arguments. So basically 2x6400 tests.

**Expected behavior**
Test Explorer should show the result in under 10s.

CLI example:
```swift
$ swift test --filter "IDNATests." 2>&1 | grep IDNATests
◇ Suite IDNATests started.
✔ Suite IDNATests passed after 1.683 seconds.
```

**Environment**
macOS 26 is required if running on macOS. Swift 6.2 nightly (use Xcode toolchains).
Otherwise run in the devcontainer included with the project (although the swift docker image is pinned and hasn't been updated recently, maybe you'd want to remove the SHA256 pinning of the docker image if you want to use devcontainers.)

**Additional context**

plemarquand:
There is definitely room for optimization here. A cpu profile shows most of the time is spent in clearEnqueuedTest. I put up https://github.com/swiftlang/vscode-swift/pull/1734 which took runtime of the IDNATests from ~45 seconds to ~6 seconds.

MahdiBM:
Thank you 🙏 

There should be more room for improvements as well considering CLI takes less than 2 seconds, but 6 seconds is still pretty fine and I would be happy with it.

plemarquand:
There is always going to be some overhead involved updating the VS Code test items (and in turn the UI) but we should be striving to keep it within the same order of magnitude.

plemarquand:
Actually I may have spoke too soon, it seems that now most of the time is spent in VS Code itself, but its still quite a while.

award999:
Verified 40bdb780debc3c1bb9e4671d89f051a8270aa6b4 that the UI does show the shorter time. There is a longer amount of time spent updating the UI due to a vscode bottleneck

<img width="1090" height="170" alt="Image" src="https://github.com/user-attachments/assets/8cc45a79-2e57-4bca-bf4b-5b74eb995b3a" />

MahdiBM:
[...]

PR #1734, opened 2025-07-21
Optimize clearing enqueued tests

plemarquand:
## Description
Refactor the enqueued tests list to be a Set so we can clear enqueued tests without the need to potentially traverse the full remaining list of enqueued tests.

Also avoid recreating the test item finder on every `get testItemFinder`, and store test items in the darwin test item finder in a Set as well for faster lookup.

Issue: #1728


award999 (commented):
CHANGELOG but looks good otherwise

PR #1736, opened 2025-07-21
Notify user if configured $schema is out of date

award999:
## Description
This change will notify users when they open their .sourcekit-lsp/config.json file if the $schema property does not exist or if the value does not match the latest computed schema URL. There is a setting that disables this check and it is set if they click the "Don't Ask Again" button

Issue: #1208

## Tasks

### DIFF
--- a/src/commands/generateSourcekitConfiguration.ts
+++ b/src/commands/generateSourcekitConfiguration.ts
@@ -112,10 +112,12 @@ const schemaURL = (branch: string) =>
 async function checkURLExists(url: string): Promise<boolean> {
     try {
         const response = await fetch(url, { method: "HEAD" });
-        if (response.status >= 500) {
+        if (response.ok) {
+            return true;
+        } else if (response.status !== 404) {
             throw new Error(`Received exit code ${response.status} when trying to fetch ${url}`);
         }
-        return response.ok;
+        return false;
     } catch {
         return false;
     }
--- a/src/commands/generateSourcekitConfiguration.ts
+++ b/src/commands/generateSourcekitConfiguration.ts
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import { join } from "path";
+import { basename, dirname, join } from "path";
 import * as vscode from "vscode";
 import { FolderContext } from "../FolderContext";
 import { selectFolder } from "../ui/SelectFolderQuickPick";
@@ -187,7 +187,10 @@ export async function handleSchemaUpdate(
 ) {
     if (
         !configuration.checkLspConfigurationSchema ||
-        !doc.uri.fsPath.endsWith("/.sourcekit-lsp/config.json")
+        !(
+            basename(dirname(doc.uri.fsPath)) === ".sourcekit-lsp" &&
+            basename(doc.uri.fsPath) === "config.json"
+        )
     ) {
         return;
     }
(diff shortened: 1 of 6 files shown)


award999:
There already is a changelog entry for this one from #1726 https://github.com/swiftlang/vscode-swift/blob/main/CHANGELOG.md

src/commands/generateSourcekitConfiguration.ts
```diff
@@ -114,8 +112,95 @@ const schemaURL = (branch: string) =>
...
+    } catch (error) {
+        if ((error as NodeJS.ErrnoException).code !== "ENOENT") {
+            workspaceContext.outputChannel.appendLine(
+                `Failed to read file at ${doc.uri.fsPath}: ${error}`
+            );
+        }
+        return;
+    }
+    const contents = Buffer.from(buffer).toString("utf-8");
+    const config = JSON.parse(contents);
```
plemarquand: nit: What do we do if the JSON is malformed?

src/commands/generateSourcekitConfiguration.ts
```diff
@@ -114,8 +112,95 @@ const schemaURL = (branch: string) =>
 async function checkURLExists(url: string): Promise<boolean> {
     try {
         const response = await fetch(url, { method: "HEAD" });
+        if (response.status >= 500) {
```
plemarquand: This should also handle 404 statuses

src/commands/generateSourcekitConfiguration.ts
award999: woooo we were definitely on the same wavelength for a minute https://github.com/swiftlang/vscode-swift/pull/1736/commits/9c1e0e88837c51528adb579eb90211e979814a6e

src/commands/generateSourcekitConfiguration.ts
award999: oooh so response.ok checks it is >= 200 and < 400. I put in this 500 check as an exception thinking that a GitHub infra issue was what was failing CI but I was wrong. Will revert

award999: ohh wait nvm this is what I then changed it to. So if 404 it doesn't exist so fallback to "main". 500 just terminate the command. Although I should just be checking for 404 so will change that

src/commands/generateSourcekitConfiguration.ts
```diff
@@ -112,7 +112,7 @@ const schemaURL = (branch: string) =>
 async function checkURLExists(url: string): Promise<boolean> {
     try {
         const response = await fetch(url, { method: "HEAD" });
-        if (response.status >= 500) {
+        if (response.status !== 404) {
```
plemarquand: I think this is the opposite of what we want

award999: If there is no config file like prior to 6.1 I believe, wanted to return false to fallback to "main". All other errors I wanted to notify the error code, like 500 show user the error but not continue. Mainly don't want a 500 error to result in user being prompted to update the schema file

award999: ooooooooh *face palm*

src/commands/generateSourcekitConfiguration.ts
```diff
@@ -114,8 +112,105 @@ const schemaURL = (branch: string) =>
 async function checkURLExists(url: string): Promise<boolean> {
     try {
         const response = await fetch(url, { method: "HEAD" });
-        return response.ok;
+        if (response.ok) {
+            return true;
+        } else if (response.status !== 404) {
+            throw new Error(`Received exit code ${response.status} when trying to fetch ${url}`);
```
award999, line 118:
```suggestion
            throw new Error(`Received response code ${response.status} when trying to fetch ${url}`);
```

rbenegal (approved):

Issue #1740, opened 2025-07-23
Error for "run test multiple times" or "until failure": Unexpected argument undefined at index 1


MahdiBM:
**Describe the bug**
What the title says:
Error for "run test multiple times" or "until failure": Unexpected argument undefined at index 1 

**To Reproduce**
* Clone https://github.com/MahdiBM/swift-dns.
* Try to run any single test using one of those 2 buttons.

**Expected behavior**
Just work.

**Environment**
 - OS: [e.g. Ubuntu 20.04] macOS 26
 - Swift version (output of `swift --version`): Xcode 26 beta 4 toolchain (6.2 nightly)
 - Visual Studio Code version:  1.99.3
 - vscode-swift version: 2 nights ago's nightly

**Additional context**

MahdiBM:
A quick search showed it must be this line: https://github.com/swiftlang/vscode-swift/blob/b781e0eb5277629efcff7dfd66e8694b32223257/src/commands.ts#L311

So i'm trying to debug this myself as there is a chance this might require my specific environment to reproduce (not sure).

MahdiBM:
This is the function call and the args:

<kbd> <img width="1938" height="491" alt="Image" src="https://github.com/user-attachments/assets/be1e3ee8-ca27-4a5b-aa38-705b6a2f964f" /> </kbd>

I'm going to stop myself from going deep down the rabbit hole, hopefully the image helps.

matthewbastien:
Thanks for the bug report!

I forgot to test from the editor gutter menu items when I was fixing another bug for these commands.

award999:
Verified with dca7546

PR #1742, opened 2025-07-23
Fix error when running tests multiple times from the editor gutter

matthewbastien:
## Description
The VS Code editor gutter menu actions for tests pass in an extra `undefined` argument to `Run tests multiple times` and `Run until failure`. Properly handle this in `extractTestItemsAndCount()`.

Issue: #1740

## Tasks
- ~~[ ] Documentation has been updated~~
- ~~[ ] Added an entry to CHANGELOG.md if applicable~~

No need to update CHANGELOG as this is a bug that only exists in pre-release.

### DIFF
--- a/src/commands/testMultipleTimes.ts
+++ b/src/commands/testMultipleTimes.ts
@@ -96,19 +96,22 @@ export async function runTestMultipleTimes(
  * also accept a final count parameter. We have to find the count parameter ourselves since JavaScript
  * only supports varargs at the end of an argument list.
  */
-export function extractTestItemsAndCount(...args: (vscode.TestItem | number)[]): {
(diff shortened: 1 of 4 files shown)


src/commands/testMultipleTimes.ts
```diff
@@ -86,3 +85,43 @@ export async function runTestMultipleTimes(
...
+): {
+    testItems: vscode.TestItem[];
+    count?: number;
+} {
+    const result = args.reduce<{
+        testItems: vscode.TestItem[];
+        count?: number;
+    }>(
+        (result, arg, index) => {
+            if (arg === undefined || arg === null) {
```
plemarquand, line 110: nit: roll this in to the if/else tree below

Issue #1739, opened 2025-07-22
Use Swiftly's list available command to find toolchains


matthewbastien:
Swiftly will soon be supporting JSON output for its commands including list available. Use this instead of parsing the config file which is not recommended.

matthewbastien:
Verification failed with b781e0e.

I built a local copy of Swiftly 1.1.0-dev from main and installed it into my `~/.swiftly/bin/` folder. Running the VS Code command `Swift: Select toolchain...` fails with an incredibly lengthy error message from zod about unexpected JSON output.

cc: @roulpriya

MahdiBM:
I got the zod error today when trying to manually debug an issue (see #1740). So just saying it's probably unrelated to this specific issue.
Not a js dev so not sure what exactly went on, but a `npm install` solved the issue I think.

award999:
> I got the zod error today when trying to manually debug an issue (see [#1740](https://github.com/swiftlang/vscode-swift/issues/1740)). So just saying it's probably unrelated to this specific issue. Not a js dev so not sure what exactly went on, but a `npm install` solved the issue I think.

That was a new module that was introduced within the last couple days. Do an `npm ci` anytime you pull changes to be safe :)

matthewbastien:
Verified with d2dc6df

PR #1743, opened 2025-07-23
Fixed JSON parsing error while selecting toolchain

roulpriya:
## Description
In the swiftly toolchain version schema, there was a missing `name` attribute, which prevented it from parsing the value. In this fix, I have removed the `name` field from the top-level and added it at the object level. 

Issue: https://github.com/swiftlang/vscode-swift/issues/1739

## Tasks

### DIFF
--- a/src/toolchain/swiftly.ts
+++ b/src/toolchain/swiftly.ts
@@ -46,6 +46,10 @@ const ListResult = z.object({
     ),
 });
 
+const InUseVersionResult = z.object({
+    version: z.string(),
+});
+
 export class Swiftly {
     /**
      * Finds the version of Swiftly installed on the system.
@@ -69,6 +73,27 @@ export class Swiftly {
     }
 
     /**
+     * Checks if the installed version of Swiftly supports JSON output.
+     *
+     * @returns `true` if JSON output is supported, `false` otherwise.
+     */
+    private static async supportsJsonOutput(
+        outputChannel?: vscode.OutputChannel
+    ): Promise<boolean> {
+        if (!Swiftly.isSupported()) {
+            return false;
+        }
+        try {
+            const { stdout } = await execFile("swiftly", ["--version"]);
+            const version = Version.fromString(stdout.trim());
+            return version?.isGreaterThanOrEqual(new Version(1, 1, 0)) ?? false;
+        } catch (error) {
+            outputChannel?.appendLine(`Failed to check Swiftly JSON support: ${error}`);
+            return false;
+        }
+    }
+
+    /**
      * Finds the list of toolchains managed by Swiftly.
      *
      * @returns an array of toolchain paths
@@ -85,7 +110,7 @@ export class Swiftly {
             return [];
         }
 
-        if (version.isLessThan(new Version(1, 1, 0))) {
+        if (!(await Swiftly.supportsJsonOutput(outputChannel))) {
             return await Swiftly.getToolchainInstallLegacy(outputChannel);
         }
 
@@ -141,6 +166,25 @@ export class Swiftly {
         return inUse.trimEnd();
     }
 
+    public static async inUseVersion(
+        swiftlyPath: string = "swiftly",
+        cwd?: vscode.Uri
+    ): Promise<string | undefined> {
+        if (!this.isSupported()) {
+            throw new Error("Swiftly is not supported on this platform");
+        }
+
+        if (!(await Swiftly.supportsJsonOutput())) {
+            return undefined;
+        }
+
+        const { stdout } = await execFile(swiftlyPath, ["use", "--format=json"], {
+            cwd: cwd?.fsPath,
+        });
+        const result = InUseVersionResult.parse(JSON.parse(stdout));
+        return result.version;
+    }
+
     public static async use(version: string): Promise<void> {
         if (!this.isSupported()) {
             throw new Error("Swiftly is not supported on this platform");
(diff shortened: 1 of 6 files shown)


award999:
Was the issue that there was a missing "name" property? If so would that mean if swiftly changed the schema in the future then the extension could fail again? Can you maybe update the PR description to explain what this issue was and why this fixes it

roulpriya:
> Was the issue that there was a missing "name" property? If so would that mean if swiftly changed the schema in the future then the extension could fail again? Can you maybe update the PR description to explain what this issue was and why this fixes it

I don't expect things to change once we release Swiftly with JSON support. Swiftly will ensure backwards compatibility with the JSON format. Newer formats will always be opt-in

matthewbastien:
> I don't expect things to change once we release Swiftly with JSON support. Swiftly will ensure backwards compatibility with the JSON format. Newer formats will always be opt-in

I guess my bigger question would be: can we only check for the fields that we actually care about and ignore the rest? E.g. the name field can just be omitted if we aren't using it at all.

I can imagine Swiftly might add new information to the output later on. This wouldn't be considered a breaking change from Swiftly's perspective, but would cause the current code to throw an error. (As long as I'm understanding this right)

award999:
that is a better worded version of my question, thanks @matthewbastien

src/toolchain/swiftly.ts
```diff
@@ -97,9 +96,9 @@ export class Swiftly {
...
         try {
-            const { stdout } = await execFile("swiftly", ["list-available", "--format=json"]);
-            const response = ListAvailableResult.parse(JSON.parse(stdout));
-            return response.toolchains.map(t => t.name);
+            const { stdout } = await execFile("swiftly", ["list", "--format=json"]);
+            const response = ListResult.parse(JSON.parse(stdout));
+            return response.toolchains.map(t => t.version.name);
         } catch (error) {
             outputChannel?.appendLine(`Failed to retrieve Swiftly installations: ${error}`);
             throw new Error(
```
matthewbastien: From my testing it looks like VS Code actually displays this error message to the user. We should probably just return an empty array instead since we're already logging the issue. That way toolchain selection won't completely break if this logic fails in the future for whatever reason.

matthewbastien: Or have the downstream code catch the error. I'm fine with either option.

src/toolchain/swiftly.ts
```diff
@@ -20,26 +20,25 @@ import * as vscode from "vscode";
...
             version: z.discriminatedUnion("type", [
                 z.object({
                     major: z.number(),
                     minor: z.number(),
                     patch: z.number().optional(),
+                    name: z.string(),
                     type: z.literal("stable"),
                 }),
                 z.object({
                     major: z.number(),
```
matthewbastien: **bug:** Looks like major and minor are also optional. This is still failing on my machine. Here's the output of `swiftly list --format json`:

```json
{
  "toolchains" : [
    {
      "inUse" : true,
      "isDefault" : true,
      "version" : {
        "branch" : "6.2",
        "date" : "2025-07-02",
        "major" : 6,
        "minor" : 2,
        "name" : "6.2-snapshot-2025-07-02",
        "type" : "snapshot"
      }
    },
    {
      "inUse" : false,
      "isDefault" : false,
      "version" : {
        "branch" : "6.2",
        "date" : "2025-05-15",
        "major" : 6,
        "minor" : 2,
        "name" : "6.2-snapshot-2025-05-15",
        "type" : "snapshot"
      }
    },
    {
      "inUse" : false,
      "isDefault" : false,
      "version" : {
        "branch" : "main",
        "date" : "2025-04-12",
        "name" : "main-snapshot-2025-04-12",
        "type" : "snapshot"
      }
    },
    {
      "inUse" : false,
      "isDefault" : false,
      "version" : {
        "major" : 6,
        "minor" : 1,
        "name" : "6.1.2",
        "patch" : 2,
        "type" : "stable"
      }
    },
    {
      "inUse" : false,
      "isDefault" : false,
      "version" : {
        "major" : 6,
        "minor" : 1,
        "name" : "6.1.0",
        "patch" : 0,
        "type" : "stable"
      }
    }
  ]
}
```

matthewbastien (changes requested):
Looking great! Just one more thing I noticed when testing the latest version of this PR on my machine.

**suggestion:** I don't think we need the `description` anymore for Swiftly toolchains in the toolchain selection dialog. It ends up just showing the same info twice now. We also don't need the `toolchainPath` or the `swiftFolderPath` since we're looking that up after the selection is made now. Both of these are going to take some amount of refactoring that could be done in a future PR, though. Up to you.

src/ui/ToolchainSelection.ts
```diff
@@ -304,19 +320,21 @@ export async function showToolchainSelectionQuickPick(activeToolchain: SwiftTool
...
+        let swiftPath: string;
 
         // Handle Swiftly toolchains specially
         if (selected.category === "swiftly") {
             try {
                 // Run swiftly use <version> and get the path to the toolchain
                 await Swiftly.use(selected.label);
-                const inUseLocation = await Swiftly.inUseLocation("swiftly");
-                swiftPath = path.join(inUseLocation, "usr", "bin");
+                swiftPath = Swiftly.getBinDir();
```
matthewbastien: **issue:** Using the Swiftly bin directory is going to break some of our logic for finding stuff like the LLDB library path. You can just leave `swiftPath` as `undefined` for now which will cause `setToolchainPath()` to remove `swift.path` from settings and allow Swiftly to manage everything.

src/ui/ToolchainSelection.ts
```diff
@@ -304,19 +320,21 @@ export async function showToolchainSelectionQuickPick(activeToolchain: SwiftTool
...
             } catch (error) {
                 void vscode.window.showErrorMessage(`Failed to switch Swiftly toolchain: ${error}`);
                 return;
             }
+        } else {
+            // For non-Swiftly toolchains, use the swiftFolderPath
+            swiftPath = (selected as PublicSwiftToolchainItem | XcodeToolchainItem).swiftFolderPath;
```
matthewbastien: **nit:** the type assertion here isn't necessary since you already checked for `category === "swiftly"` above.

```suggestion
            swiftPath = selected.swiftFolderPath;
```

src/ui/ToolchainSelection.ts
```diff
@@ -195,18 +201,28 @@ async function getQuickPickItems(
...
     if (activeToolchain) {
         const toolchainInUse = [...xcodes, ...toolchains, ...swiftlyToolchains].find(toolchain => {
-            return toolchain.toolchainPath === activeToolchain.toolchainPath;
+            if (activeToolchain.isSwiftlyManaged) {
+                if (toolchain.category !== "swiftly") {
+                    return false;
+                }
+
+                // For Swiftly toolchains, check if the label matches the active toolchain version
+                return toolchain.label === activeToolchain.swiftVersion.toString();
```
matthewbastien: **issue (non-blocking):** this won't work for snapshots or nightly branch based toolchains. We should ask Swiftly which toolchain is in use and mark that as active.

You can fix this in another PR, though. My top priority is getting the release unblocked.

matthewbastien (changes requested):

src/ui/ToolchainSelection.ts
roulpriya: Done

src/ui/ToolchainSelection.ts
```diff
@@ -178,7 +186,6 @@ async function getQuickPickItems(
                 type: "toolchain",
                 category: "public",
                 label: path.basename(toolchainPath, ".xctoolchain"),
-                detail: toolchainPath,
```
matthewbastien: **issue:** we should keep the detail for public toolchain installs.

roulpriya: Done

matthewbastien (approved):
Thank you so much for getting this ready!

PR #1745, opened 2025-07-24
Add `swift.recordTestDuration` setting

plemarquand:
When running a large test suite full of tests that complete very quickly, VS Code spends a lot of time updating the test explorer UI. For every test the tree is refreshed, and in a suite where thousands of tests complete very quickly this can make the UI unresponsive and increase the amount of time the test run takes.

The UI updates because the duration of the test item is different between runs, requiring the UI to change. As in the case described above, omitting the duration when passing or failing test items significantly improves performance.

Introduce the `swift.recordTestDuration` setting which, when set to false, will omit the duration when recording test results. The default is `true`, making this optimization opt-in.

This intended use is to opt out of duration recording on a per-project basis for projects that exhibit poor test explorer performance with many fast running tests.


award999 (approved):

award999 (commented):
Looks good, but changelog please

award999 (approved):

plemarquand:
@MahdiBM Setting `swift.recordTestDuration` to `false` will speed up your heavy test suite, once this lands you should try it out in `swift-dns`.

award999:
This PR is tracked by issue https://github.com/swiftlang/vscode-swift/issues/1728

PR #1756, opened 2025-07-29
Bump extension version

award999:
## Description
- Bump the extension version so targeting next release version, adding -dev to distinguish
- Decrement pre-release minor version since manifest will now have the next, not current version
- Allow -dev in vsix file name when downloading vsix for tests
- When downloading VSIX, add the actual file name to GITHUB_ENV, without renaming 
- Determine version from the VSIX filename or fallback to the package.json version
- Realized that in CI the VSIX was *NOT* being used for linux runs so fixed this and:
  - Moved wait for index logic into bundled extension because otherwise https://github.com/eclipse-glsp/glsp/discussions/1480 occurred
  - Added console logging in .vscode-test.js so we could see whether extension was being used and which `extensionDevelopmentPath`
- `skipExtensionDependencies` in CI
- The `RUNNING_IN_CI` does not work so introduce new logic to check if source is being run as part of the test CLI instead
  - Make "VSCODE_TEST" env var longer so harder to accidentally set
- 

Issue: N/A

## Tasks


award999:
The full nightly run also passed on my fork's action https://github.com/award999/vscode-swift/actions/runs/16681615407

Issue #1757, opened 2025-07-31
Optionally capture the lldb-dap logs into diagnostic bundle


award999:
With the work done in the LLDB DAP extension to provide log files (https://github.com/llvm/llvm-project/issues/146880), we can include those logs in the Swift extension diagnostics bundle. We should only include this in the "Full" diagnostic bundle, and need to make sure we add language in that prompt indicating that these logs will be included.

plemarquand:
Verified with 02a9529090815f25a90201dda3e3218ca1119d4f

PR #1758, opened 2025-07-31
Optionally add lldb-dap logs to Diagnostic Bundle

award999:
## Description
This change optionally includes the LLDB DAP extension logs in the diagnostic bundle. Updated language in the prompt so the user knows it is included.

Issue:  #1757

## Tasks

### DIFF
--- a/src/commands/captureDiagnostics.ts
+++ b/src/commands/captureDiagnostics.ts
@@ -25,6 +25,7 @@ import { Version } from "../utilities/version";
 import { destructuredPromise, execFileStreamOutput } from "../utilities/utilities";
 import configuration from "../configuration";
 import { FolderContext } from "../FolderContext";
+import { Extension } from "../utilities/extensions";
 
 export async function captureDiagnostics(
     ctx: WorkspaceContext,
@@ -75,6 +76,13 @@ export async function captureDiagnostics(
                     if (logFile) {
                         await copyLogFile(outputDir, logFile);
                     }
+                }
+
+                // Copy lldb-dap logs
+                const logFolder = lldbDapLogFolder(ctx);
+                const lldbLogFiles = await fsPromises.readdir(logFolder);
+                for (const log of lldbLogFiles) {
+                    await copyLogFile(outputDir, path.join(logFolder, log));
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


src/commands/captureDiagnostics.ts
```diff
@@ -137,20 +161,20 @@ async function captureDiagnosticsMode(
...
             `A Diagnostic Bundle collects information that helps the developers of the Swift for VS Code extension diagnose and fix issues.
 
-This information contains:
+This information includes:
 - Extension logs
+- Extension settings
 - Versions of Swift installed on your system
+
+If you allow capturing a Full Diagnostic Bundle, the information will also include:
+- Log message emitted by LLDB DAP
```
plemarquand:
```suggestion
- Log messages emitted by LLDB DAP
```

plemarquand: Should we add this line conditionally only if they're using LLDB DAP?

src/commands/captureDiagnostics.ts
```diff
@@ -44,13 +45,23 @@ export async function captureDiagnostics(
...
         await fsPromises.mkdir(diagnosticsDir);
-        await copyLogFile(diagnosticsDir, extensionLogFile(ctx));
 
         const singleFolderWorkspace = ctx.folders.length === 1;
         const zipDir = await createDiagnosticsZipDir();
         const zipFilePath = path.join(zipDir, `${path.basename(diagnosticsDir)}.zip`);
         const { archive, done: archivingDone } = configureZipArchiver(zipFilePath);
 
+        const archivedLldbDapLogFolders = new Set<string>();
+        if (captureMode === "Full") {
```
plemarquand: We should only do this if they're using LLDB DAP as their debugging extension, and ignore it for CodeLLDB

src/commands/captureDiagnostics.ts
award999: Done

src/commands/captureDiagnostics.ts
award999: Done

plemarquand (approved):

Issue #1747, opened 2025-07-26
Allow "Running tests multiple times" and "...until failure" under debugger


MahdiBM:
The issue are, running tests multiple times or until failure:
1- Try to rebuild the tests each run.
2- Are not in "Debug" mode aka with lldb attached, so they are of limited use.

I'd say they should just run in debug mode, but I'm fine with having the option to choose either, like when it asks for # of runs.

If it would run in debug mode, I could have used it to catch a intermittent bug by using a breakpoint and waiting for the breakpoint to get triggered.

plemarquand:
I just merged #1763 which adds a new option to debug multiple times and debug until failure. It should also be better about not rebuilding before each run.

award999:
Debugging multiple times and until failure works, but when I click Stop button for the debug session I'd expect it to stop further executions as well. Otherwise spawns all the remaining sessions and I cannot kill early. This is annoying enough we should probably fix before releasing

award999:
@plemarquand

award999:
Verified cba23ce

PR #1763, opened 2025-08-05
Add commands to debug tests multiple times

plemarquand:
Adds two new commands, `swift.debugTestsMultipleTimes` and `swift.debugTestsUntilFailure`. They behave the same as the existing run multiple and run until failure commands, except they use the debugging profile. When debugging multiple times a build is only performed for the first iteration, after which the resulting artifacts are run multiple times.

Also cleans up some of the multiple test run console output.

Issue: #1747

### DIFF
--- a/src/utilities/utilities.ts
+++ b/src/utilities/utilities.ts
@@ -147,6 +147,29 @@ export async function execFile(
     });
 }
 
+enum Color {
+    red = 31,
+    green = 32,
+    yellow = 33,
+    blue = 34,
+    magenta = 35,
+    cyan = 36,
+    white = 37,
+    grey = 90,
+    lightRed = 91,
+    lightGreen = 92,
+    lightYellow = 93,
+    lightBlue = 94,
+}
+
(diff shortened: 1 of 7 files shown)


award999:
Looks good but probably should add a changelog entry for new commands

src/utilities/utilities.ts
```diff
@@ -154,6 +154,29 @@ export async function execFile(
...
+    cyan = 36,
+    white = 37,
+    grey = 90,
+    lightRed = 91,
+    lightGreen = 92,
+    lightYellow = 93,
+    lightBlue = 94,
+}
+
+export function colorize(text: string, color: keyof typeof Color): string {
```
award999, line 172: my only concern is clashes with the users configured theme and if they have trouble seeing any of these colours. Can we extract from theme a warning, error, info, etc colour?

plemarquand: ANSI codes are automatically adjusted by themes in VS Code

Issue #1748, opened 2025-07-26
New test-run request should cancel any in-process test runs to begin the new test-run immediately


MahdiBM:
**Describe the bug**
Like in Xcode, when you hit the "test run" button, if there exists a running test-run, the old test-run should be immediately cancelled so the new test-run can begin.

**To Reproduce**
* Run a long-running test.
* Hit the run-test button again.
* Observe the second test-run request does nothing, at least until the first test is over.

**Expected behavior**
Explained above.

**Environment**
Doesn't matter.

**Additional context**

award999:
That is what the cancel test run button provided by VSCode does. Maybe one day VSCode will come up with similar logic as debug sessions where they ask the user if they want to cancel since they are managing the cancellation tokens

MahdiBM:
Hmm what about the extension at least letting users know that there is an existing test going on and that they need ro cancel that first?

award999:
Verified with e06a8270d43d91975a13a304d96b587c83c9456d

PR #1774, opened 2025-08-08
Prompt for cancellation when starting test run while one is in flight

plemarquand:
## Description
If a test run is already in progress and another one is started VS Code will queue it and start the second run once the first has finished.

Implement cancellation logic that prompts the user to cancel the active test run if one is in flight. If they chose to cancel the dialog, the second test run is not queued. If they do chose to cancel the active test run it is stopped and the new one starts immediately.

Issue: #1748

## Tasks
- [ ] ~Documentation has been updated~
- [X] Added an entry to CHANGELOG.md if applicable


award999 (approved):

src/TestExplorer/TestRunner.ts
```diff
@@ -429,7 +432,7 @@ export class TestRunner {
     private testArgs: TestRunArguments;
     private xcTestOutputParser: IXCTestOutputParser;
     private swiftTestOutputParser: SwiftTestingOutputParser;
-    private static CANCELLATION_ERROR = "Test run cancelled";
+    private static CANCELLATION_ERROR = "Test run cancelled.";
```
mzhfei, line 440: Hi, I'm just learning about this Repo at the moment and feel free to discard my question.

Would this error be used outside the class? Will there be any chances anywhere check against the value(I don't see one from a quick github search, there is a similar one but seems irrelevant)? And, probably not related to this PR, but what are the reasons that the error string is stored on the class as a static instead of in a constant file. 
Thanks in advance!

plemarquand: Good question! This is used in two places, one where it is thrown and [another where we check specifically if the error thrown is a cancellation error](https://github.com/swiftlang/vscode-swift/blob/187b848285ffbe431cce103a0ba715eb253e88ef/src/TestExplorer/TestRunner.ts#L784). 

I suppose it could live in a constants file, and probably would need to be moved out if we ever wanted to localize the extension, but for now it is encapsulated to just to the TestRunner class so for cohesion and simplicity we mark as private, and mark as static to indicate that this string should not bloat the TestRunner objects by being included in every instance.

mzhfei: Thanks a lot! That make sense. 😄

rbenegal (approved):

matthewbastien (approved):

PR #1779, opened 2025-08-12
Cleanup circular dependency in WorkspaceContext/FolderContext

plemarquand:
## Description
Break a circular dependency between WorkspaceContext and FolderContext. While WorkspaceContext can create FolderContexts the FolderContext doesn't need to know about WorkspaceContexts.

Also cleans up some promises that were being implicitly run by being prefixed with `void`.

## Tasks


plemarquand:
I'm going to break this in to smaller blocks of work

Issue #1002, opened 2024-08-08
XCTest debug session starts after stopping swift-testing session


award999:
**Describe the bug**
When I stop a swift-testing debug session, the XCTest one still proceeds

**To Reproduce**
1. Have a mix of swift-testing and XCTests
2. Debug all your tests in the test explorer
3. Hit a bp when running swift-testing
4. Press the stop button
5. Notice swift-testing stops, but XCTest session still proceeds

**Expected behavior**
Maybe through the `vscode.debug` or debug adapter tracker APIs we can check if it was stopped. Also is notion of parent debug sessions.

**Environment**
 - OS: macOS
 - Swift version (output of `swift --version`): 6.0
 - Visual Studio Code version: 1.92
 - vscode-swift version: 1.10.7

**Additional context**
Minor annoyance, feel free to just close if implementation is messy.

award999:
@plemarquand this was just a minor annoyance since startup of LLDB can be slow and I find some times it takes a while to respond to a stop or continue press. Not concerned about this if there is no elegant solution

plemarquand:
Verified with a74f2e95369a108f0e3d30344fc32d8bfa79c3b8

natinusala:
FWIW this is still happening to me with the extension v504,767

I debug the tests from the Test Explorer, then at some point I stop the debugging session. The XCTests immediately run after (which I don't even use, I don't know if there is a way to entirely disable XCTest?).

matthewbastien:
@natinusala can you double check that you're running on v2.16.2 which was just released (or the latest 2.17.20260317 pre-release). I'm not sure what v504,767 is.

natinusala:
It's the install count 🙄 since it's highlighted I thought it was the version

I'm on 2.17.20260317

plemarquand:
@natinusala what is the output in the Test Results panel when you see XCTests run unexpectedly?

natinusala:
@plemarquand we don't see much but it seems that the `.xctest` executable is run twice (once at the top, once at the bottom):

```
[...]

PR #1797, opened 2025-08-19
Don't debug further tests when stopping the debug session

award999:
## Description
Similar to #1790, if your debugging all tests in a project with a mix of XCTest and swift-testing suites, if the user stops a debug session for the swift-testing tests, then the XCTest debug session should not then startup

Issue: #1002

## Tasks


plemarquand (approved):
Nice!

PR #1800, opened 2025-08-21
Use `Invoke-Program` for the Windows build command

bnbarham:
Windows does not stop scripts when native commands exit with non-zero. Instead, their exit status has to be checked manually. This can be done through an `Invoke-Program` function that is added to the script running `windows_build_command`, which also previously prefixed the given command. This is changing in
https://github.com/swiftlang/github-workflows/pull/154 since it doesn't help with multi-line commands - update our modified `windows_build_command` to use `Invoke-Program` instead.


.github/workflows/nightly.yml
```diff
@@ -99,5 +99,5 @@ jobs:
         VSCODE_SWIFT_VSIX_PRERELEASE=1
         GITHUB_REPOSITORY=${{github.repository}}
       windows_pre_build_command: . .github\workflows\scripts\windows\setup.ps1
-      windows_build_command: scripts\test_windows.ps1
+      windows_build_command: .github/workflows/nightly.yml scripts\test_windows.ps1
```
jakepetroules: This looks odd... invoking a yaml file with a powershell script as an argument? how will that work?

bnbarham: Sorry, did these quickly 😅 Too much copy paste

PR #1802, opened 2025-08-23
Trigger parameter hints on accepting a function-like completion item

a7medev:
## Description

We're working on implementing signature help in SourceKit-LSP (https://github.com/swiftlang/sourcekit-lsp/pull/2250) and we faced a problem in VS Code where signature help wasn't triggered after accepting a completion item for a function, method, subscript, initializer, or enum case with associated values that would've normally had signature help if the user manually typed the `(` after the signature name.

Here's a demo of the issue with a development version of SourceKit-LSP:

https://github.com/user-attachments/assets/d44f2a95-ca78-40e7-a44f-b5a09ec61d58

To fix this, we added a `provideCompletionItem` middleware that inspects completion items that might have signature help (functions, methods (includes subscripts), initializers, and enum cases) and injects an `editor.action.triggerParameterHints` command (if no other command was present). This results in signature help working pretty well with completion as shown in the video below.

https://github.com/user-attachments/assets/81901cf8-f37d-49be-a2a8-b7f5b8480097

## Tasks
- [x] Documentation has been updated (N/A)

### DIFF
--- a/src/sourcekit-lsp/LanguageClientConfiguration.ts
+++ b/src/sourcekit-lsp/LanguageClientConfiguration.ts
@@ -176,6 +176,33 @@ export class LanguagerClientDocumentSelectors {
     }
 }
 
+function addParameterHintsCommandsIfNeeded(
+    items: vscode.CompletionItem[],
+    documentUri: vscode.Uri
+): vscode.CompletionItem[] {
+    if (!configuration.parameterHintsEnabled(documentUri)) {
+        return items;
+    }
+
+    return items.map(item => {
+        switch (item.kind) {
+            case vscode.CompletionItemKind.Function:
+            case vscode.CompletionItemKind.Method:
+            case vscode.CompletionItemKind.Constructor:
+            case vscode.CompletionItemKind.EnumMember:
+                return {
+                    command: {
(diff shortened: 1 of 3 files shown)


src/sourcekit-lsp/LanguageClientConfiguration.ts
```diff
@@ -176,6 +176,33 @@ export class LanguagerClientDocumentSelectors {
...
+            case vscode.CompletionItemKind.Function:
+            case vscode.CompletionItemKind.Method:
+            case vscode.CompletionItemKind.Constructor:
+            case vscode.CompletionItemKind.EnumMember:
+                return {
+                    command: {
+                        title: "Trigger Parameter Hints",
+                        command: "editor.action.triggerParameterHints",
+                    },
+                    ...item,
```
plemarquand, line 198: nit: I'd like to confirm an existing `command` should be used when one is already defined on `item`.

a7medev: Yeah, if we ever have a `command` on an item we keep that rather than overriding it with triggering signature help. This shouldn't matter as of now since SourceKit-LSP already doesn't set commands for completion items.

PR #1824, opened 2025-09-04
Fix "Run Tests Multiple Times" on Linux

plemarquand:
## Description
When we process the `...args` passed to the 'run multiple' commands we're only checking if the argument is an object before deeming it a TestItem.

On linux VS Code is passing an event object like `{ preserveFocus: true }`, which is being treated as a TestItem and causing the command to error out.

## Tasks
- [x] ~Documentation has been updated~

### DIFF
--- a/src/commands/testMultipleTimes.ts
+++ b/src/commands/testMultipleTimes.ts
@@ -128,7 +128,10 @@ export function extractTestItemsAndCount(
                 result.count = arg ?? undefined;
                 return result;
             } else if (typeof arg === "object") {
-                if (arg.hasOwnProperty("id") && arg.hasOwnProperty("uri")) {
+                if (
+                    Object.prototype.hasOwnProperty.call(arg, "id") &&
+                    Object.prototype.hasOwnProperty.call(arg, "uri")
+                ) {
                     result.testItems.push(arg);
                 }
                 return result;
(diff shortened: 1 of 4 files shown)


src/commands/testMultipleTimes.ts
```diff
@@ -127,7 +128,12 @@ export function extractTestItemsAndCount(
                 result.count = arg ?? undefined;
                 return result;
             } else if (typeof arg === "object") {
-                result.testItems.push(arg);
+                if (
```
matthewbastien: **nit:** Would it make sense to also update the function signature to accept `...args: unknown[]` just to make it more clear what's going on here? Could even pull out this check into a utility function that we could use later. For example:

```typescript
function isVSCodeTestItem(obj: unknown): obj is vscode.TestItem {
    return typeof obj === "object"
        && Object.prototype.hasOwnProperty.call(arg, "id")
        && Object.prototype.hasOwnProperty.call(arg, "uri");
}

export function extractTestItemsAndCount(...args: unknown[]): {
    // ...
    if (isVSCodeTestItem(arg)) {
        result.testItems.push(arg);
    }
    // ...
}
```

matthewbastien (approved):
Looks good! Just one nitpick.

Issue #1830, opened 2025-09-10
Improve logic for  launch configuration executable path


bkhouri:
Looking at the code base, the extension is making an assumption as to the location of the executables targets location.

See https://github.com/swiftlang/vscode-swift/blob/96192cb6451cae879a929818c458dfd48c524f24/src/debugger/launch.ts#L133-L137

This will fail when the default build system for `swift build` changes.  It would be better to use the `--show-bin-path` command line option passed to `swift build` to determine the build output location.

```
❯ swift build --show-bin-path
/Users/bkhouri/Documents/git/public/swiftlang/swift-package-manager-2/.build/arm64-apple-macosx/debug

❯ swift build --show-bin-path --configuration release
/Users/bkhouri/Documents/git/public/swiftlang/swift-package-manager-2/.build/arm64-apple-macosx/release

❯ swift build --show-bin-path --build-system swiftbuild
/Users/bkhouri/Documents/git/public/swiftlang/swift-package-manager-2/.build/arm64-apple-macosx/Products/Debug

❯ swift build --show-bin-path --build-system swiftbuild  --configuration release
/Users/bkhouri/Documents/git/public/swiftlang/swift-package-manager-2/.build/arm64-apple-macosx/Products/Release


❯ swift build --show-bin-path --configuration release --scratch-path /my-temp
/my-temp/arm64-apple-macosx/release

❯ swift build --show-bin-path --configuration release --scratch-path /my-temp --triple my-triple --build-system swiftbuild
/my-temp/my-triple/Products/Release
```

award999:
Verification fails
```
"swift.buildArguments": [
  "--build-system",
  "swiftbuild"
]
```
LLDB gets ["/Users/award999/testing/MyExecutable/.build/arm64-apple-macosx/Products/Debug/MyCLI", "--input", "hello"]

award999:
Oh I misspoke, that is right path, the issue was it ignores args in the preLaunchTask
```
{
			"type": "swift",
			"args": [
				"build",
				"--product",
				"MyCLI",
				"--build-system",
				"swiftbuild"
			],
			"cwd": "/Users/award999/testing/MyExecutable",
			"problemMatcher": [],
			"label": "swift: Build Debug MyCLI",
			"detail": "swift build --product MyCLI"
		},
```

bkhouri:
We also need to support the case where "swift build" command will eventually use SwiftBuild as the default build system.

award999:
[...]

PR #1831, opened 2025-09-11
Use `swift build --show-bin-path` to compute paths

plemarquand:
## Description
Several paths are hard coded to use the `debug` folder inside `.build`. While this works for the native build system, the swiftbuild build system stores build artifacts in a different set of folders.

Use `--show-bin-path` to determine where `swift build` is going to put the binaries and rely on that instead of hard coding. Ensure that any additional user arguments that affect the bin path are passed to the command so the true bin path is calculated.

Issue: #1830

## Tasks

### DIFF
--- a/src/toolchain/BuildFlags.ts
+++ b/src/toolchain/BuildFlags.ts
@@ -248,21 +248,20 @@ export class BuildFlags {
             return BuildFlags.buildPathCache.get(cacheKey)!;
         }
 
+        // Filters down build arguments to those affecting the bin path
+        const binPathAffectingArgs = (args: string[]) =>
+            BuildFlags.filterArguments(args, [
+                { argument: "--scratch-path", include: 1 },
+                { argument: "--build-system", include: 1 },
+            ]);
+
+        const baseArgs = ["build", "--show-bin-path", "--configuration", buildConfiguration];
+        const fullArgs = [
+            ...this.withAdditionalFlags(baseArgs),
+            ...binPathAffectingArgs(configuration.buildArguments),
+        ];
+
         try {
-            // Filters down build arguments to those affecting the bin path
-            const binPathAffectingArgs = (args: string[]) => {
-                return BuildFlags.filterArguments(args, [
-                    { argument: "--scratch-path", include: 1 },
-                    { argument: "--build-system", include: 1 },
-                ]);
-            };
-
-            const baseArgs = ["build", "--show-bin-path", "--configuration", buildConfiguration];
-            const fullArgs = [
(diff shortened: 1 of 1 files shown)


src/toolchain/BuildFlags.ts
```diff
@@ -221,6 +225,71 @@ export class BuildFlags {
...
+
+            // Execute swift build --show-bin-path
+            const result = await execSwift(fullArgs, this.toolchain, { cwd });
+            const binPath = result.stdout.trim();
+
+            // Cache the result
+            BuildFlags.buildPathCache.set(cacheKey, binPath);
+            return binPath;
+        } catch (error) {
+            logger.warn(
```
award999, line 273: can we log the swift command/arguments that was executed

plemarquand: Good call, done

award999: Still in draft? LGTM otherwise :)

src/debugger/launch.ts
```diff
@@ -121,26 +121,44 @@ export async function makeDebugConfigurations(
...
-    // Users could be on different platforms with different path annotations,
-    // so normalize before we compare.
-    const launchConfig = launchConfigs.find(
-        config => path.normalize(config.program) === path.normalize(targetPath)
-    );
-    return launchConfig;
+
+    try {
+        // Use dynamic path resolution with --show-bin-path
+        const binPath = await folderCtx.toolchain.buildFlags.getBuildBinaryPath(
```
matthewbastien, line 136: **question:** Would it be possible to do this at launch? We've got logic in there already for setting the runtime path for LLDB DAP as an example.

Though we would probably have to change our launch configs to point at a target rather than a program in this case.

plemarquand: I'd like to move away from doing things at launch if possible to streamline activation.

award999: doesn't do a build here so should be like a second? If that's the case then at debug time is probably on but if significantly longer may be worth doing on startup so more responsive when the user tries to debug

plemarquand: The command itself should be pretty quick. Longer term to improve activation times I want to move as much initialization off the 'main thread' on startup, which would prevent it from blocking activation and allow calls later to normally have the work already done. That would be a different PR though.

matthewbastien: By "at launch" I meant launching a debug session.

plemarquand: Oh, yeah sorry, we certainly could. I'd still do that in a separate PR though.

Issue #1852, opened 2025-09-24
Handle Swiftly toolchain not installed error


matthewbastien:
**Is your feature request related to a problem? Please describe.**
When Swiftly does not have an appropriate toolchain to use with the version provided by a `.swift-version` file, it will fail with an error message:
```
The swift version file `<redacted>/Documents/swiftlang/swiftly/.swift-version` uses toolchain version 6.1.2, but it doesn't match any of the installed toolchains. You can install the toolchain with `swiftly install`.
```

When this happens, the Swift extension will simply bail and display the error message. I have to manually fix this by running `swiftly install` in my terminal and then restart VS Code.

**Describe the solution you'd like**
The Swift extension should detect this failure condition and automatically run the Swiftly command to get me up and running. It should also ask before doing so.

matthewbastien:
Verified with 6c9733e1

PR #1855, opened 2025-09-25
Prompt to automatically install missing toolchain

plemarquand:
## Description
If a .swift-version exists in a project that specifies a toolchain the user doesn't have installed, prompt them to install it on extension launch.

If they cancel the dialog, fall back to using whatever toolchain is already installed, basically ignoring the .swift-version for this activation or until the .swift-version is changed again.

Issue: #1852

## Tasks

### DIFF
--- a/src/ui/ToolchainSelection.ts
+++ b/src/ui/ToolchainSelection.ts
@@ -100,6 +100,25 @@ export async function showToolchainError(folder?: vscode.Uri): Promise<boolean> 
     return false;
 }
 
+/**
+ * Shows a dialog asking user permission to install a missing Swiftly toolchain
+ * @param version The toolchain version to install
+ * @param folder Optional folder context for the error
+ * @returns Promise<boolean> true if user agrees to install, false otherwise
+ */
+export async function showMissingToolchainDialog(
+    version: string,
+    folder?: vscode.Uri
+): Promise<boolean> {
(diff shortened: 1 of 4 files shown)


src/ui/ToolchainSelection.ts
```diff
@@ -100,6 +100,25 @@ export async function showToolchainError(folder?: vscode.Uri): Promise<boolean>
...
+export async function showMissingToolchainDialog(
+    version: string,
+    folder?: vscode.Uri
+): Promise<boolean> {
+    const folderName = folder ? `${FolderContext.uriName(folder)}: ` : "";
+    const message =
+        `${folderName}Swift version ${version} is required but not installed. ` +
+        `Would you like to automatically install it using Swiftly?`;
+
+    const choice = await vscode.window.showWarningMessage(message, "Install Toolchain", "Cancel");
```
matthewbastien, line 118: Given that this blocks the whole extension from being useful, I think we should make it a modal.

plemarquand: Potentially you can still use the extension with whatever toolchain you currently have, especially if you have a version greater than whatever is required (i.e. you have 6.2 and .swift-version is 6.1).

matthewbastien: Actually yeah that makes sense. Probably better to be less invasive anyway.

PR #1857, opened 2025-09-25
Allow optional `swift.backgroundCompilation` object notation for finer control of background compilation

award999:
## Description
This is not related to background indexing, but instead the optional background compilation feature that runs a build task on save. Allow `swift.backgroundCompilation` setting to accept an object which will accept `enabled`, `release`, and `useDefaultTask` properties. The task marked as the default "build" group task will be used instead of the Build All task. This way the whole world is not built for large projects.

Issue: #1320

## Tasks

### DIFF
--- a/test/integration-tests/BackgroundCompilation.test.ts
+++ b/test/integration-tests/BackgroundCompilation.test.ts
@@ -17,7 +17,7 @@ import * as vscode from "vscode";
 import { BackgroundCompilation } from "@src/BackgroundCompilation";
 import { FolderContext } from "@src/FolderContext";
 import { WorkspaceContext } from "@src/WorkspaceContext";
-import { getBuildAllTask } from "@src/tasks/SwiftTaskProvider";
+import { createSwiftTask, getBuildAllTask } from "@src/tasks/SwiftTaskProvider";
 
 import { mockGlobalObject } from "../MockUtils";
 import { testAssetUri } from "../fixtures";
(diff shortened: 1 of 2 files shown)


test/integration-tests/BackgroundCompilation.test.ts
```diff
@@ -62,4 +75,65 @@ tag("large").suite("BackgroundCompilation Test Suite", () => {
         await taskStartPromise;
         await waitForNoRunningTasks();
     });
+
+    suite("getTask", () => {
+        const tasksMock = mockGlobalObject(vscode, "tasks");
+        let swiftTask: vscode.Task;
+        let nonSwiftTask: vscode.Task;
+        let backgroundConfiguration: BackgroundCompilation;
+        const useDefaultTaskConfig = mockGlobalValue(configuration, "useDefaultTask");
```
matthewbastien: **question:** Should we just use the actual VS Code APIs here since it's an integration test? I don't think it would add much runtime to the test itself. Can even use the `updateSettings()` utility method that Paul added a while back which comes with the ability to restore settings after the test finishes.

Might even be able to update the `tasks.json` too instead of mocking, but I'm not sure how easy that would be.

award999: the activation and reset settings does not work in a sub-suite. I'll see how hard it is to drag these to the unit-tests suite

award999: Issue was actually that setting `swift.backgroundCompilation` to `true` was causing the value of `swift.backgroundCompilation.useDefaultTask` to be ignored. The `swift.backgroundCompilation` setting is niche enough that I did not see a reason to migrate to a `swift.backgroundCompilation.enabled` setting

Issue #1848, opened 2025-09-23
Selecting a Swiftly toolchain always sets it globally


matthewbastien:
**Describe the bug**
The toolchain selection workflow asks whether or not I want to install the toolchain in User settings or Workspace settings. However, it always updates the global Swiftly toolchain regardless of which option I pick.

**To Reproduce**
1. Run the `Swift: Select Toolchain...` command.
2. When prompted, select any toolchain installed via Swiftly.
3. When prompted, select any Xcode toolchain for the `DEVELOPER_DIR`.
4. When prompted, select `Workspace Configuration` for the location to install the toolchain.

The `DEVELOPER_DIR` gets set correctly in workspace settings. However, only the Swiftly global toolchain is updated.

**Expected behavior**
The `.swift-version` file should be created/modified in my workspace to match the version I selected in the dialog.

matthewbastien:
Completed with #1860.

award999:
Verification failed. The `.swift-version` file gets modified if it does exist, but does not get created if it does not exist

award999:
Also a question @matthewbastien, we're purposely not setting the `swift.path` setting? i.e. if there is a .swift-version file then we assume we should use swiftly toolchains?

award999:
Additionally when installing a new toolchain, it does not create nor update the .swift-version file

matthewbastien:
Yeah, when the user selects a Swiftly toolchain from the toolchain selection quick pick we do purposefully remove any `swift.path` setting.

award999:
Verified

PR #1860, opened 2025-09-26
Allow setting the Swiftly toolchain at the workspace level

matthewbastien:
## Description
Respect the user's configuration selection when invoking `swiftly use`. If the user selects the workspace configuration then the toolchain will be updated via the `.swift-version` file in the workspace (and if one isn't there it will be created). For global toolchains the `--global-default` option will be set so that the toolchain gets updated globally.

Issue: #1848

## Tasks

Swiftly is a new feature that hasn't made it into a release yet. No need to update change log.

### DIFF
--- a/src/ui/ToolchainSelection.ts
+++ b/src/ui/ToolchainSelection.ts
@@ -25,7 +25,7 @@ import { showReloadExtensionNotification } from "./ReloadExtension";
 /**
  * Open the installation page on Swift.org
  */
-export async function downloadToolchain() {
+async function downloadToolchain() {
     if (await vscode.env.openExternal(vscode.Uri.parse("https://www.swift.org/install"))) {
         const selected = await showReloadExtensionNotification(
             "The Swift extension must be reloaded once you have downloaded and installed the new toolchain.",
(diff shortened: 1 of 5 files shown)


src/ui/ToolchainSelection.ts
```diff
@@ -555,7 +562,9 @@ async function setToolchainPath(
         target
     );
     await checkAndRemoveWorkspaceSetting(target);
-    return true;
```
award999, line 558: question: some tests check the boolean return value to determine success/failure. Was the return value being used somewhere?

matthewbastien: The return value was being used to determine if `toolchainItem.onDidSelect()` needed to be called, but I just moved that into `setToolchainPath()` to keep things simple.

The toolchain selection tests need to be re-done, honestly. They don't really check anything valuable right now. I'm going to rework them today as part of this PR to make sure everything is good to go.

src/ui/ToolchainSelection.ts
```diff
@@ -225,22 +234,29 @@ async function getQuickPickItems(
...
-    // Sort toolchains by label (alphabetically)
-    const sortedToolchains = toolchains.sort((a, b) => b.label.localeCompare(a.label));
+        });
 
     // Find any Swift toolchains installed via Swiftly
-    const swiftlyToolchains = (await Swiftly.list(logger)).map<SwiftlyToolchainItem>(
-        toolchainPath => ({
+    const swiftlyToolchains = (await Swiftly.list(logger))
+        // Sort in descending order alphabetically
+        .sort((a, b) => -a.localeCompare(b))
```
award999, line 242: wouldn't that put 5.10 before 5.9?

award999 (approved):
Can fix sorting in next PR

matthewbastien:
Unrelated test failure. Merging.

PR #1865, opened 2025-09-29
Ensure the extension activates with a .bsp folder

plemarquand:
## Description
A .bsp folder was considered a valid workspace folder, but the extension activation events in the package.json were not configured to search for it, meaning another valid file/folder for activation had to be present. If that was the case, then the .bsp folder would be discovered correctly, but not if it was the only file/folder in the folder that would activate the extension.

Add it to the list of valid activation file types. Also clean up this code path a bit, ignoring common folders we shouldn't search for projects.

## Tasks

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "workspaceContains:**/compile_commands.json",
     "workspaceContains:**/compile_flags.txt",
     "workspaceContains:**/buildServer.json",
+    "workspaceContains:**/.bsp/*.json",
     "onDebugResolve:swift-lldb",
     "onDebugResolve:swift"
   ],
--- a/src/utilities/workspace.ts
+++ b/src/utilities/workspace.ts
@@ -16,7 +16,7 @@ import * as path from "path";
 import { basename } from "path";
 import * as vscode from "vscode";
 
-import { globDirectory, pathExists } from "./filesystem";
+import { folderExists, globDirectory, pathExists } from "./filesystem";
 import { Version } from "./version";
 
 export async function searchForPackages(
@@ -28,7 +28,7 @@ export async function searchForPackages(
     const folders: Array<vscode.Uri> = [];
 
     async function search(folder: vscode.Uri) {
-        // add folder if Package.swift/compile_commands.json/compile_flags.txt/buildServer.json exists
+        // add folder if Package.swift/compile_commands.json/compile_flags.txt/buildServer.json/.bsp exists
         if (await isValidWorkspaceFolder(folder.fsPath, disableSwiftPMIntegration, swiftVersion)) {
             folders.push(folder);
         }
@@ -38,8 +38,18 @@ export async function searchForPackages(
         }
 
         await globDirectory(folder, { onlyDirectories: true }).then(async entries => {
+            const skipFolders = new Set<string>([
+                ".",
+                ".build",
+                "Packages",
+                "out",
+                "bazel-out",
+                "bazel-bin",
+            ]);
+
             for (const entry of entries) {
-                if (basename(entry) !== "." && basename(entry) !== "Packages") {
+                const base = basename(entry);
+                if (!skipFolders.has(base)) {
                     await search(vscode.Uri.file(entry));
                 }
             }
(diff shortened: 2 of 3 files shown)


plemarquand:
This can be verified with a project like https://github.com/spotify/sourcekit-bazel-bsp/tree/main/Example

plemarquand:
This can be verified with a project like https://github.com/spotify/sourcekit-bazel-bsp/tree/main/Example

src/utilities/workspace.ts
```diff
@@ -38,8 +38,18 @@ export async function searchForPackages(
         }
 
         await globDirectory(folder, { onlyDirectories: true }).then(async entries => {
+            const skipFolders = new Set<string>([
```
award999: probably should make this list a setting to allow excluding others and maybe user may want to add one of these folder. I'm thinking maybe bazel could generate a package for them possibly...

src/utilities/workspace.ts
rbenegal, line 38: I was trying this out and looks like the `globDirectory` logic will only work if "Search subfolders for packages" setting is enabled? It seems to be off by default I think so wonder if we should turn it on by default since folks may not know this option exists?

src/utilities/workspace.ts
```diff
@@ -38,8 +38,18 @@ export async function searchForPackages(
...
+                ".",
+                ".build",
+                "Packages",
```
rbenegal: Curious to know about why the `Packages` directory is skipped, could there not be valid swift files in there that folks would like extension support for?

plemarquand: I'm not totally sure, but this is a port of the old code on line 42 so I figured it should stay

src/utilities/workspace.ts
plemarquand: I agree, I'll make the change

src/utilities/workspace.ts
plemarquand: I'm split on the default here. Often times packages will have other packages as test fixtures, and having this setting default to `true` wouldn't be what you want in that case (see Swift Package Manager, Sourcekit-LSP).

However if you're doing local package development where your manifest points to other folders in your package, you may want it on. 

Both of these are somewhat uncommon. Having the extension start indexing a bunch of folders unexpectedly is a worse experience, so I'm leaning towards leaving this as `false` by default.

rbenegal: Good points thanks. As a new user though, I would have no idea that this setting exists. Maybe there needs to be notification along the lines of "detected packages in sub folders, or additional configuration files, would you like to index those too?", but I think that's something to be discussed in another issue.

src/utilities/workspace.ts
rbenegal: Thanks, yeah maybe having this as a setting will give people a choice to include or ignore. Possibly it just took too long to index Packages in the past so it was just skipped?

package.json
```diff
@@ -544,6 +545,22 @@
...
+            },
+            "default": [
+              ".",
+              ".build",
+              "Packages",
+              "out",
+              "bazel-out",
+              "bazel-bin"
+            ],
+            "markdownDescription": "A list of glob patterns to ignore when searching sub-folders for Swift Packages. The `swift.searchSubfoldersForPackages` must be `true` for this setting to have an effect. Always use forward-slashes in glob expressions regardless of platform. This is combined with VS Code's default `files.exclude` setting.",
```
rbenegal: Is this actually a glob pattern though? Seems like we just look at the basename of the folder and then skip if it matches something in this list?

Also wondering if the "default" could be removed from "This is combined with VS Code's default `files.exclude` setting.", and just say "This is combined with VS Code's `files.exclude` setting." to make it clearer?

award999:
```suggestion
            "markdownDescription": "A list of glob patterns to ignore when searching sub-folders for Swift Packages. The `swift.searchSubfoldersForPackages` must be `true` for this setting to have an effect. Always use forward-slashes in glob expressions regardless of platform. This is combined with VS Code's `files.exclude` setting.",
```

award999 (approved):
Looks good to merge

plemarquand:
@award999 done

award999:
Verified with 5aa67833513eef4994c86a59c50061d35bd9a569

PR #1869, opened 2025-09-30
Expand test logging during extension activation/deactivation

plemarquand:
## Description
Provide more robust logs during extension activation/deactivation in order to help track down test timeouts. These logs are only printed if a test fails to avoid cluttering up the test report.

This additional logging uncovered several test related issues:
- A background compilation task could have started after the BackgroundCompilation object was disposed
- TaskQueues were not disposed and would continue to process tasks after the parent TaskManager was disposed
- DYLD_FRAMEWORK_PATH/DYLD_LIBRARY_PATH should be set in the environment when debugging test both kinds of test executables on >=6.2
- The FolderContexts created in FolderContext.test.ts were not being disposed, causing their BackgroundCompilation to run at undesirable times
- The LanguageClientIntegration suite setup was doing a build all with test discovery enabled. Because the LSP isn't yet configured at this point, the extension would fall back to trying to discover tests with `swift test --list-tests` unexpectedly.
- The ProjectPanelProvider did not dispose its listeners, and was not always being disposed correctly in the tests.

## Tasks
****

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -127,7 +127,8 @@ export class SwiftToolchain {
         );
         const toolchainPath = await this.getToolchainPath(swiftFolderPath, folder, logger);
         const targetInfo = await this.getSwiftTargetInfo(
-            this._getToolchainExecutable(toolchainPath, "swift")
+            this._getToolchainExecutable(toolchainPath, "swift"),
+            logger
         );
         const swiftVersion = this.getSwiftVersion(targetInfo);
         const [runtimePath, defaultSDK] = await Promise.all([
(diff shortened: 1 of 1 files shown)


src/debugger/buildConfig.ts
```diff
@@ -450,6 +451,16 @@ export class TestingConfigurationFactory {
                             env: {
                                 ...this.testEnv,
                                 ...this.sanitizerRuntimeEnvironment,
+                                ...(toolchain.swiftVersion.isGreaterThanOrEqual(
+                                    new Version(6, 2, 0)
```
award999, line 455: is this related to this PR?

plemarquand: It could be broken out, but I'm trying to fix a bunch of test issues here as I encounter them. Its this line item in the description:
- DYLD_FRAMEWORK_PATH/DYLD_LIBRARY_PATH should be set in the environment when debugging test both kinds of test executables on >=6.2

award999 (approved):

PR #1875, opened 2025-10-06
Ensure that the .swift-version file exists before calling `swiftly use`

matthewbastien:
## Description
Some instances of Swiftly do not actually create the `.swift-version` file when `swiftly use` is called. They will only modify an existing `.swift-version` file. This is definitely an issue with Swiftly itself, but we can at least mitigate the problem in the Swift extension until we can determine a proper fix.

## Tasks


award999 (approved):
LGTM

matthewbastien:
Unrelated test failure. Merging.

PR #1881, opened 2025-10-07
Add Swiftly installation support

roulpriya:
## Description

This PR adds the ability for the Swift extension to install the swiftly toolchain manager if it is not already present on the user's machine. If a `.swift-version` file is present, but swiftly is not detected, a warning message will be shown to the user telling them that they can use swiftly to manage their toolchains.

The Swift extension will download swiftly, install it, and initialize it with the default environment. It will then install all of the toolchains that it found in the `.swift-version` file(s). Unfortunately, we also need to ask the user to restart VS Code in order to pick up environment changes.

Issue: https://github.com/swiftlang/vscode-swift/issues/2025

## Tasks

### DIFF
--- a/src/commands/installSwiftly.ts
+++ b/src/commands/installSwiftly.ts
@@ -23,7 +23,7 @@ import { installSwiftlyToolchainWithProgress } from "./installSwiftlyToolchain";
 /**
  * Prompts user for Swiftly installation with directory customization options
  * @param logger Optional logger
- * @returns Promise<SwiftlyInstallOptions | null> Installation options if user wants to install, null otherwise
+ * @returns A promise that resolves to true if the user has opted to install swiftly, false otherwise
  */
 export async function promptForSwiftlyInstallation(logger?: SwiftLogger): Promise<boolean> {
     const installMessage = `A .swift-version file was detected. Install Swiftly to automatically manage Swift toolchain versions for this project.`;


package.json
```diff
@@ -1122,6 +1122,13 @@
             "order": 4,
             "scope": "machine-overridable"
           },
+          "swift.suppressSwiftlyInstallPrompt": {
```
matthewbastien: **suggestion:** Just to keep this consistent with other setting names:
```suggestion
          "swift.disableSwiftlyInstallPrompt": {
```

src/extension.ts
```diff
@@ -288,11 +289,63 @@ function handleFolderEvent(logger: SwiftLogger): (event: FolderEvent) => Promise
     };
 }
 
+/**
+ * Checks if any workspace folder contains a .swift-version file
+ */
+async function hasSwiftVersionFile(): Promise<boolean> {
```
matthewbastien: **suggestion:** We should probably also check the `FolderContext` directories since a user can open a top level folder with multiple sub-folders that have a `Package.swift`. Maybe even just make this check happen at the `FolderContext` or `WorkspaceContext` level?

src/extension.ts
```diff
@@ -288,11 +289,63 @@ function handleFolderEvent(logger: SwiftLogger): (event: FolderEvent) => Promise
...
+}
+
 async function createActiveToolchain(
     extension: vscode.ExtensionContext,
     contextKeys: ContextKeys,
     logger: SwiftLogger
 ): Promise<SwiftToolchain | undefined> {
+    // Check if there's a .swift-version file in the workspace and Swiftly is not installed
+    if (await hasSwiftVersionFile()) {
+        const { Swiftly } = await import("./toolchain/swiftly");
```
matthewbastien: **issue:** Keep the imports at the top level. Makes it easier to see where the dependencies are.

test/unit-tests/toolchain/swiftly.test.ts
```diff
@@ -1494,4 +1494,611 @@ apt-get -y install libncurses5-dev
             expect(Swiftly.cancellationMessage).to.equal("Installation cancelled by user");
         });
     });
+
+    suite("installSwiftly()", () => {
```
matthewbastien, line 1513: **issue:** These tests appear to actually be attempting to decompress artifacts judging by some of the error messages in CI. You'll have to find some way to abstract away `tar` with dependency injection. It honestly might be easier to do an integration test for the install. However, performing a `swiftly init` is going to mess with environment variables which could pose problems.

A bunch of these tests are failing in CI which will need to be fixed up.

matthewbastien (changes requested):
Looking good so far, thank you! I have several comments from my first look at this.

You'll also have to rebase on top of `main` before adding to the changelog.

src/extension.ts
roulpriya: Fixed

src/extension.ts
roulpriya: Should we limit to just checking 1 or 2 levels?

package.json
roulpriya: Done

src/extension.ts
matthewbastien: Oh, I see this is happening before the creation of the `WorkspaceContext` in which case we don't have any `FolderContext` yet. Didn't realize that the first time around.

What I meant to say was: we should also perform this check when we initialize a toolchain for a `FolderContext`. However unlikely, it is possible that there exists a `.swift-version` file in a sub-folder, but not at the top level of the workspace.

src/commands/installSwiftly.ts
```diff
@@ -0,0 +1,196 @@
...
+        "Customize Directories",
+        "Don't Show Again",
+        "Cancel"
+    );
+
+    switch (selection) {
+        case "Install Swiftly":
+            return {}; // Use defaults
+
+        case "Customize Directories":
```
matthewbastien: **issue:** I was trying this out locally and was really confused by the `Customize Directories` button. It didn't seem like it was actually going to install Swiftly, but it does after asking me to select a directory.

I think it makes more sense from a user perspective to only show the buttons `Install Swiftly` and `Don't Show Again`. The directory customization can be shown as part of the installation workflow if the user wants to customize it. The `Cancel` button was just sort of cluttering the UI since there's already a close button on the notification.

src/commands/installSwiftly.ts
```diff
@@ -0,0 +1,196 @@
...
+ * Prompts user for Swiftly installation with directory customization options
+ * @param logger Optional logger
+ * @returns Promise<SwiftlyInstallOptions | null> Installation options if user wants to install, null otherwise
+ */
+export async function promptForSwiftlyInstallation(
+    logger?: SwiftLogger
+): Promise<SwiftlyInstallOptions | null> {
+    const installMessage = `A .swift-version file was detected. Install Swiftly to automatically manage Swift toolchain versions for this project.`;
+
+    const selection = await vscode.window.showInformationMessage(
```
matthewbastien: **nit:** I think this would make more sense as a warning message. If there's a `.swift-version` file in the workspace then my guess would be that the project wants people to use swiftly.

src/toolchain/swiftly.ts
```diff
@@ -182,6 +183,237 @@ export async function handleMissingSwiftlyToolchain(
...
+                env["SWIFTLY_HOME_DIR"] = swiftlyHomeDir;
+            }
+            if (swiftlyBinDir) {
+                env["SWIFTLY_BIN_DIR"] = swiftlyBinDir;
+            }
+
+            const actualSwiftlyHomeDir = swiftlyHomeDir || path.join(os.homedir(), ".swiftly");
+            const swiftlyPath = path.join(actualSwiftlyHomeDir, "bin", "swiftly");
+
+            await execFile(swiftlyPath, ["init", "--quiet-shell-followup"], { env });
```
matthewbastien: **issue:** This hangs for me because Swiftly asks whether or not to modify the profile. The extension should be prompting for this before the install and then set the `-y` or `-n` flag appropriately. Same thing for Linux below.

src/commands/installSwiftly.ts
roulpriya: Ohh right. Sure I will remove the cancel button and when user clicks "Install Swiftly", a second prompt asks: "Do you want to customize the installation directories?" with options "Use Defaults" and "Customize Directories"?

Let me know if this looks fine.

src/toolchain/swiftly.ts
roulpriya: Should I be adding a user prompt in the installation workflow and passing the flag?

matthewbastien: Yeah that's probably the best way to go.

src/commands/installSwiftly.ts
matthewbastien: I'll do another review, but that sounds about what I wanted.

src/commands/installSwiftly.ts
```diff
@@ -0,0 +1,157 @@
...
+
+import { SwiftLogger } from "../logging/SwiftLogger";
+import { Swiftly } from "../toolchain/swiftly";
+import { Workbench } from "../utilities/commands";
+import { installSwiftlyToolchainWithProgress } from "./installSwiftlyToolchain";
+
+/**
+ * Prompts user for Swiftly installation with directory customization options
+ * @param logger Optional logger
+ * @returns Promise<SwiftlyInstallOptions | null> Installation options if user wants to install, null otherwise
```
plemarquand: nit: this return type in the docs doesn't match the real return type

PR #1885, opened 2025-10-11
Add .npmrc with default registry configuration

plemarquand:
## Description
This is useful when a users default registry is not the public one

## Tasks


plemarquand:
Must have hit that button by accident. Thanks for your detailed review, copilot.

Issue #1889, opened 2025-10-14
Auto generated launch configs are no longer shareable


matthewbastien:
**Describe the bug**
In the latest pre-release of the Swift extension (v2.11.20251006), I was prompted to update my Swift launch configurations. Upon clicking the `Update` button, it expanded all of the `"${workspaceFolder}"` variables in the "program" property to a full system path. Also, the target triple appears within this path. This makes it impossible to commit the `launch.json` to source control.

**Environment**
 - OS: macOS Tahoe
 - vscode-swift version: 2.11.20251006

matthewbastien:
Fixed by #1890.

The automatic launch config generator will now set the `target` and `configuration` properties for Swift launch configurations. The binary path will be found when launching the debug session. This allows them to be reused on any platform

matthewbastien:
Verification requirements:
1. The Swift extension should prompt to change launch configurations on startup. Clicking update should replace the `program` property in all Swift launch configurations with `target` and `configuration` properties.
2. Launching a debug session with a `target` and `configuration` property should debug the correct binary.
3. Launching a debug session with a `program` property should debug that binary.
4. Launching a debug session with both a `program` and `target` property should fail.
5. Launching a debug session with a `configuration` property that is not `release` or `debug` should fail.

award999:
Verified

PR #1890, opened 2025-10-14
Add a "target" property to launch configurations

matthewbastien:
## Description
With the new changes to use the `--show-bin-path` option from SwiftPM, the target triple is now present in the `"program"` for auto generated launch configurations. This makes it impossible to share the configuration between different platforms.

I've added new `"target"` and `"configuration"` properties to all Swift launch configurations that can be used instead of the `"program"` to debug a target from the `Package.swift`. All auto-generated launch configs will use these new properties going forward, but the `"program"` property can still be used to support cmake and other non-SwiftPM build systems.

Issue: #1889

## Tasks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Prompt to cancel and replace the active test run if one is in flight ([#1774](https://github.com/swiftlang/vscode-swift/pull/1774))
 - A walkthrough for first time extension users ([#1560](https://github.com/swiftlang/vscode-swift/issues/1560))
 - Allow `swift.backgroundCompilation` setting to accept an object where enabling the `useDefaultTask` property will run the default build task, and the `release` property will run the `release` variant of the Build All task ([#1857](https://github.com/swiftlang/vscode-swift/pull/1857))
-- Added a new `target` property to `swift` launch configurations that can be used instead of `program` for SwiftPM based projects ([#1890](https://github.com/swiftlang/vscode-swift/pull/1890))
+- Added new `target` and `configuration` properties to `swift` launch configurations that can be used instead of `program` for SwiftPM based projects ([#1890](https://github.com/swiftlang/vscode-swift/pull/1890))
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


src/debugger/debugAdapterFactory.ts
```diff
@@ -94,10 +95,22 @@ export class LLDBDebugConfigurationProvider implements vscode.DebugConfiguration
...
+
+        // Convert the "target" property to a "program"
+        if (typeof launchConfig.target === "string") {
+            const targetName = launchConfig.target;
+            if (!folderContext) {
+                throw new Error(
+                    `Unable to resolve target "${targetName}". No Swift package is available to search within.`
+                );
+            }
+            launchConfig.program = await getTargetBinaryPath(targetName, folderContext);
```
plemarquand: What do we do if there is already a program path?

package.json
```diff
@@ -1725,14 +1725,15 @@
         },
         "configurationAttributes": {
           "launch": {
-            "required": [
```
plemarquand, line 1728: I'm not sure if this supports the same syntax as the settings, but it would be good if we could mark `oneOf` `target` or `program` as required

src/debugger/debugAdapterFactory.ts
matthewbastien: If there's a program path then the logic behaves the same way as before. If both a target and program path exist then the target takes precedence.

package.json
matthewbastien: Doesn't seem like it. There's nothing mentioning it in the docs and my attempts to set it haven't worked.

I updated the logic to also check to make sure that either a "target" or "program" are present. If not, the user is shown an error that they need to update their debug configuraiton.

award999: will explicit "program" take precedence?

award999:
Adding full test run label to make sure there are no surprises in nightly after merging

package.json
matthewbastien: Right now `target` takes precedence, but I'm fine with changing it if we feel that `program` makes more sense. I'm not sure what would be more intuitive there.

There's the third option of just making that an error that the user has to fix.

award999: In my mind if I am explicitly setting the `program` that would take precedence, or fail on the launch config's resolve with an error that says there can only be 1

matthewbastien: I can see arguments for either "target" or "program" taking precedence. So, in the name of clarity I've opted to just show an error asking the user to remove one of them if they're both present.

PR #1895, opened 2025-10-20
Select `DEVELOPER_DIR` when installing a toolchain via Swiftly

matthewbastien:
## Description
Swiftly installation on macOS will now also ask the user to select an Xcode if multiple are installed. This matches the behavior of toolchain selection since we do end up selecting the new toolchain as well during this command.

## Tasks

### DIFF
--- a/src/commands/installSwiftlyToolchain.ts
+++ b/src/commands/installSwiftlyToolchain.ts
@@ -16,7 +16,12 @@ import * as vscode from "vscode";
 import { WorkspaceContext } from "../WorkspaceContext";
 import { SwiftLogger } from "../logging/SwiftLogger";
 import { Swiftly, SwiftlyProgressData } from "../toolchain/swiftly";
-import { askWhereToSetToolchain } from "../ui/ToolchainSelection";
+import { SwiftToolchain } from "../toolchain/toolchain";
+import {
+    askWhereToSetToolchain,
+    setToolchainPath,
+    showDeveloperDirQuickPick,
+} from "../ui/ToolchainSelection";
 
 /**
  * Installs a Swiftly toolchain and shows a progress notification to the user.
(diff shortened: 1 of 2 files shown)


src/commands/installSwiftlyToolchain.ts
```diff
@@ -165,6 +170,12 @@ export async function promptToInstallSwiftlyToolchain(
         return;
     }
 
+    const xcodes = await SwiftToolchain.findXcodeInstalls();
```
award999, line 173: So this would be called from the Install Swiftly Toolchain too, not just the Select Toolchain one?

matthewbastien: That is correct.

award999: Is that what we want? Or maybe question is what is the line of what the Install command should do vs. Select Toolchain command? Why do we have both commands btw? I almost look at it like using the Install command instead of Select would suggest I only want to install the toolchain, but what context would the user want that? And that may signal we may not want both commands....

award999: Does `swiftly install <version>` also make it the default toolchain?

matthewbastien: > Is that what we want? Or maybe question is what is the line of what the Install command should do vs. Select Toolchain command? Why do we have both commands btw? I almost look at it like using the Install command instead of Select would suggest I only want to install the toolchain, but what context would the user want that? And that may signal we may not want both commands....

We have both commands because otherwise the toolchain selection dialog would be way too long. There are a lot of toolchains that Swiftly can install.

I agree that it's a bit weird to have the install also select the toolchain, but I think it's a pretty reasonable assumption when using VS Code that someone would want to immediately use the toolchain that they're installing. Otherwise it'd just be annoying to have to run the install command and then the toolchain selection right after. The install command is also available as an option in toolchain selection. In that instance I would definitely say that we should also select it afterwards.

> Does `swiftly install <version>` also make it the default toolchain?

No it doesn't.

award999: Can leave for now, not apposed to 2 commands but not very clear what expected behaviour is

matthewbastien: Yeah I think this is fine for the upcoming release, but we can discuss further if there are changes we want to make afterwards.

PR #1898, opened 2025-10-21
Support prelaunchTasks that build with swiftbuild

plemarquand:
## Description
If a prelaunch task is configured to build with --build-system swiftbuild we need to respect that when setting the program path of the launch configuration.

Issue: #1830

## Tasks

### DIFF
--- a/src/debugger/launch.ts
+++ b/src/debugger/launch.ts
@@ -379,16 +379,10 @@ export async function swiftPrelaunchBuildTaskArguments(
             return undefined;
         }
 
-        // Check if args contain "build" and "--build-system"
+        // Check if args contain "build"
         const args = (task.definition.args as string[]) || [];
         const hasBuild = args.includes("build");
-        const hasBuildSystem = args.includes("--build-system");
-
-        if (hasBuild && hasBuildSystem) {
-            return args;
-        }
-
-        return undefined;
+        return hasBuild ? args : undefined;
     } catch (error) {
         // Log error but don't throw - return undefined for safety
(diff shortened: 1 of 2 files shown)


src/debugger/launch.ts
```diff
@@ -328,3 +332,65 @@ function updateConfigWithNewKeys(
...
+
+        // Check if task type is "swift"
+        if (task.definition.type !== "swift") {
+            return undefined;
+        }
+
+        // Check if args contain "build" and "--build-system"
+        const args = (task.definition.args as string[]) || [];
+        const hasBuild = args.includes("build");
+        const hasBuildSystem = args.includes("--build-system");
```
jakepetroules: I'm not quite understanding the reason you want to check for `--build-system` here? Why do we want to take a different path for `swift build` alone vs `swift build --build-system X`?

plemarquand: We only need to calculate the bin-path if the build-system is set, as (AFAIK) this is the only setting that controls the bin-path

award999: Couldn't `--scratch-path` change it too?

plemarquand: @award999 you're right. I refactored this to `swiftPrelaunchBuildTaskArguments` so it doesn't care about specific arguments, just that it's a `swift build` invocation. The `getTargetBinaryPath` method already handles passing through `--build-system` and `--scratch-path`, so when we forward along the arguments from `swiftPrelaunchBuildTaskArguments` we'll get the desired result.

PR #1901, opened 2025-10-22
Improve Swiftly installation progress notification

matthewbastien:
## Description
Swiftly performs a signature verification and extraction of the toolchain even after the download completes. Show a notification in this case and reset the progress to indicate to the user that Swiftly is still performing some work.

## Tasks

### DIFF
--- a/src/commands/installSwiftlyToolchain.ts
+++ b/src/commands/installSwiftlyToolchain.ts
@@ -49,30 +49,35 @@ export async function installSwiftlyToolchainWithProgress(
                 await Swiftly.installToolchain(
                     version,
                     (progressData: SwiftlyProgressData) => {
-                        if (
-                            progressData.step?.percent !== undefined &&
-                            progressData.step.percent > lastProgress
-                        ) {
-                            const increment = progressData.step.percent - lastProgress;
+                        if (progressData.complete) {
+                            // Swiftly will also verify the signature and extract the toolchain after the
+                            // "complete" message has been sent, but does not report progress for this.
+                            // Provide a suitable message in this case and reset the progress back to an
+                            // indeterminate state (0) since we don't know how long it will take.
                             progress.report({
-                                increment,
-                                message:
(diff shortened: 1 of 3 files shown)


src/commands/installSwiftlyToolchain.ts
```diff
@@ -49,31 +49,36 @@ export async function installSwiftlyToolchainWithProgress(
                 await Swiftly.installToolchain(
                     version,
                     (progressData: SwiftlyProgressData) => {
-                        if (
-                            progressData.step?.percent !== undefined &&
-                            progressData.step.percent > lastProgress
-                        ) {
-                            const increment = progressData.step.percent - lastProgress;
+                        if (progressData.complete) {
+                            // Swiftly will also verify the signature and extract the toolchain after the
```
award999, line 53: how funky would it be to popup a withProgress notification? That way at least get the progress animation

tsconfig-base.json
```diff
@@ -5,8 +5,8 @@
     "rootDir": ".",
     "outDir": "dist",
 
-    "lib": ["ES2021"],
-    "target": "ES2020",
+    "lib": ["ES2022"],
```
award999, line 8: why do we need to bump for this change?

src/commands/installSwiftlyToolchain.ts
matthewbastien: This is within a withProgress notification. Resetting the progress to 0 resets the animation.

award999 (approved):

tsconfig-base.json
matthewbastien: This is to get error causations. So, you can provide a cause (kind of like in Java) for your errors:
```typescript
try {
   // Do something that will throw
} catch (error) {
   throw new Error("Something went wrong", { cause: error });
}
```

NodeJS will print the stack trace of both errors in this case so that we don't have to do this manually like what was in the existing code.

PR #1913, opened 2025-11-04
Move log activation off the critical extension activation path

plemarquand:
Initializing both the file transport and containing folder for the log files was blocking extension activation, sometimes taking up to 1,500ms.

Move log activation off the critical path of extension activation by avoiding awaiting promises before continuing activation. Any logs logged before the file transport is ready are buffered and replayed into the file transport once file system operations are complete.

This cuts extension activation time down by about 50%.

This patch also adds timing metrics to extension startup, printing them to the logs, and also adds a new transport to print to the console when doing extension development.

## Tasks

### DIFF
--- a/src/logging/FileTransport.ts
+++ b/src/logging/FileTransport.ts
@@ -2,7 +2,7 @@
 //
 // This source file is part of the VS Code Swift open source project
 //
-// Copyright (c) 2024 the VS Code Swift project authors
+// Copyright (c) 2025 the VS Code Swift project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information


src/extension.ts
```diff
@@ -177,6 +196,26 @@ export async function activate(context: vscode.ExtensionContext): Promise<Api> {
...
+    // Create log directory asynchronously but don't await it to avoid blocking activation
+    const logDirPromise = vscode.workspace.fs.createDirectory(context.logUri);
+
+    const logger = new SwiftLoggerFactory(context.logUri).create(
+        "Swift",
+        "swift-vscode-extension.log"
+    );
+    context.subscriptions.push(logger);
+
+    void Promise.resolve(logDirPromise)
```
award999, line 209:
```suggestion
    void logDirPromise
```
does it need to wrap in Promise.resolve?

src/logging/SwiftLogger.ts
```diff
@@ -130,17 +193,25 @@ export class SwiftLogger implements vscode.Disposable {
         return fullMessage;
     }
 
+    private _cachedOutputChannelLevel: string | undefined;
```
award999: nit: add to list of properties at the top of the class

src/logging/SwiftLogger.ts
```diff
@@ -61,6 +67,43 @@ export class SwiftLogger implements vscode.Disposable {
                 winston.format.colorize()
             ),
         });
+
+        // Add file transport asynchronously to avoid blocking startup
+        setTimeout(() => {
```
award999: instead of putting this logic into the logger, why don't we create our own file transport. Logic to open a file is simple, can implement a `public close?(): void;` handler to close the file, and then at least we don't need to replicate the timestamp and message format below. The logger itself can just keep feeding the messages to the transport without considering how each transport is setup

award999: The `OutputChannelTransport` shows how you can get the formatted message

src/extension.ts
plemarquand: Yep, `vscode.workspace.fs.createDirectory` returns a `Thenable<T>` only has `then()`. One of the ways to catch errors off of it is to wrap the Thenable in a Promise.resolve.

src/logging/FileTransport.ts
```diff
@@ -0,0 +1,92 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2024 the VS Code Swift project authors
```
award999:
```suggestion
// Copyright (c) 2025 the VS Code Swift project authors
```

src/logging/SwiftLogger.ts
```diff
@@ -71,37 +82,53 @@ export class SwiftLogger implements vscode.Disposable {
...
+                    this.fileTransport.close();
                 },
             },
             vscode.workspace.onDidChangeConfiguration(e => {
                 if (
                     e.affectsConfiguration("swift.outputChannelLogLevel") ||
                     e.affectsConfiguration("swift.diagnostics")
                 ) {
+                    // Clear cache when configuration changes
+                    this.cachedOutputChannelLevel = undefined;
```
award999, line 94: one day it would be create to make this a common pattern. So many of our setting lookups could likely be cached

Issue #1782, opened 2025-08-13
Support `#Playground` macro / `swift play`


MahdiBM:
**Is your feature request related to a problem? Please describe.**
This is an issue to track progress on `#Playground` and `swift play` support.
I'm not sure if there is much actionable right now.

Repo: https://github.com/apple/swift-play-experimental

**Describe the solution you'd like**
Xcode 26 supports a `#Playground` macro. The VSCode extension should too.
This is helpful for quick experiments.

**Describe alternatives you've considered**
Live on with `swift repl` or `swift -e`. Have a "playground file" which is ignored by git and manually run it from the command palette.

**Additional context**
I find myself using `swift repl` or `swift -e` a bunch of times every week, and the new `#Playground` macro could be a better alternatives to those.

adam-fowler:
You know you can run the current file as a Swift script regardless of whether it has been saved or not using the command `run Swift script`. This is how I test small Swift scripts. Otherwise I create a temporary Swift snippet if I want to test something in the current library I'm developing.

MahdiBM:
Yeah that's what I meant by "Have a "playground file" which is ignored by git and manually run it from the command palette.".

I guess in that sense we're less in need of the #Playground macro than Xcode users since I dont recall Xcode having an option to run a single file?

MahdiBM:
(Added a link to the [swift-play](https://github.com/apple/swift-play-experimental) repo. Forgot to paste that link when I created the issue)

adam-fowler:
I think @compnerd started something related to integrating the repl in VSCode. It was just as I was extracting myself from VSCode dev so didn't pay too much attention.

MahdiBM:
yeah I remember some movements around Jupyter notebook support ... not sure where that went.

compnerd:
It didn't go anywhere AFAIK. @plemarquand seemed interested but I think that they lost interest or forgot about it.

plemarquand:
I played with it a bit but ultimately figured a proper Swift Jupyter kernel would be a better experience

PR #1924, opened 2025-11-05
Support "swift.play" CodeLens

award999:
## Description
- Advertise to the LSP we support "swift.play"
- Add a "swift.play" command, hiding it from the command palette
- Fix order of env variables passed to task so variables like DYLD_LIBRARY_PATH are not overwritten accidentally

Issue: #1782

## Tasks

### DIFF
--- a/src/commands/runPlayground.ts
+++ b/src/commands/runPlayground.ts
@@ -38,11 +38,8 @@ export interface WorkspacePlaygroundItem extends PlaygroundItem {
 export async function runPlayground(
     folderContext: FolderContext,
     tasks: TaskManager,
-    item?: PlaygroundItem
+    item: PlaygroundItem
 ) {
-    if (!item) {
-        return false;
-    }
     const id = item.label ?? item.id;
     const task = createSwiftTask(
         ["play", id],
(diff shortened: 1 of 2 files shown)


src/commands/runPlayground.ts
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Executes a {@link vscode.Task task} to run swift playground.
+ */
+export async function runPlayground(
+    folderContext: FolderContext,
+    tasks: TaskManager,
+    item?: PlaygroundItem
+) {
+    if (!item) {
```
plemarquand: nit: Can item be made non optional?

src/tasks/SwiftTaskProvider.ts
```diff
@@ -469,9 +473,9 @@ export class SwiftTaskProvider implements vscode.TaskProvider {
         const env = platform?.env ?? task.definition.env;
         const fullCwd = resolveTaskCwd(task, platform?.cwd ?? task.definition.cwd);
         const fullEnv = {
-            ...configuration.swiftEnvironmentVariables,
-            ...swiftRuntimeEnv(),
-            ...env,
+            ...swiftRuntimeEnv(), // From process.env first
```
plemarquand, line 476: nice, it probably should have always been in this order

award999: Yep :)

src/commands/runPlayground.ts
award999: Done

plemarquand (approved):

PR #1932, opened 2025-11-11
Reduce extension activation time

matthewbastien:
## Description
This PR moves the creation of the `WorkspaceContext` off of the critical path for extension activation. Most of the extension activation logic has been moved into a new `InternalSwiftExtensionApi` class that has synchronous `activate()` and `deactivate()` methods to help avoid future performance issues with activation. Going forward we should keep the `activate()` method as slim as possible in `src/extension.ts`.

Acquiring the `WorkspaceContext` is now asynchronous using new `waitForWorkspaceContext()` and `withWorkspaceContext()` methods. This will be a breaking change for users of the extension API. I've bumped the version to 1.0.0 to indicate this.

Issue: #2082

## Tasks


src/extension.ts
```diff
@@ -16,299 +16,91 @@ import "source-map-support/register";
...
+     */
+    withWorkspaceContext<T>(
+        task: (ctx: WorkspaceContext) => T | Promise<T>,
+        token?: vscode.CancellationToken
+    ): Promise<T>;
+
+    /**
+     * Activate the extension.
+     *
+     * **DO NOT** use this method directly. It is exposed for testing purposes only.
```
award999: We can make `activate` optional and only expose in development environment or if testing. A non-nice extension may ignore the `**DO NOT**`

matthewbastien: Good idea. I split this up into two different interfaces and use a proxy to hide the `activate()`, `deactivate()`, and `dispose()` methods when not running under test.

src/commands.ts
```diff
@@ -122,170 +100,200 @@ export enum Commands {
...
-        vscode.commands.registerCommand(
-            Commands.RESTART_LSP,
-            async () => await restartLSPServer(ctx)
+        vscode.commands.registerCommand(Commands.DEBUG_SNIPPET, target =>
+            api.withWorkspaceContext(ctx => debugSnippet(ctx, ...unwrapTreeItem(target)))
         ),
-        vscode.commands.registerCommand(
-            "swift.reindexProject",
-            async () => await reindexProject(ctx)
+        vscode.commands.registerCommand(Commands.RUN_PLUGIN_TASK, () => () => runPluginTask()),
```
plemarquand: This is going to break with the doubly wrapped call:
```suggestion
        vscode.commands.registerCommand(Commands.RUN_PLUGIN_TASK, () => runPluginTask()),
```

src/InternalSwiftExtensionApi.ts
```diff
@@ -0,0 +1,496 @@
...
+            const totalActivationTime = Date.now() - activationStartTime;
+            this.logger.info(
+                `Extension activation completed in ${totalActivationTime}ms (subscriptions: ${subscriptionsElapsed}ms, final-steps: ${finalStepsElapsed}ms)`
+            );
+        } catch (error) {
+            // Handle configuration validation errors with UI that points the user to the poorly configured setting
+            if (error instanceof ConfigurationValidationError) {
+                return; // User is notified by code in configuration.ts
+            }
+            const errorMessage = getErrorDescription(error);
```
plemarquand, line 201: We should set this.state to the `failed` state here. Eventually we should give them a way to re-activate after they've fixed the setting.

src/InternalSwiftExtensionApi.ts
```diff
@@ -0,0 +1,496 @@
...
+        }
+
+        switch (operation) {
+            case FolderOperation.add:
+                // Create launch.json files based on package description, don't block execution.
+                void makeDebugConfigurations(folder);
+
+                if (await folder.swiftPackage.foundPackage) {
+                    // do not await for this, let packages resolve in parallel
+                    void folderAdded(folder, workspace);
```
plemarquand: We should either `await` both `foundPackage` _and_ `folderAdded`, OR we should do something like:
```
                void folder.swiftPackage.foundPackage.then(async foundPackage => {
                    if (foundPackage) {
                        await folderAdded(folder, workspace);
                    }
                });
```
Came across this in https://github.com/swiftlang/vscode-swift/pull/2093/changes#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626R304

src/InternalSwiftExtensionApi.ts
```diff
@@ -0,0 +1,496 @@
...
+    ) {
+        this.contextKeys = new ContextKeyManager();
+        const logSetupStartTime = Date.now();
+        this.logger = configureLogging(this.extensionContext);
+        const logSetupElapsed = Date.now() - logSetupStartTime;
+        this.logger.info(`Log setup completed in ${logSetupElapsed}ms`);
+    }
+
+    async waitForWorkspaceContext(): Promise<WorkspaceContext> {
+        if (!this.state) {
```
plemarquand: `this.state == undefined` is an implicit 4th state; we could explicitly add this 4th state to `type State` as `uninitialized`

plemarquand (changes requested):

src/InternalSwiftExtensionApi.ts
matthewbastien: Added an `uninitialized` state.

src/InternalSwiftExtensionApi.ts
matthewbastien: I like that approach. I ended up just using the code you provided.

src/InternalSwiftExtensionApi.ts
matthewbastien: Good catch!

src/commands.ts
matthewbastien: Good catch!

rbenegal (approved):
Great work! Thank you!

src/InternalSwiftExtensionApi.ts
```diff
@@ -0,0 +1,518 @@
...
+
+            const finalStepsStartTime = Date.now();
+            const cancellationSource = new vscode.CancellationTokenSource();
+            this.state = {
+                type: "initializing",
+                activatedBy,
+                promise: this.initializeWorkspace(cancellationSource.token)
+                    .then(({ workspaceContext, subscriptions }) => {
+                        if (cancellationSource.token.isCancellationRequested) {
+                            throw new vscode.CancellationError();
```
plemarquand, line 166: Should we dispose the workspaceContext and subscriptions here? Or is this always going to be cancelled in `deactivate` which disposes things right after calling cancel?

src/InternalSwiftExtensionApi.ts
```diff
@@ -0,0 +1,518 @@
...
+        if (this.state.type === "failed") {
+            throw this.state.error;
+        }
+        if (this.state.type === "active") {
+            return this.state.context;
+        }
+        return await this.state.promise;
+    }
+
+    async withWorkspaceContext<T>(task: (ctx: WorkspaceContext) => T | Promise<T>): Promise<T> {
```
plemarquand, line 113: The API in SwiftExtensionApi.ts states this takes a `token?: vscode.CancellationToken` but we never declare or use it.

src/SwiftExtensionApi.ts
```diff
@@ -48,8 +48,37 @@ export interface SwiftExtensionApi {
...
+    /**
+     * The {@link WorkspaceContext} if it is currently available.
+     *
+     * The Swift Extension API starting in version 1.0.0 delays workspace initialization in order to
+     * speed up activation. Use {@link waitForWorkspaceContext} or {@link withWorkspaceContext}
+     * to wait for the workspace to be initialized.
+     *
+     * Prior to 1.0.0 this was the only way to access the workspace context.
+     */
+    workspaceContext?: WorkspaceContext;
```
plemarquand: Was it necessary to drop `readonly`?

src/WorkspaceContext.ts
```diff
@@ -266,7 +262,7 @@ export class WorkspaceContext implements ExternalWorkspaceContext, vscode.Dispos
         }
     }
 
-    dispose() {
+    async dispose(): Promise<void> {
```
plemarquand, line 257: `async` dispose? Doesn't this make it no longer conform to vscode.Disposable?

src/sourcekit-lsp/LanguageClientToolchainCoordinator.ts
```diff
@@ -150,7 +150,13 @@ export class LanguageClientToolchainCoordinator implements vscode.Disposable {
         return client;
     }
 
-    dispose() {
+    async dispose(): Promise<void> {
```
plemarquand, line 154: I think we dispose of this in in WorkspaceContext but we aren't awaiting it. Also, same comment about no longer conforming to vscode.Disposable, though I think that would be a compiler error so I guess it does?

src/InternalSwiftExtensionApi.ts
```diff
@@ -0,0 +1,518 @@
...
+        this.logger.info(
+            `Workspace initialization completed in ${totalActivationTime}ms (toolchain: ${toolchainElapsed}ms, workspace-context: ${workspaceContextElapsed}ms, subscriptions: ${subscriptionsElapsed}ms, workspace-folders: ${workspaceFoldersElapsed}ms)`
+        );
+
+        return { workspaceContext, subscriptions };
+    }
+
+    async deactivate(): Promise<void> {
+        this.contextKeys.isActivated = false;
+        if (this.state?.type === "initializing") {
```
plemarquand: States type is not nullable, the `?.` shouldn't be required

src/sourcekit-lsp/LanguageClientToolchainCoordinator.ts
matthewbastien: Wow, good catch. I intentionally removed the `implements vscode.Disposable` from `LanguageClientManager` for exactly this reason, but apparently I forgot to do the same here. Honestly, I thought the same thing you did about the compiler error, but it turns out that the return type of `vscode.Disposable.dispose()` is `any`. 🙄

We should probably define our own Disposable type that's more strictly typed to avoid this issue. I'll look into it.

src/InternalSwiftExtensionApi.ts
matthewbastien: Yup, good catch!

src/InternalSwiftExtensionApi.ts
matthewbastien: This was leftover from a different change I wanted to make. Removed it.

src/SwiftExtensionApi.ts
matthewbastien: No that was an accident. I've added it back.

src/WorkspaceContext.ts
matthewbastien: Yeah and this is intentional. The `LanguageClient`'s `dispose()` method is also asynchronous and we need to dispose of it here.

src/sourcekit-lsp/LanguageClientToolchainCoordinator.ts
matthewbastien: I've fixed the issue, but I'm going to put up a separate PR for adding our own `Disposable` and `AsyncDisposable` interfaces with better type safety.

src/InternalSwiftExtensionApi.ts
matthewbastien: Removed this and a few in other places.

plemarquand (approved):

PR #1936, opened 2025-11-13
Properly handle toolchains from Command Line Tools on macOS

matthewbastien:
## Description
Adjust toolchain discovery to detect which tool manages the `swift` binary that was found in `PATH`. This allows us to delegate to the appropriate tool when trying to find toolchain executables. Possible toolchain managers are:
- `xcrun`: Toolchain was either installed via `xcode-select --install` or by downloading Xcode. Executes `xcrun --find swift` to find the path to the toolchain. Only available on macOS.
- `swiftly`: Toolchain was installed via swiftly. Executes `swiftly use --print-location` to find the path to the toolchain.
- `swiftenv`: Toolchain was installed via swiftenv. Executes `swiftenv which swift` to find the path to the toolchain.
- `unknown`: A toolchain manager was not recognized. The path to the toolchain is assumed to be two levels above the path to the `swift` binary in `PATH`.

Detecting toolchains that need `xcrun` is done by performing an `objdump` on the `swift` binary and looking for an `__xcrun_shim` data segment. If one is found then we know that we have to use `xcrun` when querying for toolchain executables even if the path to the toolchain does not live within an `Xcode.app`. This makes the toolchain detection logic more resilient instead of using brittle heuristics that only look at the toolchain path.

Issue: #1077

## Tasks

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -542,7 +542,7 @@ export class SwiftToolchain {
         }
         // Use objdump to determine if this is an xcrun shim.
         try {
-            const objdumpOutput = await execFile("/usr/bin/objdump", ["-h", binary]);
+            const objdumpOutput = await execFile("xcrun", ["objdump", "-h", binary]);
             return objdumpOutput.stdout.includes("__xcrun_shim");
         } catch (error) {
             logger?.error(error);


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 ### Fixed 
 
 - Fixed an issue where the activation of the extension was held up while waiting on the debug configuration to update ([#1914](https://github.com/swiftlang/vscode-swift/pull/1914))
+- Fixed an issue where `lldb-dap` could not be found in Command Line Tools toolchains ([#1936](https://github.com/swiftlang/vscode-swift/pull/1936))
```
award999: wrong release

src/toolchain/toolchain.ts
```diff
@@ -532,139 +517,88 @@ export class SwiftToolchain {
...
     private static async getToolchainPath(
-        swiftPath: string,
+        swiftBinaryPath: string,
         extensionRoot: string,
         cwd?: vscode.Uri,
         logger?: SwiftLogger
-    ): Promise<string> {
+    ): Promise<{
+        toolchainPath: string;
+        toolchainManager: "xcrun" | "swiftly" | "swiftenv" | "unknown";
```
award999: duplicated this string list twice, let's extract to a type or make a string enum

test/chai-path-plugin.ts
```diff
@@ -0,0 +1,49 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2024 the VS Code Swift project authors
```
award999: 2025

test/unit-tests/toolchain/toolchain.test.ts
```diff
@@ -27,23 +26,32 @@ suite("SwiftToolchain Unit Test Suite", () => {
...
+            Error("execFile() was not properly mocked for this test.")
+        );
     });
 
     teardown(() => {
         mockFS.restore();
     });
 
+    suite("create()", () => {
+        suite("macOS", () => {
```
award999: no tests?

test/unit-tests/toolchain/toolchain.test.ts
```diff
@@ -174,12 +178,12 @@ suite("SwiftToolchain Unit Test Suite", () => {
                     },
                 });
                 const sut = createSwiftToolchain({
-                    swiftFolderPath: "/toolchains/swift-6.0.0/usr/bin",
-                    toolchainPath: "/toolchains/swift-6.0.0",
+                    swiftBinaryPath: "/toolchains/swift-6.0.0/usr/bin",
```
award999: Is this an accurate path? If so `swiftBinaryPath` would make be think path to `swift` and we should make it `swiftBinPath` instead

test/chai-path-plugin.ts
```diff
@@ -0,0 +1,49 @@
...
+             * compare them to ensure consistent behavior on all platforms.
+             *
+             * @param expected The expected path string.
+             */
+            path(expected: string): Assertion;
+        }
+    }
+}
+
+export function chaiPathPlugin(chai: Chai.ChaiStatic, _utils: Chai.ChaiUtils): void {
```
award999, line 65: Did this serve a purpose beyond making the assertions below easier? If not then it's in place now I guess but in the future we should do these in a separate tech debt PR to cut down on the noise in this bug fix. I assume there are a lot of cases across our tests that could have made use of this that are going to be missed

src/toolchain/toolchain.ts
```diff
@@ -532,139 +517,88 @@ export class SwiftToolchain {
...
-            } catch {
-                return swiftPath;
-            }
-        } else {
-            return swiftPath;
+    private static async isXcrunShim(binary: string): Promise<boolean> {
+        if (!(await fileExists(binary))) {
+            return false;
         }
+        const objdumpOutput = await execFile("objdump", ["-h", binary]);
```
award999: this may not be the objdump we want. Could be brew installed. Similar to lldb-dap, we should see use the objdump in the toolchain as it behaves differently

award999:
If you set `swift.path` to `<home>/.swiftly/bin` or whatever the path is, `swiftly` was getting resolved as the swift binary so just want to make sure we didn't regress there

test/unit-tests/toolchain/toolchain.test.ts
matthewbastien: Whoops. This was a remnant from me trying to create tests around toolchain creation which ended up being a herculean effort.

test/unit-tests/toolchain/toolchain.test.ts
matthewbastien: Remnant from me experimenting with having a `swiftBinaryPath` instead of `swiftFolderPath` since we don't actually use it anywhere except to determine if `swift.path` changed. Changed this back to the original behavior.

src/toolchain/toolchain.ts
matthewbastien: Good point. I've changed this to explicitly execute `/usr/bin/objdump`.

test/chai-path-plugin.ts
matthewbastien: I can split this into a separate PR if we want. Some code paths use `path.resolve()` which changes the slashes to the current platform while others don't. This meant that certain tests would need to handle this while others didn't (and it isn't obvious why when looking at the test code). I wanted a consistent way to check for path equality.

matthewbastien: Split into a separate PR that needs to be merged first: #1950

matthewbastien:
> If you set `swift.path` to `<home>/.swiftly/bin` or whatever the path is, `swiftly` was getting resolved as the swift binary so just want to make sure we didn't regress there

Yes, this still works when I test it locally.

src/toolchain/toolchain.ts
```diff
@@ -532,139 +519,88 @@ export class SwiftToolchain {
...
-            } catch {
-                return swiftPath;
-            }
-        } else {
-            return swiftPath;
+    private static async isXcrunShim(binary: string): Promise<boolean> {
+        if (!(await fileExists(binary))) {
+            return false;
         }
+        const objdumpOutput = await execFile("/usr/bin/objdump", ["-h", binary]);
```
plemarquand: `objdump` is only installed if the user has the CLT or Xcode installed. Are we guaranteed to know whether they have one of these installed at this point? Otherwise we should handle the case where `objdump` doesn't exist.

src/toolchain/toolchain.ts
```diff
@@ -423,51 +429,32 @@ export class SwiftToolchain {
...
-            if (error instanceof Error) {
-                errorMessage += `:\n${error.message}`;
-            }
-            throw new Error(errorMessage);
+        // Fallback to xcrun if the active toolchain comes from Xcode/CommandLineTools
+        if (
+            this.manager === "xcrun" ||
+            (this.manager === "swiftly" && (await Swiftly.inUseVersion()) === "xcode")
+        ) {
+            const { stdout } = await execFile("xcrun", ["--find", executable]);
```
plemarquand, line 452: What if this execFile fails?

src/utilities/shell.ts
```diff
@@ -13,20 +13,66 @@
...
+            }
+            case "win32": {
+                const command = "where.exe";
+                const args = [binaryName];
+                const { stdout, stderr } = await execFile(command, args);
+                const foundPaths = stdout.trimEnd().split("\r\n");
+                if (foundPaths.length === 0) {
+                    throw createParsingError({ command, args, stdout, stderr });
+                }
+                return foundPaths;
```
plemarquand: Seems like we always default to the first element where this function is called; any reason to not just return the first element? If this is PATH based binary resolution the platform is going to use the first one it encounters.

`findBinaryInPath` also implies a single result.

package.json
```diff
@@ -498,7 +498,7 @@
           "swift.path": {
             "type": "string",
             "default": "",
-            "markdownDescription": "Override the default path of the folder containing the Swift executables. The default is to look in the `PATH` environment variable. This path is also used to search for other executables used by the extension like `sourcekit-lsp` and `lldb`.",
```
plemarquand, line 501: Is this no longer true? If so, how would I specify a custom path to look for other tools?

src/toolchain/toolchain.ts
matthewbastien: Fair point. I'll add some form of error handling to suggest that they need to at least `xcode-select --install` to get the CommaneLineTools.

Even `swiftly` requires you to have at least the CommandLineTools installed for it to work. So, I think it's a reasonable requirement for the extension to as well.

package.json
matthewbastien: This is only partially true.

The only case where this is true is if you download a public toolchain, install it manually, and then point the Swift extension at it with `swift.path`. This is because the path to the toolchain and `swift.path` happen to be the same thing.

However, with `xcrun` shims or `swiftly` this is not the case. For example, I could set the `swift.path` to `/usr/bin` which is an `xcrun` shim on macOS. In this case, the actual toolchain path lives in the CommandLineTools or Xcode. We would really want to use the `sourcekit-lsp` or `lldb-dap` from the toolchain path because they don't always get their own shim in `/usr/bin`.

src/toolchain/toolchain.ts
matthewbastien: `findToolchainExecutable()` will throw an error and expects whoever called it to handle it if they want to. It's expected that this function will throw an error if it fails to find the executable.

src/utilities/shell.ts
matthewbastien: Yeah, I was just trying to keep the existing behaviour of logging when multiple paths are found, but I think that's pretty rare. I just removed the ability for this function to return multiple paths.

src/toolchain/toolchain.ts
matthewbastien: So, it seems like calling `objdump` on a fresh macOS defaults to showing a pop-up to install the command line tools. I've added a check that calls `xcode-select -p` to determine if there's an Xcode/CommandLineTools installed before calling out to `objdump`. If either of these fails then we just assume that the swift binary is not an xcrun shim and toolchain selection will possibly fail elsewhere.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Suggest "Open Documentation" when toolchain not found ([#1939](https://github.com/swiftlang/vscode-swift/pull/1939))
 - Make sure all folder operation listeners get past folder add events ([#1945](https://github.com/swiftlang/vscode-swift/pull/1945))
+- Fixed an issue where `lldb-dap` could not be found in Command Line Tools toolchains ([#1936](https://github.com/swiftlang/vscode-swift/pull/1936))
```
award999: need to rebase now that 2.14.1 is out

src/toolchain/toolchain.ts
```diff
@@ -532,139 +519,94 @@ export class SwiftToolchain {
...
+        // Make sure that either Xcode or CommandLineTools are installed before attempting to run objdump.
+        try {
+            await execFile("xcode-select", ["-p"]);
+        } catch (error) {
+            logger?.error(error);
+            return false;
+        }
+        // Use objdump to determine if this is an xcrun shim.
+        try {
+            const objdumpOutput = await execFile("/usr/bin/objdump", ["-h", binary]);
```
award999: could `xcrun objdump ...` be more bulletproof?

matthewbastien: Good idea. Done.

CHANGELOG.md
award999: needs to be up by line 4, otherwise looks good

matthewbastien: Whoops, didn't notice that. Fixed.

PR #1942, opened 2025-11-17
Remove Swift > Clean Build Folder menu item in swift editor area

plemarquand:
Right clicking the editor area of a Swift file would show a Swift menu entry that just contained Clean Build Folder. This doesn't make much sense to have here, so clean it up.

To clean the build folder use the "Swift > Clean Build Folder" command from the command pallete.

## Tasks

### DIFF
--- a/package.json
+++ b/package.json
@@ -1423,11 +1423,6 @@
           "command": "swift.debugSnippet",
           "group": "1_file@5",
           "when": "swift.fileIsSnippet"
-        },
-        {
-          "command": "swift.cleanBuild",
-          "group": "2_pkg@1",
-          "when": "swift.hasPackage"
         }
       ],
       "view/title": [


adam-fowler:
I use the right click menu clean project all the time. Why does it not make sense to include it there?

plemarquand:
@adam-fowler this came up during the [VS Code Office Hours](https://github.com/swiftlang/vscode-swift/discussions/1935). The concern was this is a project wide action that appears on a context menu for a single swift file. In the discussion they were looking for some Swift code actions (found under the Refactoring context menu item) and having a `Swift` entry in this context menu was a red herring, as they expected this would be where actions to modify the current Swift file would live.

award999:
You could also add a keybinding for this command so you don't need to click through the context menu

PR #1943, opened 2025-11-17
Add logging getLLDBLibPath to better diagnose failures

plemarquand:
This will help diagnose the failure to find LLDB on Windows 6.2

## Tasks

### DIFF
--- a/src/debugger/lldb.ts
+++ b/src/debugger/lldb.ts
@@ -17,6 +17,7 @@ import * as fs from "fs/promises";
 import * as path from "path";
 import * as vscode from "vscode";
 
+import { SwiftLogger } from "../logging/SwiftLogger";
 import { SwiftToolchain } from "../toolchain/toolchain";
 import { Result } from "../utilities/result";
 import { IS_RUNNING_UNDER_TEST, execFile } from "../utilities/utilities";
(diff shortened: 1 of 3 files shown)


src/debugger/lldb.ts
```diff
@@ -45,11 +46,15 @@ export function updateLaunchConfigForCI(
...
  * @returns Library path for LLDB
  */
-export async function getLLDBLibPath(toolchain: SwiftToolchain): Promise<Result<string>> {
+export async function getLLDBLibPath(
+    toolchain: SwiftToolchain,
+    logger?: SwiftLogger
+): Promise<Result<string>> {
     let executable: string;
     try {
         executable = await toolchain.getLLDB();
```
award999, line 55: while we're here, maybe a debug log of what was found

src/debugger/lldb.ts
```diff
@@ -68,6 +73,7 @@ export async function getLLDBLibPath(toolchain: SwiftToolchain): Promise<Result<
         // should just the version of lldb that comes with CodeLLDB. We return a failure with no message
         // to indicate we want it to fail silently.
         if (process.platform === "win32") {
+            logger?.error(`Failed to get LLDB library path for lldb on Windows, error: ${error}`);
             return Result.makeFailure(undefined);
         }
     }
```
award999, line 84: Can't seem to add comment to lines not in patch, but in `findLibLLDB` can we add debug logging for which folders we try to find it in, and debug result of `await findLibLLDB(pathHint)`

award999: Also don't think it would be too verbose to log directory contents returned by `readdir` in `findFileByPattern`

award999:
Some suggestions of how we could taker debug logging even further

PR #1950, opened 2025-11-19
Add a chai plugin to check for path equality

matthewbastien:
## Description
Make it easier to check for path equality in tests by creating a new chai plugin. It can be used like so:
```typescript
expect(somePath).to.equalPath("/path/to/some/file.txt");
```

This assertion converts both paths to the same file separator before comparison to ensure consistent behavior on all platforms.

## Tasks


matthewbastien:
The two failed tests are also failing in Nightly. Merging.

Issue #1966, opened 2025-11-28
Activation of extension hangs and never completes


aoberoi:
**Describe the bug**
When viewing Running extensions, I see an "Unresponsive" description next to this extension. The extension seems to never finish activation. See screenshot. 

<img width="1166" height="105" alt="Image" src="https://github.com/user-attachments/assets/e1571b01-d915-4223-8238-b5751922905d" />

I believe this is blocking other extensions from completing their activation, although I'm not totally sure about that. I have vscodevim installed, and that extension is not activating. Also there's a pending badge next to the Source Control lens, so that seems to be blocked too.

I'm happy to collect more logs, or profile, or whatever. But I would need instructions on what might be useful and how to collect it.

**To Reproduce**
1. Open a folder with Swift source files

**Expected behavior**
Swift language server to be running.

**Environment**
 - OS: macOS
 - Swift version (output of `swift --version`): Apple Swift version 6.2.1 (swift-6.2.1-RELEASE)
 - Visual Studio Code version: 1.106.3 (Universal)
 - vscode-swift version: 2.14.1

**Diagnostics Bundle**
The diagnostics bundle can help determine the root cause of bugs much quicker. To share this information execute the `>Swift: Capture Diagnostics Bundle` command and attach the archive to the bug report.

The command never completes. I just see "Activating Extensions..." in the status bar for a few seconds.

*Select the "Minimal" bundle to only share the Swift extension's log and your configured `swift.` settings. Select the "Full" bundle if you're comfortable sharing the SourceKit-LSP diagnostic bundle and LLDB-DAP logs, which may contain source code*

**Additional context**

---

Info from VSCode issue template:

- Issue Type: `Performance`
- Extension Name: `swift-vscode`
- Extension Version: `2.14.1`
- OS Version: `Darwin arm64 25.1.0`
- VS Code version: `1.106.3`

[swiftlang.swift-vscode-unresponsive.cpuprofile.txt](https://github.com/user-attachments/files/23829973/swiftlang.swift-vscode-unresponsive.cpuprofile.txt)

Find more details here: https://github.com/microsoft/vscode/wiki/Explain-extension-causes-high-cpu-load

plemarquand:
Thanks for the report. I haven't seen this behaviour before and have some questions to try and narrow down the cause.

- Is this happening consistently every time you open VS Code?
- Are you opening an empty workspace, a Swift project, or some other kind of project? 
[...]

PR #1975, opened 2025-12-02
Don't follow the swiftly symlink when determining in use toolchain

plemarquand:
When swiftly is installed via Homebrew it is added to the path at a location like `/opt/homebrew/bin/swiftly`, which is a symlink to the brew cellar located at `/opt/homebrew/bin/Cellar/swiftly/1.1.0/bin/`

When printing the in use toolchain location with `swiftly use --print-location` we were launching swiftly with the realpath of the symlink, which caused Swiftly to print the "swiftly is unlinked" warning message before printing the toolchain location.

The extension went on to take this message and parse it as a path, causing the extension activation to hang.

Instead of using this realpath when invoking swiftly to get the toolchain path, if we determine the use is using swiftly just invoke it normally using `swiftly` instead of using the full path.

Issue: #1966

## Tasks

### DIFF
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -571,7 +571,7 @@ export class SwiftToolchain {
 
             if (path.basename(realSwift) === "swiftly") {
                 try {
-                    const inUse = await Swiftly.inUseLocation(realSwift, cwd);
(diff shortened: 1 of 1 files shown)


matthewbastien (approved):
LGTM

Issue #1871, opened 2025-10-01
Breakpoints are not being installed


award999:
**Describe the bug**

**To Reproduce**
1. Set a source breakpoint in a Swift source file
2. Debug a test or Swift binary
3. Notice how the breakpoint gets hit, but the red breakpoint dot disappears because it is not marked as installed

**Expected behavior**
Always mark the breakpoint as installed

**Environment**
 - OS: All
 - Swift version (output of `swift --version`): All Swift 6.0+

**Additional context**
- Always happens for Swift executable binary
- Always happens we debugging a swift-testing test
<img width="479" height="180" alt="Image" src="https://github.com/user-attachments/assets/09e454b4-97a8-43ff-94a3-1c54f0c8ec60" />

- Doesn't seem to happen for XCTest
<img width="431" height="180" alt="Image" src="https://github.com/user-attachments/assets/9d8f4148-8a64-4a32-8aa4-c8b9c4a83379" />

amanmaurya92:
I can work on this!

matthewbastien:
The underlying lldb-dap issue has been fixed in more recent Swift toolchains. I added a workaround to the Swift extension that will force breakpoints to be re-resolved when they are hit in the debugger, but aren't shown as verified in the UI.

For verification you'll need a 6.0.x toolchain.

PR #1979, opened 2025-12-03
Add dynamic local schema detection for SourceKit-LSP

plemarquand:
## Description
Implement support for using local schema files from the Swift toolchain when available, with automatic fallback to remote GitHub schemas for older toolchains.

This enables the extension to use the accurate schema for the users toolchain, if its present, while maintaining backward compatibility with older toolchains.

## Tasks
- [X] Required tests have been written
- [X] Documentation has been updated
- [x] Added an entry to CHANGELOG.md if applicable

PR #1989, opened 2025-12-06
debug: fallback to mark breakpoints verified on stopped/breakpoint events

amanmaurya92:
## Description
LLDB-DAP sometimes fails to mark breakpoints as verified in the `setBreakpoints` response, even though the breakpoint is hit during debugging. This leads to the VS Code UI showing the breakpoint as “unverified” while execution actually stops there.

This PR adds a safe fallback that marks breakpoints as verified when we detect them being hit via:
- a `stopped` event with `reason: "breakpoint"`, or
- a `breakpoint` event containing a valid source and line.

This improves consistency between actual debugger behavior and the UI, especially for Swift executables and `swift test`.

Issue: https://github.com/swiftlang/vscode-swift/issues/1871

## Tasks


src/debugger/logTracker.ts
```diff
@@ -131,6 +133,102 @@ export class LoggingDebugAdapterTracker implements vscode.DebugAdapterTracker {
         }
     }
 
+    private handleBreakpointFallback(rawMsg: unknown): void {
+        try {
+            // Minimal typed view of the event payload we care about.
+            type FrameLike = { source?: { path?: string }; line?: number };
+            type BodyLike = {
```
award999: suggestion: instead of defining these types ourselves, we already have "@vscode/debugprotocol" as a dependency in the package.json and can make use of `BreakpointEvent` and other types defined in there. That way the type checker can flag mismatches :)

src/debugger/logTracker.ts
```diff
@@ -131,6 +133,102 @@ export class LoggingDebugAdapterTracker implements vscode.DebugAdapterTracker {
         }
     }
 
+    private handleBreakpointFallback(rawMsg: unknown): void {
```
award999:
```suggestion
    private handleBreakpointEvent(rawMsg: unknown): void {
```

Ideally we could make use of `ProtocolMessage` or some other type defined in "@vscode/debugprotocol" instead of `unknown`

src/debugger/logTracker.ts
```diff
@@ -131,6 +133,102 @@ export class LoggingDebugAdapterTracker implements vscode.DebugAdapterTracker {
...
+                const sourcePath = top?.source?.path;
+                const line = top?.line;
+                if (!sourcePath || typeof line !== "number") {
+                    return;
+                }
+
+                const bpLine0 = line - 1; // VS Code uses 0-based lines
+
+                const breakpoints = vscode.debug.breakpoints.filter(
+                    b => b instanceof vscode.SourceBreakpoint
```
award999: thought: we probably are ok since looks like a lot of tests passed, but something to keep in mind, is `instanceof` we try to avoid because if an object was instantiated by the test runner, the extension under test will return false because the types are not the same (different node instances), so we could make the check `b => !!b.location` since that is the distinguishing property of the `SourceBreakpoint`

award999: actually I see you seem to have that check below, so ya we could simplify by either not performing the filter and let check below make sure we have a location, or if we check for location when filter we can drop that extra `if (!loc)` check below

src/debugger/logTracker.ts
```diff
@@ -131,6 +133,102 @@ export class LoggingDebugAdapterTracker implements vscode.DebugAdapterTracker {
...
+                ) as vscode.SourceBreakpoint[];
+
+                for (const bp of breakpoints) {
+                    const loc = bp.location;
+                    if (!loc) {
+                        continue;
+                    }
+                    if (loc.uri.fsPath !== sourcePath) {
```
award999: question: does the DAP spec ensure that the `sourcePath` we get will be for the full path?

src/debugger/logTracker.ts
```diff
@@ -131,6 +133,102 @@ export class LoggingDebugAdapterTracker implements vscode.DebugAdapterTracker {
...
+
+                    vscode.debug.removeBreakpoints([bp]);
+                    vscode.debug.addBreakpoints([bp]);
+                    break;
+                }
+                return;
+            }
+        } catch (err) {
+            // eslint-disable-next-line no-console
+            console.error("Breakpoint fallback error:", err);
```
award999: todo: Let's get the logger into `LoggingDebugAdapterTracker` so that way the error ends up in the log file

award999 (commented):
@amanthatdoescares thanks so much for the PR 🙏 I have some initial comments here to get you started. @matthewbastien is more familiar with the debugging codebase so let's have him take a look to.

Also the open todo for updating changelog in the PR template, you just need to do something like https://github.com/swiftlang/vscode-swift/pull/1979/commits/85d5132ed277fde61ec883ebfbf177a2310b155b, but this being a bug fix, you'd put it under the `### Fixed` section for the latest release.

src/debugger/logTracker.ts
amanmaurya92: Got it. I’ll switch to the types from @vscode/debugprotocol instead of redefining them.

src/debugger/logTracker.ts
amanmaurya92: Sure, I can take this change. I’ll switch the param to a proper protocol type instead of unknown.

src/debugger/logTracker.ts
amanmaurya92: Right — the DAP spec doesn’t guarantee absolute paths. I’ll normalize the path before comparing.

src/debugger/logTracker.ts
amanmaurya92: Sure, I can move this into the tracker so it gets logged properly.

src/debugger/logTracker.ts
```diff
@@ -131,6 +136,126 @@ export class LoggingDebugAdapterTracker implements vscode.DebugAdapterTracker {
...
+            const eventMsg = msg as DebugProtocol.Event;
+
+            if (!msg || msg.type !== "event") {
+                return;
+            }
+
+            const normalizePath = (p: string): string => {
+                try {
+                    // If adapter sends a URI-like path, parse it
+                    if (p.startsWith("file://") || p.includes("://")) {
```
matthewbastien: **suggestion:** You can use `vscode.URI.parse()` with strict mode enabled here. That way we don't have to do any manual parsing of the string. Using `parse()` with strict mode set to `true` will throw an error when the string is not a valid URI. We can fall back on `vscode.Uri.file()` like you have below if an error is thrown.

src/debugger/logTracker.ts
```diff
@@ -131,6 +136,126 @@ export class LoggingDebugAdapterTracker implements vscode.DebugAdapterTracker {
...
+
+                    vscode.debug.removeBreakpoints([bp]);
+                    vscode.debug.addBreakpoints([bp]);
+                    break;
+                }
+                return;
+            }
+        } catch (err) {
+            // eslint-disable-next-line no-console
+            this.logger?.error(`Breakpoint fallback error: ${String(err)}`);
```
matthewbastien:
```suggestion
            this.logger?.error(Error('Breakpoint fallback error', { cause: err });
```

**nitpick:** let the logger handle converting the error to a string. You can create a new Error with a `cause` so that both stack traces will appear in the logs.

src/debugger/logTracker.ts
```diff
@@ -131,6 +136,126 @@ export class LoggingDebugAdapterTracker implements vscode.DebugAdapterTracker {
...
+                const top = frames[0];
+                const sourcePath = top?.source?.path;
+                const line = top?.line;
+                if (!sourcePath || typeof line !== "number") {
+                    return;
+                }
+
+                const bpLine0 = line - 1; // VS Code uses 0-based lines
+
+                const breakpoints = vscode.debug.breakpoints.filter(
```
matthewbastien, line 186: **question:** Is there anything stopping us from using `.filter(b => b instanceof vscode.SourceBreakpoint)` here instead of type casting? Would make this code a lot easier to read.

src/debugger/logTracker.ts
```diff
@@ -131,6 +136,126 @@ export class LoggingDebugAdapterTracker implements vscode.DebugAdapterTracker {
...
+                    b => !!(b as vscode.SourceBreakpoint).location
+                ) as vscode.SourceBreakpoint[];
+
+                for (const bp of breakpoints) {
+                    const loc = bp.location;
+                    if (!loc) {
```
matthewbastien: **suggestion:** This check is unnecessary as it should be handled by the filter above.

src/debugger/logTracker.ts
```diff
@@ -131,6 +136,126 @@ export class LoggingDebugAdapterTracker implements vscode.DebugAdapterTracker {
...
+                    }
+                    if (loc.uri.fsPath !== normalizePath(sourcePath)) {
+                        continue;
+                    }
+                    if (loc.range.start.line !== bpLine0) {
+                        continue;
+                    }
+
+                    // Force a UI refresh so the breakpoint shows as installed.
+                    vscode.debug.removeBreakpoints([bp]);
```
matthewbastien, line 201: **question:** Will this also trigger when the breakpoint is already resolved? It's probably not a huge issue, but I think we should be only doing this when the breakpoint that's being hit isn't marked as resolved.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### Fixed
 
+- Fix breakpoint filtering fallback in debugger logging ([#1989](https://github.com/swiftlang/vscode-swift/pull/1989))
```
matthewbastien: **chore:** You'll have to rebase this since the CHANGELOG was recently updated for version 2.14.2. This may require moving this bullet to a new `### Fixed` section.

matthewbastien (changes requested):
Thank you for fixing this! I just have some minor comments to make the code easier to read.

src/debugger/logTracker.ts
amanmaurya92: We avoid instanceof vscode.SourceBreakpoint here because it’s unreliable in VS Code tests. Breakpoints can be created in a different JS context, causing instanceof checks to fail even though the object behaves like a SourceBreakpoint. Checking for the presence of location avoids that cross-context issue and is more robust

src/debugger/logTracker.ts
amanmaurya92: Good point, that check is redundant given the filter above. I’ve removed it.

CHANGELOG.md
amanmaurya92: Will do — I’ll rebase and move the entry to the current unreleased ### Fixed section.

matthewbastien: Thanks! The location looks good now.

Though I just noticed that the wording here would only be meaningful to contributors and probably won't make much sense to users of the extension. Maybe we could go with something like `Fixed breakpoints showing up as unresolved when using lldb-dap`?

src/debugger/logTracker.ts
matthewbastien: I don't think that applies in this case. We've had issues with `instanceof` in the past, but that problem occurs when performing checks on classes like `WorkspaceContext` (basically any class in `src/`) because the tests and the extension use transpiled versions of that class from different paths. The extension uses the one from the VSIX and the tests use the one from `dist/`. So, `instanceof` will always fail when comparing `WorkspaceContext` within any of our tests. The `vscode.SourceBreakpoint` class, on the other hand, comes from the extension host which is the same between the tests and the extension.

That being said I realize this is more of a nitpick than anything else since they accomplish the same thing here. So, I'm just going to mark this as resolved.

matthewbastien:
@amanthatdoescares I tried this out locally with Swift 6.2.0 and it looks like the logic isn't working. My breakpoint is not being resolved despite the debug session stopping at the breakpoint. Debugging through the tracker code you added shows that the breakpoint coming from the DAP does not have a source path and so the code is just ignoring it.

What version of Swift did you test this with locally? There might be some differences in `lldb-dap` between that and the version I tested with?

I'd like some integration tests to be added for this before we can merge. Preferably by starting a debugging session with a breakpoint and then making sure that it gets resolved once hit.

amanmaurya92:
@matthewbastien Thanks for the detailed review. After looking at it again, I realized I completely missed the real issue. My fallback logic only handles the happy case where the DAP breakpoint includes a source.path, and I didn’t actually test the scenario you ran into — the one where the adapter sends a breakpoint without a path. So the code ends up ignoring exactly the situation it was supposed to fix. Definitely a blooper on my part.

It’s clear now that this isn’t an environment/version difference — the logic just isn’t handling the missing-path case at all.
I’ll take another pass at this with the right approach:

- Reproduce the behavior properly and capture the actual setBreakpoints request + the breakpoint event.
- Fix the resolution logic so it works even when the adapter doesn’t provide a source.path.
- Add the integration test you mentioned: start a debug session with a breakpoint and assert that it gets verified once hit.

I’ll update the PR once I’ve got the full flow working as intended. Thanks again for pointing out.

matthewbastien:
Closing for inactivity.

PR #1996, opened 2025-12-11
Ignore `--scratch-path` when calling `swift package plugin --list`

plemarquand:
If the user has a `swift.buildPath` set, and they've got backgroundCompilation turned on, the extension will attempt to list their package's plugins with a `--scratch-path`, which `swift package plugin --list` does not accept. This causes the command to fail.

Omit this argument from the command.

## Tasks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - When editing .sourcekit-lsp/config.json use the JSON schema from the toolchain ([#1979](https://github.com/swiftlang/vscode-swift/pull/1979))
 
+### Fixed
+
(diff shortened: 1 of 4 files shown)


matthewbastien (approved):

thibault-ml:
> which `swift package plugin --list` does not accept

Maybe things have changed since then, but it does work today, if the argument are in the right place.

The following command execute successfully and has the expected behaviour (of using the scratch path):
```
swift package --scratch-path /tmp/codecrafters-swift-scratch plugin --list
```

From looking at the PR, I believe the previous code emitted `swift package plugin --scratch-path /tmp/codecrafters-swift-scratch --list`, which indeed doesn't work, since the `--scratch-path` arguments need to be passed to the `package` command, not the `plugin` one.

thibault-ml:
As a follow-up, I've created #2234 with more details.

Issue #1647, opened 2025-06-25
Add a command to create a documentation catalog


award999:
To accompany the docc and markdown documentation preview work for Swift, a new command should be added that walks the user through the generation of a documentation catalog.

matthewbastien:
Generating a DocC archive is very straightforward. SourceKit-LSP is just looking for a folder within a target's sources that has the `.docc` extension. So, this command should allow the user to select any target(s) to generate the folder for and then create it. We can also add a markdown file for the module as well which just needs to have the module name of the target as the title:

```markdown
# ``ModuleName``
```

amanmaurya92:
@award999  I’ve implemented this and will open a PR shortly.

Flamki:
Quick status check on this feature:

- I confirmed there is currently no command contributed for creating a DocC catalog (there is `swift.previewDocumentation`, but not a create-catalog command).
- The implementation PR is still open: https://github.com/swiftlang/vscode-swift/pull/2006

Would maintainers like to continue/review #2006, or should this issue be reopened for a fresh PR if that branch is stale?

PR #2006, opened 2025-12-17
Add command to create a Swift DocC documentation catalog

amanthatdoescares:
## Description
Add command to create DocC documentation catalog (#1647)

- Add Swift: Create DocC Documentation Catalog command that lets users
  select one or more SwiftPM targets and creates a .docc catalog inside
  each target's directory (SourceKit-LSP discovers docs there).
- Module markdown uses title format # ``ModuleName`` per maintainer.
- Register command in register(WorkspaceContext); add unit tests.

This removes the need to manually set up the DocC folder structure and makes it easier to start documenting Swift packages directly from VS Code.

Issue: https://github.com/swiftlang/vscode-swift/issues/1647

## Tasks
- [x] Documentation has been updated (not applicable)

### DIFF
--- a/src/commands/createDocumentationCatalog.ts
+++ b/src/commands/createDocumentationCatalog.ts
@@ -21,28 +21,49 @@ type TargetPickItem = vscode.QuickPickItem & {
     targetName: string;
 };
 
+async function findAvailableTargets(folder: FolderContext): Promise<TargetPickItem[]> {
+    const items: TargetPickItem[] = [];
+    const rootPath = folder.folder.fsPath;
+    if (!folder.swiftPackage) {
+        return items;
+    }
+    const targets = await folder.swiftPackage.getTargets();
+    for (const target of targets) {
+        const base = path.join(rootPath, target.path);
+        if (await folderExists(base)) {
+            items.push({
+                label: target.name,
+                description: target.type,
+                detail: target.path,
+                basePath: base,
+                targetName: target.name,
+            });
+        }
+    }
+    if (items.length === 0) {
+        void vscode.window.showErrorMessage(
+            "No Swift package targets found. Open a folder that contains a Package.swift."
+        );
+    }
+    return items;
+}
+
 /**
  * Walks the user through creating DocC documentation catalogs for one or more
  * SwiftPM targets. SourceKit-LSP discovers documentation via a .docc folder
  * within a target's sources.
  */
-export async function createDocumentationCatalog(
-    ctx: WorkspaceContext,
-    folderContext?: FolderContext
-): Promise<void> {
-    let folder: FolderContext | undefined = folderContext;
+export async function createDocumentationCatalog(ctx: WorkspaceContext): Promise<void> {
+    let folder: FolderContext | undefined;
 
-    if (!folder && ctx.folders.length === 1) {
+    if (ctx.folders.length === 1) {
         folder = ctx.folders[0];
-    }
-
-    if (!folder) {
-        if (ctx.folders.length === 0) {
-            void vscode.window.showErrorMessage(
-                "Creating a documentation catalog requires an open workspace folder."
-            );
-            return;
-        }
+    } else if (ctx.folders.length === 0) {
+        void vscode.window.showErrorMessage(
+            "Creating a documentation catalog requires an open workspace folder."
+        );
+        return;
+    } else {
         const selected = await selectFolder(
             ctx,
             "Select a workspace folder to create the DocC catalog in"
@@ -53,38 +74,7 @@ export async function createDocumentationCatalog(
         folder = selected[0];
     }
 
-    const rootPath = folder.folder.fsPath;
-
-    const itemsPromise = (async (): Promise<TargetPickItem[]> => {
-        const items: TargetPickItem[] = [];
-        if (!folder.swiftPackage) {
-            return items;
-        }
-        const targets = await folder.swiftPackage.getTargets();
-        for (const target of targets) {
-            const base = path.join(rootPath, target.path);
-            if (await folderExists(base)) {
-                items.push({
-                    label: target.name,
-                    description: target.type,
-                    detail: target.path,
-                    basePath: base,
-                    targetName: target.name,
-                });
-            }
-        }
-        return items;
-    })();
-
-    const items = await itemsPromise;
-    if (items.length === 0) {
-        void vscode.window.showErrorMessage(
-            "No Swift package targets found. Open a folder that contains a Package.swift."
-        );
-        return;
-    }
-
-    const selections = await vscode.window.showQuickPick(items, {
+    const selections = await vscode.window.showQuickPick(findAvailableTargets(folder), {
         title: "Create DocC Documentation Catalog",
         placeHolder: "Select one or more targets to add a documentation catalog to.",
         canPickMany: true,
@@ -94,17 +84,19 @@ export async function createDocumentationCatalog(
         return;
     }
 
+    const results = await Promise.all(
+        selections.map(selection => provisionDoccCatalog(selection.basePath, selection.targetName))
+    );
+
     const created: string[] = [];
     const skipped: string[] = [];
-
-    for (const selection of selections) {
-        const ok = await provisionDoccCatalog(selection.basePath, selection.targetName);
-        if (ok) {
+    selections.forEach((selection, index) => {
+        if (results[index]) {
             created.push(`${selection.targetName}.docc`);
         } else {
             skipped.push(`${selection.targetName}.docc`);
         }
-    }
+    });
 
     if (created.length > 0) {
         void vscode.window.showInformationMessage(`Created DocC catalog(s): ${created.join(", ")}`);
--- a/test/unit-tests/commands/createDocumentationCatalog.test.ts
+++ b/test/unit-tests/commands/createDocumentationCatalog.test.ts
@@ -13,15 +13,16 @@
 //===----------------------------------------------------------------------===//
 import { expect } from "chai";
 import * as path from "path";
-import { stub } from "sinon";
 import * as vscode from "vscode";
 
 import { FolderContext } from "@src/FolderContext";
+import { Target } from "@src/SwiftExtensionApi";
+import { SwiftPackage } from "@src/SwiftPackage";
 import { WorkspaceContext } from "@src/WorkspaceContext";
 import { createDocumentationCatalog } from "@src/commands/createDocumentationCatalog";
 import * as filesystem from "@src/utilities/filesystem";
 
-import { instance, mockGlobalModule, mockGlobalObject, mockObject } from "../../MockUtils";
+import { instance, mockFn, mockGlobalModule, mockGlobalObject, mockObject } from "../../MockUtils";
 
 suite("createDocumentationCatalog Command Test Suite", () => {
     const windowMock = mockGlobalObject(vscode, "window");
@@ -33,22 +34,17 @@ suite("createDocumentationCatalog Command Test Suite", () => {
     const targetBasePath = path.join(rootPath, targetPath);
 
     function createMockFolderContext(
-        overrides: {
-            getTargets?: () => Promise<{ name: string; path: string; type: string }[]>;
-        } = {}
+        targets: Partial<Target>[] = [{ name: targetName, path: targetPath, type: "library" }]
     ): FolderContext {
-        const getTargets =
-            overrides.getTargets ??
-            (() => Promise.resolve([{ name: targetName, path: targetPath, type: "library" }]));
-        const swiftPackage = {
-            getTargets: stub().callsFake(getTargets),
-        };
+        const mockedSwiftPackage = mockObject<SwiftPackage>({
+            getTargets: mockFn(s => s.resolves(targets)),
+        });
         return instance(
             mockObject<FolderContext>({
                 folder: vscode.Uri.file(rootPath),
-                swiftPackage: swiftPackage as unknown as FolderContext["swiftPackage"],
+                swiftPackage: instance(mockedSwiftPackage),
             })
-        ) as unknown as FolderContext;
+        );
     }
 
     function createMockWorkspaceContext(folders: FolderContext[]): WorkspaceContext {
@@ -56,7 +52,7 @@ suite("createDocumentationCatalog Command Test Suite", () => {
             mockObject<WorkspaceContext>({
                 folders,
             })
-        ) as unknown as WorkspaceContext;
+        );
     }
 
     setup(() => {
@@ -81,9 +77,7 @@ suite("createDocumentationCatalog Command Test Suite", () => {
     });
 
     test("shows error when package has no targets", async () => {
-        const folder = createMockFolderContext({
-            getTargets: () => Promise.resolve([]),
-        });
+        const folder = createMockFolderContext([]);
         const ctx = createMockWorkspaceContext([folder]);
         filesystemMock.folderExists.resolves(true);
 
@@ -111,10 +105,10 @@ suite("createDocumentationCatalog Command Test Suite", () => {
         const folder = createMockFolderContext();
         const ctx = createMockWorkspaceContext([folder]);
         (windowMock.showQuickPick as import("sinon").SinonStub<any[], any>).callsFake(
-            (items: any) =>
-                Promise.resolve(items).then((resolved: readonly vscode.QuickPickItem[]) =>
-                    resolved && resolved.length > 0 ? [resolved[0]] : undefined
-                )
+            async items => {
+                const resolved = await items;
+                return resolved.length > 0 ? [resolved[0]] : undefined;
+            }
         );
 
         await createDocumentationCatalog(ctx);
@@ -134,10 +128,10 @@ suite("createDocumentationCatalog Command Test Suite", () => {
         const ctx = createMockWorkspaceContext([folder]);
         filesystemMock.provisionDoccCatalog.resolves(false);
         (windowMock.showQuickPick as import("sinon").SinonStub<any[], any>).callsFake(
-            (items: any) =>
-                Promise.resolve(items).then((resolved: readonly vscode.QuickPickItem[]) =>
-                    resolved && resolved.length > 0 ? [resolved[0]] : undefined
-                )
+            async items => {
+                const resolved = await items;
+                return resolved.length > 0 ? [resolved[0]] : undefined;
+            }
         );
 
         await createDocumentationCatalog(ctx);
@@ -153,23 +147,16 @@ suite("createDocumentationCatalog Command Test Suite", () => {
     });
 
     test("creates catalogs for multiple selected targets", async () => {
-        const getTargets = () =>
-            Promise.resolve([
-                { name: "LibA", path: "Sources/LibA", type: "library" },
-                { name: "LibB", path: "Sources/LibB", type: "library" },
-            ]);
-        const folder = createMockFolderContext({ getTargets });
+        const folder = createMockFolderContext([
+            { name: "LibA", path: "Sources/LibA", type: "library" },
+            { name: "LibB", path: "Sources/LibB", type: "library" },
+        ]);
         const ctx = createMockWorkspaceContext([folder]);
-        filesystemMock.provisionDoccCatalog
-            .onFirstCall()
-            .resolves(true)
-            .onSecondCall()
-            .resolves(true);
         (windowMock.showQuickPick as import("sinon").SinonStub<any[], any>).callsFake(
-            (items: any) =>
-                Promise.resolve(items).then((resolved: readonly vscode.QuickPickItem[]) =>
-                    resolved && resolved.length > 0 ? [...resolved] : undefined
-                )
+            async items => {
+                const resolved = await items;
+                return resolved.length > 0 ? [...resolved] : undefined;
+            }
         );
 
         await createDocumentationCatalog(ctx);
@@ -185,25 +172,6 @@ suite("createDocumentationCatalog Command Test Suite", () => {
         );
         expect(windowMock.showInformationMessage).to.have.been.calledOnceWith(
             "Created DocC catalog(s): LibA.docc, LibB.docc"
-        );
-    });
-
-    test("uses folderContext when provided", async () => {
-        const folder = createMockFolderContext();
-        const ctx = createMockWorkspaceContext([folder]);
-        (windowMock.showQuickPick as import("sinon").SinonStub<any[], any>).callsFake(
-            (items: any) =>
-                Promise.resolve(items).then((resolved: readonly vscode.QuickPickItem[]) =>
-                    resolved && resolved.length > 0 ? [resolved[0]] : undefined
-                )
-        );
-
-        await createDocumentationCatalog(ctx, folder);
-
-        expect(windowMock.showQuickPick).to.have.been.calledOnce;
-        expect(filesystemMock.provisionDoccCatalog).to.have.been.calledOnceWith(
-            targetBasePath,
-            targetName
         );
     });
 });


src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,40 @@
+import * as fs from "fs";
```
matthewbastien: **issue:** Missing copyright header.

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,40 @@
...
+        placeHolder: "MyModule",
+        validateInput: value =>
+            value.trim().length === 0 ? "Module name cannot be empty" : undefined,
+    });
+
+    if (!moduleName) {
+        return; // user cancelled
+    }
+
+    const rootPath = folders[0].uri.fsPath;
```
matthewbastien: **issue:** We can't assume there will only be a single workspace folder. The command should guide the user through creating a catalog in any workspace folder they wish to use.

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,40 @@
+import * as fs from "fs";
+import * as path from "path";
+import * as vscode from "vscode";
+
+export async function createDocumentationCatalog(): Promise<void> {
+    const folders = vscode.workspace.workspaceFolders;
+    if (!folders || folders.length === 0) {
+        await vscode.window.showErrorMessage("Open a workspace first.");
```
matthewbastien: **issue:** This error message is a bit vague. Can we change it to something like `Creating a documentation catalog requires that a folder or workspace be opened`?

**suggestion:** We can also make it less likely for this situation to occur by only showing `swift.createDocumentationCatalog` in the command palette if a workspace is present. See the other commands with `when` clauses in the `package.json` for examples.

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,40 @@
...
+        validateInput: value =>
+            value.trim().length === 0 ? "Module name cannot be empty" : undefined,
+    });
+
+    if (!moduleName) {
+        return; // user cancelled
+    }
+
+    const rootPath = folders[0].uri.fsPath;
+    const doccDir = path.join(rootPath, `${moduleName}.docc`);
```
matthewbastien: **issue:** We can't assume that the user will only place a catalog at the root of the workspace. Documentation catalogs can be located anywhere in the project structure. However, they are most commonly placed within targets from the `Package.swift` to provide documentation for them. I would show a quick pick with the available targets and an option at the end for creating a standalone documentation catalog.

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,40 @@
...
+    const markdownFile = path.join(doccDir, `${moduleName}.md`);
+
+    if (fs.existsSync(doccDir)) {
```
matthewbastien: **suggestion:** We can show this error in the input and not allow continuing instead of forcing the user to restart the whole command. See the create new project command for an example of this.

src/commands/createDocumentationCatalog.ts
award999: also you don't want to `await` if not waiting for input. Error messages don't go away until cleared. In this case use `void` instead of `await`

src/commands/createDocumentationCatalog.ts
award999: should probably use `fs/promises` not `fs` so that you don't need to use all the synchronous methods below. Most cases we want to use asynchronous for file operations

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,40 @@
...
+        await vscode.window.showErrorMessage(
+            `Documentation catalog "${moduleName}.docc" already exists.`
+        );
+        return;
+    }
+
+    fs.mkdirSync(doccDir, { recursive: true });
+    fs.writeFileSync(markdownFile, `# ${moduleName}\n`, { encoding: "utf8" });
+
+    await vscode.window.showInformationMessage(
```
award999: `await` => `void`

src/commands/createDocumentationCatalog.ts
amanmaurya92: yes i understand. i have resolved it in recent commit resolved it.

src/commands/createDocumentationCatalog.ts
amanmaurya92: yes, i added that in latest commit

src/commands/createDocumentationCatalog.ts
amanmaurya92: @matthewbastien VS Code may still show commands in the palette search or history even when their when clause is false. The command is correctly gated and cannot be executed without an open workspace.

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,74 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2021-2024 the VS Code Swift project authors
```
matthewbastien:
```suggestion
// Copyright (c) 2025 the VS Code Swift project authors
```

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,74 @@
...
+import * as fsSync from "fs";
+import * as fs from "fs/promises";
+import * as path from "path";
+import * as vscode from "vscode";
+
+export async function createDocumentationCatalog(): Promise<void> {
+    const folders = vscode.workspace.workspaceFolders;
+    if (!folders || folders.length === 0) {
+        void vscode.window.showErrorMessage(
+            "Creating a documentation catalog requires that a folder or workspace be opened"
```
matthewbastien:
```suggestion
            "Creating a documentation catalog requires that a folder or workspace be opened."
```

src/commands/createDocumentationCatalog.ts
matthewbastien: There's still a check using `existsSync()` which should be changed to its async counterpart.

src/commands/createDocumentationCatalog.ts
matthewbastien: I agree that there's a check in the command for this which is good. However, a `when` clause will make it so that the command does not appear in the command palette when it shouldn't. We do this with a bunch of other commands as well.

Which search history are you referring to? I tested this locally with another command (`Swift: Preview Documentation`) and its `when` clause makes it so that it doesn't show in the command palette if the workspace is not opened even if it was a recently used command.

matthewbastien (changes requested):

src/commands/createDocumentationCatalog.ts
amanmaurya92: Thanks for pointing that out — I aligned the when clause with Swift: Preview Documentation, and the command is now hidden when no workspace is open. It’s working as intended now.

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,150 @@
...
+        await fs.access(path.join(rootPath, "Package.swift"));
+    } catch {
+        hasPackageSwift = false;
+    }
+
+    let targets: string[] = [];
+
+    if (hasPackageSwift) {
+        try {
+            const { stdout } = await execFileAsync("swift", ["package", "dump-package"], {
```
plemarquand: This information is already computed and stored on the `folderContext.swiftPackage` for each Swift Package folder open in the workspace. See https://github.com/swiftlang/vscode-swift/blob/main/src/commands.ts#L80 for a reference of how to get the folderContext for the currently active folder and pass it in to the command.

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,150 @@
...
+            targets = [];
+        }
+    }
+
+    const items: DoccLocationPickItem[] = [];
+
+    for (const name of targets) {
+        const srcPath = path.join(rootPath, "Sources", name);
+        try {
+            await fs.access(srcPath);
```
plemarquand: you can use `fileExists` in `utilities/filesystem` instead

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,150 @@
...
+        prompt: "Enter Swift module name",
+        placeHolder: "MyModule",
+        validateInput: async value => {
+            if (value.trim().length === 0) {
+                return "Module name cannot be empty";
+            }
+
+            const doccDir = path.join(basePath, `${value}.docc`);
+            try {
+                await fs.access(doccDir);
```
plemarquand: `fileExists` here as well

src/commands/createDocumentationCatalog.ts
amanmaurya92: Thanks for pointing this out — I wasn’t aware this information was already available via FolderContext. I’ll refactor the command to use folderContext.swiftPackage instead of invoking swift package dump-package.

src/commands/createDocumentationCatalog.ts
amanmaurya92: i think folderExists is much accurate here

src/commands/createDocumentationCatalog.ts
amanmaurya92: once i read filesystem i think folderExists is much more accurate in this case

src/commands/createDocumentationCatalog.ts
amanmaurya92: done

src/commands/createDocumentationCatalog.ts
matthewbastien: `fileExists()` and `folderExists()` check for the path in question to be a file or folder respectively. You're going to want to use `pathExists()` because you can't create the folder if anything exists at that path regardless of whether or not it is a file, folder, symlink, etc.

src/commands/createDocumentationCatalog.ts
matthewbastien: You'll have to use `pathExists()`.

matthewbastien:
I'd also like to have an integration test for this since the logic will be interacting with the `FolderContext`. Something that runs the command and makes sure that all of the targets show up correctly in the quick pick. It should also make sure that the catalog gets created properly.

src/commands/createDocumentationCatalog.ts
amanmaurya92: @matthewbastien Target discovery still uses folderExists() since those must be directories.

src/commands/createDocumentationCatalog.ts
amanmaurya92: Yes, i have made that commit now

amanmaurya92:
@matthewbastien  I’ve added an integration test that runs the command end-to-end, verifies SwiftPM targets appear in the QuickPick, and asserts the DocC catalog is created on disk.

This is my first time writing an integration test in this codebase, so I’d really appreciate a review on the test structure and coverage.

test/integration-tests/commands/createDocumentationCatalog.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    activateExtensionForSuite({
+        async setup(ctx) {
+            workspaceContext = ctx;
+            folderContext = await folderInRootWorkspace("defaultPackage", workspaceContext);
+            await workspaceContext.focusFolder(folderContext);
+        },
+    });
+
+    test("creates a DocC catalog for a SwiftPM target", async () => {
```
matthewbastien: **issue:** The nested `test()` is unnecessary and will make it so that this test never actually runs.

test/integration-tests/commands/createDocumentationCatalog.test.ts
```diff
@@ -0,0 +1,63 @@
...
+import * as sinon from "sinon";
+import * as vscode from "vscode";
+
+import { FolderContext } from "@src/FolderContext";
+import { WorkspaceContext } from "@src/WorkspaceContext";
+
+import { tag } from "../../tags";
+import { activateExtensionForSuite, folderInRootWorkspace } from "../utilities/testutilities";
+
+tag("large").suite("Create Documentation Catalog Command", function () {
```
matthewbastien: **issue:** This suite should be marked as either `"medium"` or `"small"`. It definitely doesn't need 10 minutes just to create a directory. I'm leaning towards `"small"` in this instance since it really shouldn't take long at all.

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,138 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2025 the VS Code Swift project authors
+// Licensed under Apache License v2.0
+//
```
matthewbastien, line 7: **issue:** This is missing part of the copyright header.

test/integration-tests/commands/createDocumentationCatalog.test.ts
```diff
@@ -0,0 +1,63 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2025 the VS Code Swift project authors
+// Licensed under Apache License v2.0
```
matthewbastien: **issue:** This is missing part of the copyright header.

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,138 @@
...
+    });
+
+    if (!selection) {
+        return;
+    }
+
+    const basePath = selection.basePath;
+
+    // ---- module name input ----
+    const moduleName = await vscode.window.showInputBox({
```
matthewbastien: **issue:** By convention the module name for a target should match the target name if the user selected one in the previous step. The extension should only ask for a folder if the user selected a standalone documentation catalog.

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,138 @@
...
+    // ---- workspace folder resolution (standard pattern) ----
+    if (!folder) {
+        if (ctx.folders.length === 0) {
+            void vscode.window.showErrorMessage(
+                "Creating a documentation catalog requires an open workspace folder."
+            );
+            return;
+        }
+
+        if (ctx.folders.length === 1) {
```
matthewbastien: **question:** Do we really need to ask which folder to use before showing available targets? We can just remove a step here and show all of the available targets from all known folders. We can also use separators if needed in the quick pick to show which folder each target came from.

If the user selects a standalone documentation catalog and there are multiple folders then we can ask.

test/integration-tests/commands/createDocumentationCatalog.test.ts
```diff
@@ -0,0 +1,63 @@
...
+import { activateExtensionForSuite, folderInRootWorkspace } from "../utilities/testutilities";
+
+tag("large").suite("Create Documentation Catalog Command", function () {
+    let folderContext: FolderContext;
+    let workspaceContext: WorkspaceContext;
+
+    activateExtensionForSuite({
+        async setup(ctx) {
+            workspaceContext = ctx;
+            folderContext = await folderInRootWorkspace("defaultPackage", workspaceContext);
```
matthewbastien: **suggestion:** You can add multiple asset folders here to properly test the logic of finding targets within multiple `Package.swift`.

test/integration-tests/commands/createDocumentationCatalog.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    test("creates a DocC catalog for a SwiftPM target", async () => {
+        test("creates a DocC catalog for a SwiftPM target", async () => {
+            const quickPickStub = sinon.stub(vscode.window, "showQuickPick");
+            const inputBoxStub = sinon.stub(vscode.window, "showInputBox");
+
+            try {
+                inputBoxStub.resolves("MyModule");
+                quickPickStub.callsFake(async itemsOrPromise => {
+                    const items = await Promise.resolve(itemsOrPromise);
+                    return items.find(item => item.label.startsWith("Target:"));
```
matthewbastien: **suggestion:** Validate whether or not the list of targets contains all of the targets that we expect to be there.

test/integration-tests/commands/createDocumentationCatalog.test.ts
```diff
@@ -0,0 +1,63 @@
...
+                await vscode.commands.executeCommand("swift.createDocumentationCatalog");
+                const basePath = folderContext.folder.fsPath;
+                const doccDir = path.join(basePath, "MyModule.docc");
+                const markdownFile = path.join(doccDir, "MyModule.md");
+
+                expect(await fs.stat(doccDir)).to.exist;
+                expect(await fs.stat(markdownFile)).to.exist;
+
+                const contents = await fs.readFile(markdownFile, "utf8");
+                expect(contents).to.contain("# MyModule");
```
matthewbastien: **issue:** The module name when selecting a target should match the module name of that target.

test/integration-tests/commands/createDocumentationCatalog.test.ts
```diff
@@ -0,0 +1,63 @@
...
+            } finally {
+                quickPickStub.restore();
+                inputBoxStub.restore();
+            }
+        });
+    });
+});
```
matthewbastien: **suggestion:** We should also test the logic for creating a standalone documentation catalog.

test/integration-tests/commands/createDocumentationCatalog.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        async setup(ctx) {
+            workspaceContext = ctx;
+            folderContext = await folderInRootWorkspace("defaultPackage", workspaceContext);
+            await workspaceContext.focusFolder(folderContext);
+        },
+    });
+
+    test("creates a DocC catalog for a SwiftPM target", async () => {
+        test("creates a DocC catalog for a SwiftPM target", async () => {
+            const quickPickStub = sinon.stub(vscode.window, "showQuickPick");
```
matthewbastien: **suggestion:** We have methods in `MockUtils.ts` that will do this for you using mocha's `setup()` and `teardown()` blocks. You don't need to use `sinon.stub()` or the `try {} finally {}` here. For example:

```typescript
import { mockGlobalObject } from "../../MockUtils.ts";

suite("some-suite", () => {
    const windowMock = mockGlobalObject(vscode, "window");

    test("some-test", () => {
        windowMock.showQuickPick.resolves("blah");
    });
});
```

test/integration-tests/commands/createDocumentationCatalog.test.ts
```diff
@@ -0,0 +1,63 @@
...
+            const inputBoxStub = sinon.stub(vscode.window, "showInputBox");
+
+            try {
+                inputBoxStub.resolves("MyModule");
```
matthewbastien: **issue:** The module name quick pick should not be shown if selecting a target.

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,138 @@
...
+            basePath: rootPath,
+        });
+
+        return items;
+    });
+
+    // ---- show QuickPick (toolchain-style pattern) ----
+    const selection = await vscode.window.showQuickPick(itemsPromise, {
+        title: "Create DocC Documentation Catalog",
+        placeHolder: "Select where to create the documentation catalog",
```
matthewbastien:
```suggestion
        placeHolder: "Select where to create the documentation catalog.",
```

src/commands/createDocumentationCatalog.ts
amanmaurya92: That makes sense and I agree the flow would be cleaner. I’ll keep the current behavior for now to finish addressing the existing feedback, and then follow up with the multi-folder target selection improvement.

test/integration-tests/commands/createDocumentationCatalog.test.ts
amanmaurya92: done

test/integration-tests/commands/createDocumentationCatalog.test.ts
amanmaurya92: i changed it to small

src/commands/createDocumentationCatalog.ts
amanmaurya92: done

test/integration-tests/commands/createDocumentationCatalog.test.ts
amanmaurya92: yes, done

test/integration-tests/commands/createDocumentationCatalog.test.ts
amanmaurya92: added

test/unit-tests/commands/createDocumentationCatalog.test.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    test("creates catalog for selected target and shows success message", async () => {
+        const folder = createMockFolderContext();
+        const ctx = createMockWorkspaceContext([folder]);
+        (windowMock.showQuickPick as import("sinon").SinonStub<any[], any>).callsFake(
+            (items: any) =>
+                Promise.resolve(items).then((resolved: readonly vscode.QuickPickItem[]) =>
+                    resolved && resolved.length > 0 ? [resolved[0]] : undefined
+                )
+        );
```
matthewbastien, line 112: This can be simplified so that it's easier to read. There are many other instances like this that could be changed as well.

```suggestion
        windowMock.showQuickPick.callsFake(async items => {
            const resolved = await items;
            return resolved.length > 0 ? [resolved[0]] : undefined;
        });
```

test/unit-tests/commands/createDocumentationCatalog.test.ts
```diff
@@ -0,0 +1,209 @@
...
+            })
+        ) as unknown as FolderContext;
+    }
+
+    function createMockWorkspaceContext(folders: FolderContext[]): WorkspaceContext {
+        return instance(
+            mockObject<WorkspaceContext>({
+                folders,
+            })
+        ) as unknown as WorkspaceContext;
```
matthewbastien: Unnecessary cast.

```suggestion
        );
```

test/unit-tests/commands/createDocumentationCatalog.test.ts
```diff
@@ -0,0 +1,209 @@
...
+        const swiftPackage = {
+            getTargets: stub().callsFake(getTargets),
+        };
+        return instance(
+            mockObject<FolderContext>({
+                folder: vscode.Uri.file(rootPath),
+                swiftPackage: swiftPackage as unknown as FolderContext["swiftPackage"],
+            })
+        ) as unknown as FolderContext;
+    }
```
matthewbastien, line 48: This function is way more complicated than it needs to be.

```suggestion
    function createMockFolderContext(
        targets: Partial<Target>[] = [{ name: targetName, path: targetPath, type: "library" }]
    ): FolderContext {
        const mockedSwiftPackage = mockObject<SwiftPackage>({
            getTargets: mockFn(s => s.resolves(targets)),
        });
        return instance(
            mockObject<FolderContext>({
                folder: vscode.Uri.file(rootPath),
                swiftPackage: instance(mockedSwiftPackage),
            })
        );
    }
```

test/unit-tests/commands/createDocumentationCatalog.test.ts
```diff
@@ -0,0 +1,209 @@
...
+                { name: "LibA", path: "Sources/LibA", type: "library" },
+                { name: "LibB", path: "Sources/LibB", type: "library" },
+            ]);
+        const folder = createMockFolderContext({ getTargets });
+        const ctx = createMockWorkspaceContext([folder]);
+        filesystemMock.provisionDoccCatalog
+            .onFirstCall()
+            .resolves(true)
+            .onSecondCall()
+            .resolves(true);
```
matthewbastien: This is unnecessary. Your setup function already does a blanket `.resolve(true)`.

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,117 @@
...
+};
+
+/**
+ * Walks the user through creating DocC documentation catalogs for one or more
+ * SwiftPM targets. SourceKit-LSP discovers documentation via a .docc folder
+ * within a target's sources.
+ */
+export async function createDocumentationCatalog(
+    ctx: WorkspaceContext,
+    folderContext?: FolderContext
```
matthewbastien: This folder context is never used by any caller of `createDocumentationCatalog()`. You can just remove it and simplify the logic below.

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,117 @@
...
+        return items;
+    })();
+
+    const items = await itemsPromise;
+    if (items.length === 0) {
+        void vscode.window.showErrorMessage(
+            "No Swift package targets found. Open a folder that contains a Package.swift."
+        );
+        return;
+    }
```
matthewbastien: Why not just extract this into its own method called `findAvailableTargets()` instead of using an IIFE?

This would not only make the code easier to read, but also allow `vscode.window.showQuickPick()` to indicate progress while the promise is in flight:
```typescript
const selections = await vscode.window.showQuickPick(findAvailableTargets(), {
    title: "Create DocC Documentation Catalog",
    placeHolder: "Select one or more targets to add a documentation catalog to.",
    canPickMany: true,
});
```

src/commands/createDocumentationCatalog.ts
```diff
@@ -0,0 +1,117 @@
...
+    const skipped: string[] = [];
+
+    for (const selection of selections) {
+        const ok = await provisionDoccCatalog(selection.basePath, selection.targetName);
+        if (ok) {
+            created.push(`${selection.targetName}.docc`);
+        } else {
+            skipped.push(`${selection.targetName}.docc`);
+        }
+    }
```
matthewbastien: This should be done inside a `Promise.all()` so that the filesystem operations can be executed in parallel.

src/commands/createDocumentationCatalog.ts
amanmaurya92: can you please review this again, i have had folderContext removed; folder is now derived only from ctx.

src/commands/createDocumentationCatalog.ts
amanmaurya92: i have made the findAvailableTarget()

src/commands/createDocumentationCatalog.ts
amanmaurya92: done :+1:

amanmaurya92:
@matthewbastien i have implemented the changes and it is ready for another review.

PR #2011, opened 2025-12-18
Fix SwiftPM's dropdown showing on projects disabling the SwiftPM integration

rockbruno:
## Description

We use a BSP with `disableSwiftPMIntegration` set to true, but we still had a couple of SwiftPM features showing around. This seemed to be because the extension was returning an empty package instead of `undefined` like it would if the package happened to not exist at all. Making it return `undefined` fixed the issue.

Screenshot 2025-12-18 at 09 11 13
Screenshot 2025-12-18 at 09 29 50

## Tasks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - Fix the wrong toolchain being shown as selected when using swiftly v1.0.1 ([#2014](https://github.com/swiftlang/vscode-swift/pull/2014))
+- Fix extension displaying SwiftPM's project view and automatic build tasks even when `disableSwiftPMIntegration` was true ([#2011](https://github.com/swiftlang/vscode-swift/pull/2011))
 
 ## 2.14.3 - 2025-12-15
 
--- a/src/SwiftPackage.ts
+++ b/src/SwiftPackage.ts
@@ -180,10 +180,16 @@ export interface PackagePlugin {
 type SwiftPackageState = PackageContents | Error | undefined;
 
 function isPackage(state: SwiftPackageState): state is PackageContents {
+    if (state === undefined) {
+        return false;
+    }
     return (state as PackageContents).products !== undefined;
 }
 
 function isError(state: SwiftPackageState): state is Error {
+    if (state === undefined) {
+        return false;
+    }
     return state instanceof Error;
 }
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Fix extension failing to find `lldb-dap` when `swift.path` points at an Xcode toolchain ([#1998](https://github.com/swiftlang/vscode-swift/pull/1998))
 - Omit `--scratch-path` when enumerating plugins with `swift package plugin --list` ([#1996](https://github.com/swiftlang/vscode-swift/pull/1996))
+- Fix extension displaying SwiftPM's project view and automatic build tasks even when `disableSwiftPMIntegration` was true
```
matthewbastien: **chore:** We just released 2.14.3 and updated the changelog. You'll have to rebase the PR on top of main and move this bullet point into the new `### Fixed` header.

src/SwiftPackage.ts
```diff
@@ -262,14 +262,9 @@ export class SwiftPackage {
         toolchain: SwiftToolchain,
         disableSwiftPMIntegration: boolean = false
     ): Promise<SwiftPackageState> {
```
matthewbastien, line 270: **issue:** You'll need to change this to `Promise<SwiftPackageState | undefined>` to match the new behaviour. I'm surprised that the TypeScript compiler didn't catch this which is a bit worrisome....

Some tests are failing in CI because there are parts of `src/SwiftPackage.ts` that rely on the `SwiftPackageState` to exist. Those will need to be cleaned up as well.

CHANGELOG.md
matthewbastien: Oh, and we also like to include a link to the PR at the end.
```suggestion
- Fix extension displaying SwiftPM's project view and automatic build tasks even when `disableSwiftPMIntegration` was true ([#2011](https://github.com/swiftlang/vscode-swift/pull/2011))
```

src/SwiftPackage.ts
rockbruno: I believe it's because SwiftPackage itself is an alias that already covers it slightly above (this case was already possible to hit if Package.swift did not exist at all)

`type SwiftPackageState = PackageContents | Error | undefined;`

PR #2028, opened 2026-01-06
Show progress for package description tasks on activation

plemarquand:
The `swift package show-dependencies` and `swift package describe` commands run on extension activation, and several extension features are gated behind their results. These were run with `execSwift` calls which show no progress in the VS Code UI.

The `swift package show-dependencies` command specifically could kick off a package resolution if the dependencies are missing, and for large packages this could take quite some time. As a result the extension could look like it was not activating.

Move these two commands into VS Code tasks that report their status in the progress bar, and let the user see the actual commands being run in the terminal.

## Tasks
- [X] Added an entry to CHANGELOG.md if applicable

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,15 @@
 
 ## {{releaseVersion}} - {{releaseDate}}
 
+### Added
+
+- Show progress when describing/listing dependencies on package load ([#2028](https://github.com/swiftlang/vscode-swift/pull/2028))
+
 ### Fixed
 
 - Fix the wrong toolchain being shown as selected when using swiftly v1.0.1 ([#2014](https://github.com/swiftlang/vscode-swift/pull/2014))
 - Fix extension displaying SwiftPM's project view and automatic build tasks even when `disableSwiftPMIntegration` was true ([#2011](https://github.com/swiftlang/vscode-swift/pull/2011))
 - Validate extension settings and warn if they are invalid ([#2016](https://github.com/swiftlang/vscode-swift/pull/2016))
-- Show progress when describing/listing dependencies on package load ([#2028](https://github.com/swiftlang/vscode-swift/pull/2028))
 
 ## 2.14.3 - 2025-12-15
 
--- a/src/SwiftPackage.ts
+++ b/src/SwiftPackage.ts
@@ -279,7 +279,7 @@ export class SwiftPackage implements vscode.Disposable {
             return undefined;
         }
 
-        // If there is an existing package load, cancel any running taks first before loading a new one.
+        // If there is an existing package load, cancel any running tasks first before loading a new one.
         this.tokenSource.cancel();
         this.tokenSource.dispose();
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Fix the wrong toolchain being shown as selected when using swiftly v1.0.1 ([#2014](https://github.com/swiftlang/vscode-swift/pull/2014))
 - Fix extension displaying SwiftPM's project view and automatic build tasks even when `disableSwiftPMIntegration` was true ([#2011](https://github.com/swiftlang/vscode-swift/pull/2011))
 - Validate extension settings and warn if they are invalid ([#2016](https://github.com/swiftlang/vscode-swift/pull/2016))
+- Show progress when describing/listing dependencies on package load ([#2028](https://github.com/swiftlang/vscode-swift/pull/2028))
```
matthewbastien: **question:** should this be listed as a new feature rather than a bug fix?

src/SwiftPackage.ts
```diff
@@ -259,45 +257,54 @@ export class SwiftPackage {
...
+    private async performLoadPackageState(
+        folderContext: FolderContext,
         disableSwiftPMIntegration: boolean = false
     ): Promise<SwiftPackageState> {
         // When SwiftPM integration is disabled, return undefined to disable all features
         if (disableSwiftPMIntegration) {
             return undefined;
         }
 
+        // If there is an existing package load, cancel any running taks first before loading a new one.
```
matthewbastien:
```suggestion
        // If there is an existing package load, cancel any running tasks first before loading a new one.
```

src/FolderContext.ts
```diff
@@ -156,16 +153,22 @@ export class FolderContext implements vscode.Disposable {
...
-            );
-        }
+        // List the package's dependencies without blocking folder creation
+        void swiftPackage
+            .loadPackageState(folderContext)
+            .then(async () => await swiftPackage.error)
+            .catch(error => error)
+            .then(async error => {
+                if (error) {
+                    void vscode.window.showErrorMessage(
```
matthewbastien, line 163: **question:** Is there any more information we can give the user or, more importantly, some action that they can perform to resolve the issue? I'd imagine that most of these issues are a result of errors with the `Package.swift`. We could add a button to bring them to that file.

I know that this is the same error message we had before. So, I've got no problem with having this done in another PR (or just raising an issue for it)

matthewbastien (approved):

src/FolderContext.ts
plemarquand: I think it would be a good idea to add a button to take you to your Package.swift, but I'm not 100% sure that the errors presented here will always be able to be resolved by fixing something in that file.

matthewbastien: In an ideal world SPM would have some way of letting VS Code know what the error was (like an exit code or something with common issues) and then we could show a better error, but yeah I agree that it won't always be something that can be fixed in the `Package.swift`.

PR #2030, opened 2026-01-07
Add typings and documentation for the Swift extension's API

matthewbastien:
## Description
The Swift extension exports an API that can be used by other Visual Studio Code extensions. We've had this for a while, but it was not properly documented. This PR creates `src/SwiftExtensionApi.ts` which defines the API surface of the Swift extension. It can be downloaded by other extensions so that they don't have to re-define types. It also allows us to be explicit in what is considered public API.

I've also added a version number to the API which will start at 0.1.0. It follows the semantic versioning standard and will only be updated when the public API changes. Downstream extensions can use this to check for compatibility with the Swift extension.

## Tasks

### DIFF
--- a/src/SwiftExtensionApi.ts
+++ b/src/SwiftExtensionApi.ts
@@ -174,6 +174,15 @@ export interface SwiftToolchain {
 
     /** The version number of this Swift toolchain. */
     readonly swiftVersion: Version;
+
+    /** The SDK currently in use by this toolchain. */
+    readonly sdk?: string;
+
+    /** The user-specified SDK as configured by the `swift.sdk` setting. */
+    readonly customSDK?: string;
+
+    /** The default SDK for this toolchain. */
+    readonly defaultSDK?: string;
 }
 
 /** Workspace Folder Operation types. */
--- a/src/toolchain/toolchain.ts
+++ b/src/toolchain/toolchain.ts
@@ -105,6 +105,10 @@ export function getDarwinTargetTriple(target: DarwinCompatibleTarget): string | 
 export class SwiftToolchain implements ExternalSwiftToolchain {
     public swiftVersionString: string;
 
+    public get sdk(): string | undefined {
+        return this.customSDK ?? this.defaultSDK;
+    }
+
     constructor(
         public manager: ToolchainManager,
         public swiftFolderPath: string, // folder swift executable in $PATH was found in
@@ -446,11 +450,10 @@ export class SwiftToolchain implements ExternalSwiftToolchain {
     }
 
     private basePlatformDeveloperPath(): string | undefined {
-        const sdk = this.customSDK ?? this.defaultSDK;
-        if (!sdk) {
+        if (!this.sdk) {
             return undefined;
         }
-        return path.resolve(sdk, "../../");
+        return path.resolve(this.sdk, "../../");
     }
 
     /**
@@ -490,6 +493,9 @@ export class SwiftToolchain implements ExternalSwiftToolchain {
         }
         if (this.targetInfo.target?.triple) {
             str += `\nDefault Target: ${this.targetInfo.target?.triple}`;
+        }
+        if (this.sdk) {
+            str += `\nSelected SDK: ${this.sdk}`;
         }
         if (this.defaultSDK) {
             str += `\nDefault SDK: ${this.defaultSDK}`;


src/extension.ts
```diff
@@ -187,6 +186,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<Api> {
         return {
             workspaceContext,
             logger,
+            version: new Version(0, 1, 0),
```
plemarquand: We should formalize what requires an update to this version value.

src/SwiftExtensionApi.ts
```diff
@@ -0,0 +1,443 @@
...
+export interface SwiftExtensionApi {
+    /**
+     * The version number of the Swift Extension API. Follows the semantic versioning standard.
+     *
+     * This version number is separate from the version number of the Swift Extension itself. It will only be updated
+     * when changes are made to the API.
+     *
+     * This was added to the API in Swift Extension version 2.16.0. Older versions do not provide an API version number.
+     */
+    readonly version?: Version;
```
Joannis, line 49: Idea; maybe we can have a featureset enum, akin to `grpc`'s services. If an extension can call into the extension API expecting a list of features, you can accept the subset that is supported for the current version. Alternatively you can make that enum and just expose the featureset. Breaking changes would be a different feature e.g. `featureA@1` or `featureA@2`. Not sure if that's a good idea, but might be worth exploring.

Joannis: Otherwise it's very open to interpretation by the implementation and prone to mistakes to see what you still support on a version-by-version basis.

src/SwiftExtensionApi.ts
```diff
@@ -0,0 +1,443 @@
...
+
+    /** The {@link WorkspaceContext} if it is currently available. */
+    readonly workspaceContext?: WorkspaceContext;
```
Joannis, line 52: It would be nice to get an observation on this. Since right now we don't know whether the workspace context becomes available later.

src/SwiftExtensionApi.ts
```diff
@@ -0,0 +1,443 @@
...
+
+    /**
+     * The document URI that currently has focus.
+     *
+     * Focus changes as files are opened and closed by the user.
+     */
+    readonly currentDocument: vscode.Uri | null;
+
+    /** The global toolchain used as the default for this workspace. */
+    readonly globalToolchain: SwiftToolchain;
```
Joannis, line 75: Very helpful!

Joannis: Would it be possible for an extension to prompt & enact an overwrite of this value? For example, Swift on Android will be 6.3 minimum.

src/SwiftExtensionApi.ts
matthewbastien: Definitely. This is going to happen soon (have a look at #1932 for phase 1 of the plan). Right now a missing `WorkspaceContext` basically means that the extension failed to load. This can only be resolved by reloading the extension host so other extensions don't have to worry about it (for now).

src/SwiftExtensionApi.ts
Joannis, line 75: A similar API for the selected SDK would be helpful too.

src/SwiftExtensionApi.ts
```diff
@@ -0,0 +1,443 @@
...
+    readonly foundPackage: Promise<boolean>;
+
+    /** A promise that resolves to the name of this package. */
+    readonly name: Promise<string>;
+
+    /** The URI of the folder containing this package. */
+    readonly folder: vscode.Uri;
+
+    /** Array of available package plugins. */
+    readonly plugins: PackagePlugin[];
```
Joannis, line 393: Excellent!

src/SwiftExtensionApi.ts
matthewbastien: I want to completely get rid of the extension host reloading on toolchain/configuration changes which means adding events around `WorkspaceContext` changes. That will definitely be a breaking change to the API which is why I want to formalize what that API is now.

src/SwiftExtensionApi.ts
Joannis: Would we still manually trigger plugins, or does the VSCode extension intent to trigger these for us?

src/extension.ts
plemarquand: After chatting offline, we should also move this version value into the package.json in a custom field and then read it and use it here. This keeps both the extension version and the extension API version in the same place.

matthewbastien: Added documentation for incrementing the version number and moved it into a new `api-version` property in the `package.json`.

src/SwiftExtensionApi.ts
matthewbastien: We could use `@since` tags in the API documentation comments. That way it's clear in the API which version supports what.

src/SwiftExtensionApi.ts
matthewbastien: The Swift extension provides VS Code tasks for each plugin that can be used to trigger them. We could expose that in the public API depending on what you need.

src/SwiftExtensionApi.ts
matthewbastien: The toolchain selection logic is based on a combination of VS Code settings and looking at what's configured in `PATH`. I think in this case it probably makes more sense to give the user a notification that they need to use Swift 6.3 and then invoke the `swift.selectToolchain` command to get them to change toolchains.

Speaking of commands: it would also be nice to expose their IDs as an enum in the API. I'll have to mull that over.

matthewbastien: Done. It's available inside the `SwiftToolchain` interface.

matthewbastien:
Merging this in preparation for 2.16.0. PRs and issues are welcome for additions to this API!

Issue #2022, opened 2025-12-23
`BuildFlags.buildDirectoryFromWorkspacePath` hardcodes `.build` instead of using the `--scratch-path` if set


plemarquand:
**Describe the bug**
The `BuildFlags.buildDirectoryFromWorkspacePath` doesn't take in to account the user's `--scratch-path` if they have it set via additional build args. This method should be updated and the code that uses it should be tested to make sure it takes in to account a custom build directory set  by the user.

kartikangiras:
will work on this one as well

PR #2032, opened 2026-01-11
Update BuildFlags.buildDirectoryFromWorkspacePath method for scratch-path

kartikangiras:
## Description
- Add `extract(scratchpath())` method to detect `scratch-path` and `buildpath` args.
- Update `buildDirectoryFromWorkspacePath` method with appropriate precedence.
- Add test cases which pass succsefully for the following code change.

Fixes: #2022

## Tasks

### DIFF
--- a/src/toolchain/BuildFlags.ts
+++ b/src/toolchain/BuildFlags.ts
@@ -132,13 +132,30 @@ export class BuildFlags {
         for (let i = 0; i < args.length; i++) {
             const arg = args[i];
             if ((arg === "--scratch-path" || arg === "--build-path") && i + 1 < args.length) {
-                return args[i + 1];
+                const path = args[i + 1];
+                if (path === "") {
+                    throw new Error(`Invalid ${arg}: path cannot be empty`);
+                }
+                if (path.startsWith("--")) {
+                    throw new Error(
+                        `Invalid ${arg}: expected a path but got another flag '${path}'`
+                    );
+                }
+                return path;
             }
             if (arg.startsWith("--scratch-path=")) {
-                return arg.substring("--scratch-path=".length);
+                const path = arg.substring("--scratch-path=".length);
+                if (path === "") {
+                    throw new Error("Invalid --scratch-path: path cannot be empty");
+                }
+                return path;
             }
             if (arg.startsWith("--build-path=")) {
-                return arg.substring("--build-path=".length);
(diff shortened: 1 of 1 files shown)


src/toolchain/BuildFlags.ts
```diff
@@ -139,7 +160,20 @@ export class BuildFlags {
...
+
+        // If not in buildArguments, check packageArguments
+        if (!buildPath) {
+            buildPath = BuildFlags.extractScratchPath(configuration.packageArguments);
+        }
+
+        // If not in either arguments list, check the buildPath configuration
+        if (!buildPath) {
+            buildPath = configuration.buildPath.length > 0 ? configuration.buildPath : ".build";
+        }
```
plemarquand: nit: could shorten this up to:
```ts
  const buildPath =
      BuildFlags.extractScratchPath(configuration.buildArguments) ??
      BuildFlags.extractScratchPath(configuration.packageArguments) ??
      (configuration.buildPath.length > 0 ? configuration.buildPath : ".build");
```

src/toolchain/BuildFlags.ts
```diff
@@ -128,6 +128,27 @@ export class BuildFlags {
         }
     }
 
+    /**
+     * Extract scratch-path or build-path value from an array of arguments
+     * @param args Array of command-line arguments to search
+     * @returns The path value if found, otherwise undefined
+     */
+    private static extractScratchPath(args: string[]): string | undefined {
+        for (let i = 0; i < args.length; i++) {
```
plemarquand, line 132: Some validation here may be helpful, since a misconfigured path would cause failures downstream. I'd throw an error if:
- The path is ""
- The path isn't a valid path
- `args[i+1]` starts with `--` (i.e. its another flag, a user error like `--build-path --filter`)

plemarquand:
Looks great! Just one comment about validation since we've had issues reported in he past where users have misconfigured arguments, and they're often hard to track down unless we guard against it defensively.

src/toolchain/BuildFlags.ts
kartikangiras: done

kartikangiras:
> Looks great! Just one comment about validation since we've had issues reported in he past where users have misconfigured arguments, and they're often hard to track down unless we guard against it defensively.


@plemarquand I have added the necessary validations which were mentioned.

Issue #2064, opened 2026-01-29
Ignoring `swiftly` install prompt blocks extension activation


rbenegal:
**Describe the bug**
If the user ignores the `swiftly` install prompt the extension may not activate.

**To Reproduce**
1. Create a `.swift-version` file with the contents of something like `6.2.0` for a swift project (such as the defaultProject)
2. Delete swiftly if you have it already with `rm -rf ~/.swiftly`
3. Open the swift project and notice that the extension prompts to install swiftly
4. Ignore the prompt and wait for it to go away (it never actually goes away and can be accessed by clicking the "bell" icon in the bottom right hand corner of the window)
5. Notice that the extension no longer activates

**Expected behavior**
The extension should always activate as the user may already have an Xcode swift toolchain or some other local toolchain installed.

**Environment**
 - OS: macOS 26
 - Visual Studio Code version: 
 - vscode-swift version: 2.16

rbenegal:
Confirmed with https://github.com/swiftlang/vscode-swift/commit/3de454eb6b73787d9e745aea08f8d10ccdcff3bf that the notification prompt no longer blocks extension activation and that the Disable Swiftly Install Prompt works well.

Issue #2063, opened 2026-01-29
Activating Swift extension failed: Reduce of empty array with no initial value


aoberoi:
**Describe the bug**
Clicking on the "Create Swift Project" button in the Explorer pane, first use of the Swift extension, on Windows. Seeing this error:

> Activating Swift extension failed: Reduce of empty array with no initial value
> Source: Swift

**To Reproduce**
1. Install extension
2. Open Explorer sidebar
3. Click "Create Swift Project"
4. Observe the error above.

**Expected behavior**
Extension to activate and prompts about the creation of a new project.

**Environment**
 - OS: Windows 11
 - Swift version
    ```
    Swift version 6.2.3 (swift-6.2.3-RELEASE)
    Target: x86_64-unknown-windows-msvc
    Build config: +assertions
    ```
 - Visual Studio Code version:
    ```
    Version: 1.108.2 (user setup)
    Commit: c9d77990917f3102ada88be140d28b038d1dd7c7
    Date: 2026-01-21T13:52:09.270Z
    Electron: 39.2.7
    ElectronBuildId: 12953945
    Chromium: 142.0.7444.235
    Node.js: 22.21.1
    V8: 14.2.231.21-electron.0
    OS: Windows_NT x64 10.0.22631
    ```
 - vscode-swift version: 2.16.0

**Diagnostics Bundle**
Could not capture diagnostics bundle. `command 'swift.captureDiagnostics' not found`. This is likely because the extension never activated.

**Additional context**

In the Output pane

Swift:
```
2026-01-29 11:01:17.838 [info] Activating Swift for Visual Studio Code 2.16.0...
2026-01-29 11:01:17.840 [info] Log setup completed in 7ms
```

Extension Host (interesting part):
```
2026-01-29 11:01:17.541 [info] ExtensionService#_doActivateExtension swiftlang.swift-vscode, startup: false, activationEvent: 'onCommand:swift.createNewProject'
2026-01-29 11:01:18.648 [error] Activating extension swiftlang.swift-vscode failed due to an error:
2026-01-29 11:01:18.648 [error] TypeError: Reduce of empty array with no initial value
    at Array.reduceRight (<anonymous>)
    at c:\Users\Ankur\.vscode\extensions\swiftlang.swift-vscode-2.16.0\src\extension.ts:358:31
    at LL (c:\Users\Ankur\.vscode\extensions\swiftlang.swift-vscode-2.16.0\src\extension.ts:82:35)
```
[...]

Issue #2065, opened 2026-01-29
Enable installing `swiftly` via Select Toolchain command


rbenegal:
**Is your feature request related to a problem? Please describe.**
Currently the extension only asks if you'd like to install `swiftly` and a toolchain if there is a `.swift-version` file. If we run the `Swift: Select Toolchain` and the user does not have `swiftly` the user is directed to go to the swift.org website to install it.

<img width="603" height="146" alt="Image" src="https://github.com/user-attachments/assets/806e05c6-b550-4f8a-aa52-a5b856411420" />

<img width="276" height="254" alt="Image" src="https://github.com/user-attachments/assets/19b14691-9b57-4da7-aa1e-86ade67f7f4d" />

**Describe the solution you'd like**
The option should allow the user to install `swiftly` directly from VS Code if they are on macOS or Linux.

vanshkamra12:
@rbenegal can i work on thiss...????

PR #2066, opened 2026-01-29
fix extension activation issues involving swiftly

matthewbastien:
## Description
This PR fixes two issues with extension activation that happen in 2.16.0:

- If there are no `.swift-version` files in the workspace then `.reduceRight()` can throw an error. I'm not sure how this made it past CI since none of our test fixtures include swift version files.
- If the user never explicitly dismisses the warning about installing swiftly then activation hangs. However, VS Code will hide the warning on its own making it difficult to see why the extension is not working.

I've added two new ESLint rules to catch future issues with `.reduce()` and `.reduceRight()` and made the swiftly install check non-blocking. I also updated the `.swift-version` file check so that it won't run on Windows which doesn't support Swiftly anyway.

Issue: #2063, #2064

## Tasks

### DIFF
--- a/package.json
+++ b/package.json
@@ -1134,7 +1134,7 @@
             "default": false,
             "markdownDescription": "Disable the automatic Swiftly installation prompt when no Swift toolchain is found.",
             "order": 98,
-            "scope": "application"
+            "scope": "machine-overridable"
           },
           "swift.diagnostics": {
             "type": "boolean",


src/extension.ts
```diff
@@ -344,6 +334,65 @@ function handleFolderEvent(logger: SwiftLogger): (event: FolderEvent) => Promise
...
+    logger: SwiftLogger
+): void {
+    contextKeys.supportsSwiftlyInstall = false;
+    if (!Swiftly.isSupported()) {
+        logger.debug(`Swiftly is not available on ${process.platform}`);
+        return;
+    }
+
+    // Don't block while updating Swiftly context keys.
+    Swiftly.isInstalled().then(
```
aoberoi, line 352: There's now 3 separate checks for `Swiftly.isInstalled()` in this flow: here, one below that guards looking for `.swift-version` files, and one within `handleMissingSwiftly()`.

In my opinion, I think it would be better (avoid race conditions and harder debugging) if there was just one check.

I think we need to think a bit more carefully about the overall flow (including my thoughts in the review).

aoberoi (commented):
I think there's a design issue with the user flow.

This PR helps because it unblocks the creation/activation of the toolchain, and the rest of the extension.

However, why should the user be prompted to install Swiftly if that toolchain is already working? I think the sequence of steps should be:
1. Attempt to create/activate the toolchain (already done in this PR).
2. If there's an issue with the toolchain creation (sync check) or the created toolchain does not match the discovered `.swift-version` files (async check):
  2a) Prompt the user to remedy the situation (currently done in `showToolchainError()`). The potential choices should be (1) select a toolchain (sorted so that the version in `.swift-version` is at the top if it is found) (2) install a toolchain using Swiftly (only presented if the platform supports Swiftly even when Swiftly is not already installed) or (3) to ignore the issue (when there is a toolchain, even though the version doesn't match). (4) Open documentation to install from Swift.org. This would eliminate the "Remove from settings" option that's currently presented. I don't see how that would ever be an acceptable remedy.
  2b) If the user has selected (2). Check if Swiftly is already installed. If it isn't installed and the config doesn't prevent installation, then perform the Swiftly installation. Then trigger the command to install a version using Swiftly.

One state worth highlighting: if the user is on Windows and has no Swift toolchain installed, that means the only remedy that can be presented is (4).

Just wanted to share this idea quickly. I haven't yet looked into what the other calls to `showToolchainError()` expect, so it might need to be broken apart further.

src/extension.ts
matthewbastien: Good catch! I've removed the extra calls.

matthewbastien:
@aoberoi you bring up some good points. Toolchain selection is something that was added early on and definitely needs some attention. Let me try to explain some of our reasoning here.

> why should the user be prompted to install Swiftly if that toolchain is already working?

It's my impression that most projects that use a `.swift-version` file expect you to use swiftly as a toolchain manager (though I could be wrong about that). This scenario was the easiest way to implement the first pass of having VS Code install Swiftly automatically on the user's behalf.

> If there's an issue with the toolchain creation (sync check) or the created toolchain does not match the discovered .swift-version files (async check): Prompt the user to remedy the situation

Yeah this is definitely something we want to put in! I don't think I raised an issue for it yet, but check out #2065 for another spot where we want to add Swiftly installation. I've also got a PR in the works to do this. It just required some refactoring of the Swiftly class so that we're not passing around stuff like `extensionRoot` everywhere. The Swiftly refactoring can be found in #2051.

> to ignore the issue (when there is a toolchain, even though the version doesn't match). This would eliminate the "Remove from settings" option that's currently presented. I don't see how that would ever be an acceptable remedy.

Yeah, the `Remove from Settings` button is something that is less useful now that we've got a proper toolchain selection workflow. It originally existed because it's possible to set `swift.path` to an incorrect value. Something we've gotten a few bug reports about due to various AI tools adding hallucinated settings.

I'm open to discussing this further in a separate issue if you want to raise one. I'm going to limit the scope of this PR to unblocking the extension activation issues.

aoberoi:
Thanks for the added context and consideration! I agree, this fix makes sense to land regardless of the other potential improvements.

Would it be useful for me to think through the motivation for #2065, #2051,  and the issues I raised here - and open an issue with a proposal for discussion? It might be something that can be broken into discrete changes, but I think that's easiest if there's an end goal for what the flow should look like.

matthewbastien:
> Would it be useful for me to think through the motivation for https://github.com/swiftlang/vscode-swift/issues/2065, https://github.com/swiftlang/vscode-swift/pull/2051, and the issues I raised here - and open an issue with a proposal for discussion? It might be something that can be broken into discrete changes, but I think that's easiest if there's an end goal for what the flow should look like.

Go for it! This is definitely a bit of a pain point that we should fix.

rockbruno:
Could we also have a flag to turn off this code path entirely, similarly to the one we have to skip SwiftPM features? We have a .swift-version file on our iOS repo because some linters require it, so it would be good if I could get the extension to ignore it

I'm also having issues having the extension start at all due to this change, not sure what exactly but it might be glob it now runs (we have a massive monorepo). So it would be great me to be able to skip this entirely

matthewbastien:
> Could we also have a flag to turn off this code path entirely, similarly to the one we have to skip SwiftPM features? We have a .swift-version file on our iOS repo because some linters require it, so it would be good if I could get the extension to ignore it

We have a setting to disable the prompt, but you bring up a good point. I'm going to move the setting check to before the `.swift-version` check since there's no point in doing it if it's disabled.

src/extension.ts
```diff
@@ -344,6 +334,69 @@ function handleFolderEvent(logger: SwiftLogger): (event: FolderEvent) => Promise
...
+
+/**
+ * Checks for .swift-version file(s) in the workspace. If any are found then the user will be prompted to install Swiftly.
+ */
+async function checkAndPromptToInstallSwiftly(
+    context: vscode.ExtensionContext,
+    logger: SwiftLogger
+): Promise<void> {
+    // Bail early if the user has disabled the swiftly install prompt
+    if (vscode.workspace.getConfiguration("swift").get("disableSwiftlyInstallPrompt", false)) {
```
rockbruno, line 383: Could we make this setting configurable per-workspace like the SwiftPM one? We do some auto-setup to disable SwiftPM on our repo at the workspace-level so it would be neat to also be able to set this there

Image

matthewbastien: Done.

Issue #2070, opened 2026-01-30
Better handling for incorrect or unavailable swift toolchain in `.swift-version`


rbenegal:
**Is your feature request related to a problem? Please describe.**
If someone types an incorrect or outdated swift toolchain in a `.swift-version` file. Then the `swiftly` install may succeed but the swift toolchain will fail. The user is still prompted to restart vs code to complete the swiftly install. There is a "Failed to install" prompt that shows up in the corner but it is hard to see behind the pop-up:

<img width="2007" height="1228" alt="Image" src="https://github.com/user-attachments/assets/343a7a96-0f65-4e7b-9547-714493237620" />

**Describe the solution you'd like**
Instead of asking the user to restart the extension could show a notification that the `swiftly` install succeeded but also show another notification that the toolchain install failed. This may reduce any confusion from the `Restart VS Code` prompt where the user expects that the reload will have the unavailable toolchain.

thePianoKid:
I can take this one :)

thePianoKid:
After diving into this issue a bit more, there are three things that are annoying about this flow. The PR should fix all three:

1. The installation failed error should be suppressed until restart: it just confuses the user when it's rendered behind the modal. The error should be rendered when VS Code restarts.
2. When the extension is loaded, `swiftly list-available` should be run, and we should check whether the version in `.swift-version` is available. This will make for a better error message.
3. The warning telling the user the swift version is not installed currently blocks the extension.

PR #2078, opened 2026-02-02
Add swiftly to walkthrough

thePianoKid:
## Description
Now that Swiftly support has been integrated into the extension, it makes sense to add it to the welcome walkthrough. This PR adds a new step to the welcome walkthrough that explains what a toolchain is, what Swiftly is, and includes a button users can press to install Swiftly on their system. 

Issue: partially addresses https://github.com/swiftlang/vscode-swift/issues/1561

## Tasks
- [ ] Required tests have been written
- No automated tests were written, but these manual tests were performed:
    - Made sure the "install Swiftly" step did not render when the extension was running on a Windows machine: 
        Screenshot 2026-02-02 at 4 38 26 PM
    - The "install Swiftly" step renders on macOS and Linux
    - The "install" Swiftly step is marked as completed only when Swiftly has been added to PATH
        - ⚠️ You need to restart VSCode for it to release that the PATH has been updated. This might be confusing to users. Thankfully, we [prompt the user to restart their editor](https://github.com/swiftlang/vscode-swift/blob/f7a44aec10bc8f7ae7ac5be7e2fef7f3a08af2d4/src/commands/installSwiftly.ts#L107) once Swiftly is installed. 
- [ ] Documentation has been updated
- [x] Added an entry to CHANGELOG.md if applicable

PR #2080, opened 2026-02-03
Fail when Swift installation fails

thePianoKid:
## Description
Currently, the Swiftly installation flow doesn't know if the Swift install failed — it prompts the user to restart their editor as if nothing happened: 

image

This PR checks if any of the installations failed. If any installation failed, a popup is rendered telling the user to check their Swift version in `.swift-version`. 

Issue: https://github.com/swiftlang/vscode-swift/issues/2070

## Tasks
- Manual tests: 
    - macOS:
        -  Included a wrong Swift version number in `.swift-version`, and confirmed that the "restart VS Code" modal was replaced by the "installation failed" modal. 
        - Included a correct Swift version number in `.swift-version`, and made sure the "restart VS Code" model was still rendered.

### DIFF
--- a/src/commands/installSwiftly.ts
+++ b/src/commands/installSwiftly.ts
@@ -17,7 +17,10 @@ import * as vscode from "vscode";
 import configuration from "../configuration";
 import { SwiftLogger } from "../logging/SwiftLogger";
 import { Swiftly } from "../toolchain/swiftly";
-import { promptToRestartAfterInstallation } from "../ui/RestartEditor";
+import {
+    promptToRestartAfterFailedToolchainInstallation,
+    promptToRestartAfterInstallation,
+} from "../ui/RestartEditor";
 import { installSwiftlyToolchainWithProgress } from "./installSwiftlyToolchain";
 
 /**
@@ -128,11 +131,20 @@ export async function handleMissingSwiftly(
 
     // Install toolchains
     const swiftlyPath = path.join(Swiftly.defaultHomeDir(), "bin/swiftly");
+    const failedVersions: string[] = [];
     for (const version of swiftVersions) {
-        await installSwiftlyToolchainWithProgress(version, extensionRoot, logger, swiftlyPath);
+        await installSwiftlyToolchainWithProgress(version, extensionRoot, logger, swiftlyPath, {
+            onError: () => {
+                failedVersions.push(version);
+            },
+        });
     }
 
     // VS Code needs to be restarted after installing swiftly
-    await promptToRestartAfterInstallation("Swiftly");
+    if (failedVersions.length > 0) {
+        await promptToRestartAfterFailedToolchainInstallation(failedVersions);
+    } else {
+        await promptToRestartAfterInstallation("Swiftly");
+    }
     return true;
 }
--- a/src/ui/ToolchainSelection.ts
+++ b/src/ui/ToolchainSelection.ts
@@ -114,7 +114,7 @@ export async function showMissingToolchainDialog(
             `${folderName}Swift ${version} is not installed`,
             {
                 modal: true,
-                detail: `Swift version ${version} is required but not installed. Would you like to install it using Swiftly?`,
+                detail: `This project's .swift-version requires ${version}, but it is not installed. Would you like to install it using Swiftly?`,
             },
             "Install Toolchain"
         );
@@ -125,7 +125,7 @@ export async function showMissingToolchainDialog(
         `${folderName}Swift ${version} is not available`,
         {
             modal: true,
-            detail: `Swift version ${version} is required but is not available to install via Swiftly.`,
+            detail: `This project's .swift-version requires ${version}, but it is not available to install via Swiftly.`,
         },
         "Select Toolchain"
     );
--- a/src/ui/ToolchainSelection.ts
+++ b/src/ui/ToolchainSelection.ts
@@ -94,10 +94,9 @@ export async function showToolchainError(folder?: vscode.Uri): Promise<boolean> 
  * Handles a required Swift toolchain that is not installed.
  *
  * Checks whether the required version is available to install via Swiftly. If it is, the user is
- * prompted to install it. If it isn't (for example the user relies on a custom toolchain that
- * Swiftly does not manage), the user is offered the option to select a toolchain instead.
+ * prompted to install it. If it isn't, the user is offered the option to select a toolchain instead.
  *
- * @param version The required toolchain version, e.g. from the folder's `.swift-version` file.
+ * @param version The required toolchain version
  * @param logger Logger used while querying the toolchains available via Swiftly.
  * @param folder Optional folder context for the message.
  * @returns Promise<boolean> true if the user chose to install the toolchain via Swiftly, false otherwise.
--- a/src/ui/ToolchainSelection.ts
+++ b/src/ui/ToolchainSelection.ts
@@ -104,7 +104,7 @@ export async function showToolchainError(folder?: vscode.Uri): Promise<boolean> 
  */
 export async function showMissingToolchainDialog(
     version: string,
-    logger?: SwiftLogger,
+    logger: SwiftLogger,
     folder?: vscode.Uri
 ): Promise<boolean> {
     const folderName = folder ? `${FolderContext.uriName(folder)}: ` : "";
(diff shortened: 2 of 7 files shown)


src/commands/installSwiftly.ts
```diff
@@ -142,11 +142,26 @@ export async function handleMissingSwiftly(
 
     // Install toolchains
     const swiftlyPath = path.join(Swiftly.defaultHomeDir(), "bin/swiftly");
+    let installSuccess = true;
```
plemarquand: we should validate that the versions array is not empty, otherwise installSuccess will report true without installing

src/commands/installSwiftly.ts
```diff
@@ -142,11 +142,26 @@ export async function handleMissingSwiftly(
...
     }
 
-    // VS Code needs to be restarted after installing swiftly
-    await promptToRestartVSCode();
-    return true;
+    if (installSuccess) {
+        await promptToRestartVSCode();
+        return true;
+    } else {
+        await vscode.window.showErrorMessage("Installation failed", {
```
plemarquand: Do we get any error message from the installation failure? It would be useful to show the failure message from the install to help the user diagnose why it failed.

src/commands/installSwiftly.ts
thePianoKid: If there is nothing to install, we still want to prompt the user the restart, and `return true`. Which is what happens currently: 

```ts
let installSuccess = true;
    for (const version of swiftVersions) {
        installSuccess =
            installSuccess &&
            (await installSwiftlyToolchainWithProgress(
                version,
                extensionRoot,
                logger,
                swiftlyPath
            ));
    }

    // The for loop is skipped, and the user is prompted to restart
    if (installSuccess) {
        await promptToRestartVSCode();
        return true;
    } 
```

The issue seems to a bad variable name: `promptRestart` might be better.

matthewbastien: Thinking on the original implementation: I think we should, at the very least, install the latest toolchain if no toolchain versions are provided. That's the whole point of installing swiftly, after all.

Thankfully, swiftly makes this easy by letting us do `swiftly install latest`. So, we can just add `"latest"` if the array is empty.

thePianoKid:
I merged the code from #2078 to this PR to avoid code duplication. That means that #2078 should be merged before this one.

thePianoKid:
After some offline conversations with @rbenegal and @matthewbastien, we decided to split up the swift installation flow. 

On `main`, the extension installs swiftly and swift in one step. This PR splits up this flow: now, the user will install Swiftly, then restart VS Code. Once the editor has been restarted, the extension prompts the user to install the Swift toolchain. 

This is less convenient than the old flow, but I think it's better because Swiftly has been added to PATH, so even if the toolchain installation fails, Swiftly is correctly installed.

Here's a video demo of how this works: 


https://github.com/user-attachments/assets/fd28b196-1e98-4095-864d-f21d33314f2f

test/unit-tests/utilities/workspace.test.ts
```diff
@@ -227,4 +234,88 @@ suite("Workspace Utilities Unit Test Suite", () => {
             expect(await isValidWorkspaceFolder("/project", false, testSwiftVersion)).to.be.true;
         });
     });
+
+    suite("findSwiftVersionFiles", () => {
+        const globDirectoryStub = mockGlobalFunction(filesystem, "globDirectory");
+        const workspaceFolders = mockGlobalValue(vscode.workspace, "workspaceFolders");
```
thePianoKid: These are slower integration tests. I think they are worth adding: give define what this new feature does can give pretty good signal. 

That said, if you feel this slows CI down too much, lmk :)

src/commands/installSwiftly.ts
```diff
@@ -100,42 +101,106 @@ async function installSwiftlyWithProgress(logger?: SwiftLogger): Promise<boolean
...
+            logger,
+            swiftlyPath
+        );
+
+        if (result.success) {
+            continue;
+        }
+
+        if (result.errorMsg) {
+            await vscode.window.showErrorMessage("Installation failed", {
```
plemarquand: `installSwiftlyToolchainWithProgressAndErrorMsgs` already performs `vscode.window.showErrorMessage` when there is an errorMsg, so we'll get two similar dialogs in a row on a failed installation

package.json
```diff
@@ -242,6 +272,11 @@
       }
     ],
     "commands": [
+      {
+        "command": "swift.checkInstallation",
```
plemarquand: I can't find a usage of this command

src/commands/createNewProject.ts
```diff
@@ -102,12 +102,15 @@ async function executeOpenAction(
...
+    toolchain: SwiftToolchain | undefined
+): Promise<void> {
     // It is possible for this command to be run without a valid toolchain because it can be
     // run before the Swift extension is activated. Show the toolchain error notification in
     // this case.
     if (!toolchain) {
-        void showToolchainError();
+        void showToolchainError(extensionPath);
         return;
     }
```
matthewbastien, line 112: Not caused by your change, but this logic no longer works because `registerCommand()` will always pass in a toolchain. I tried locally with my Swift toolchain in a bad state and got an unhelpful error message.

I think I'll raise a separate issue to track this because it affects all commands (and debugging). At the very least we can just remove this code, remove the extensionPath from the arguments, and make toolchain a required argument.

src/ui/ToolchainSelection.ts
```diff
@@ -79,13 +101,17 @@ export async function showToolchainError(folder?: vscode.Uri): Promise<boolean>
...
-    } else if (selected === "Open Documentation") {
-        void vscode.env.openExternal(
-            vscode.Uri.parse(
-                "https://docs.swift.org/vscode/documentation/userdocs/supported-toolchains"
-            )
-        );
-        return false;
+    } else if (selected === "Install Swiftly") {
+        // Only offered when Swiftly is not installed.
+        await installSwiftly({ skipPrompt: true });
```
matthewbastien: Should we also be running `installMissingToolchains()` here? The `Swiftly` class has the option to pass in the path to the swiftly binary directly. This would make it so that we don't have to prompt the user twice. You can take a look at how `handleMissingSwiftly()` worked before your change.

src/commands/installSwiftly.ts
```diff
@@ -100,42 +101,106 @@ async function installSwiftlyWithProgress(logger?: SwiftLogger): Promise<boolean
     }
 }
 
+async function promptToRestartVSCode(): Promise<void> {
+    const editorName = vscode.env.appName;
+    const selection = await vscode.window.showInformationMessage(
+        `Restart ${editorName}`,
+        {
+            modal: true,
+            detail: `You must restart ${editorName} in order for the Swiftly installation to take effect. \n\nWhen you reopen ${editorName}, you will be prompted to install the Swift toolchain now that you have Swiftly installed.`,
```
matthewbastien: If you change it so that installing swiftly always installs at least one toolchain then we can drop the second half of the prompt.
```suggestion
            detail: `You must restart ${editorName} in order for the Swiftly installation to take effect.`,
```

matthewbastien (changes requested):

src/ui/ToolchainSelection.ts
```diff
@@ -91,22 +91,65 @@ export async function showToolchainError(folder?: vscode.Uri): Promise<boolean>
...
+    const availableToolchains = await Swiftly.listAvailable(branch, logger);
+    return availableToolchains.some(toolchain => toolchain.version.name === version);
+}
+
+/**
+ * Derives the Swiftly snapshot branch (e.g. "main-snapshot", "6.1-snapshot") from a snapshot
+ * toolchain version. Returns undefined for stable releases, which are listed without a branch.
+ */
+function swiftlySnapshotBranch(version: string): string | undefined {
+    return /^(.+-snapshot)-\d{4}-\d{2}-\d{2}$/.exec(version)?.[1];
```
thePianoKid, line 159: This function will never be called because of a bug where snapshot versions are treated as invalid versions by VS Code (https://github.com/swiftlang/vscode-swift/issues/2308). 

I still included it so once this issue is fixed, everything works.

matthewbastien (approved):

src/commands/createNewProject.ts
thePianoKid: @matthewbastien could you link to the issue?

PR #2086, opened 2026-02-05
Always notify user when an unexpected setting value is found

plemarquand:
## Description
We started validating that setting values were of the expected type in da0e04c however the user was only notified with a popup if the malformed setting was used during extension activation.

A user recently encountered an issue with a malfored `swift.sourcekit-lsp.backgroundIndexing` setting. Because sourcekit-lsp is set up so its creation doesn't block extension activation, this error never bubbled up to the activation error handler, and they were not notified.

Instead notify the user immediately when we detect the malformed setting. The malformed value is cached so we only notify the first time we detect it, avoiding spamming the user with prompts.

## Tasks

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -799,7 +799,7 @@ function validateObjectSetting<T extends object>(obj: T, settingName: string): T
     return obj;
 }
 
-// To avoid spanning the user with multiple error messages for the same bad setting/value
+// To avoid spamming the user with multiple error messages for the same bad setting/value
 // we keep track of the bad settings we've already reported during this session.
 const badSettingLookup: { [key: string]: unknown } = {};
 


src/configuration.ts
```diff
@@ -778,14 +788,40 @@ function validateStringArraySettings(arr: string[], settingName: string): string
...
+                settingName,
+                obj,
+                `The setting \`${settingName}\` must be an object`
+            )
         );
     }
     return obj;
 }
 
+// To avoid spanning the user with multiple error messages for the same bad setting/value
```
matthewbastien: **nit:** spelling
```suggestion
// To avoid spamming the user with multiple error messages for the same bad setting/value
```

PR #2096, opened 2026-02-10
Enable SonarJS ESLint rules

plemarquand:
## Description
Configure SonarJS and use the recommended rule set. SonarJS is a set of static analysis rules for code quality and consistency.

There are several rules that are disabled that we should progressively enable over time (denoted with a comment in the `.eslintrc.json`). 

## Tasks

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "swift-vscode",
   "version": "2.18.0-dev",
-  "lockfileVersion": 3,
+  "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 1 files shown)


test/unit-tests/testexplorer/TestCodeLensProvider.test.ts
```diff
@@ -146,15 +146,7 @@ suite("TestCodeLensProvider", () => {
...
         );
-        expect(testItemIds).to.include.members([
-            "test1",
-            "test1",
-            "test1",
-            "test3",
-            "test3",
-            "test3",
-        ]);
-        expect(testItemIds).to.not.include.members(["test2", "test4"]);
```
rbenegal, line 157: Is this expectation supposed to be ignored? Seems like it was trying to check for something.

plemarquand: The new expectation on line 149 is now covers both of these more explicitly. It should exactly match the array provided there.

rbenegal (approved):

Issue #1965, opened 2025-11-28
Add tests for Playgrounds in the ProjectPanelProvider


award999:
Write a reasonable set of tests to verify #1782. A lot of tests were written between #2371 and #1924 but until "swift-play" is in the toolchain, tests cannot be added to for the ProjectPanelProvider integration. Will likely need to add so #Playground macro usages to the `assets/test/defaultPackage` project.

Flamki:
Opened PR with the ProjectPanelProvider playground integration coverage requested here: https://github.com/swiftlang/vscode-swift/pull/2115

PR #2115, opened 2026-02-20
Add playground integration tests for ProjectPanelProvider

Flamki:
## Description
Add integration test coverage for playground entries in `ProjectPanelProvider` as requested in #1965.

This adds a dedicated `Playgrounds` test suite to `ProjectPanelProvider.test.ts` that verifies:
- playground nodes are shown under a `Playgrounds` header when playgrounds are available
- label fallback behavior (`label` preferred, otherwise `id`)
- expected tree item behavior for playground nodes (`contextValue` + open command)
- the `Playgrounds` header is removed when no playgrounds are returned
- existing `folderContext.playgroundProvider` is restored after each test to avoid state leakage/flakiness

Issue: https://github.com/swiftlang/vscode-swift/issues/1965

## Tasks
- [x] Documentation has been updated (N/A for test-only change)
- [x] Added an entry to CHANGELOG.md if applicable (N/A for test-only change)

## Validation
- `npm run lint` (pass)
- `npm run format` (pass)
- Integration tests were started locally but intentionally stopped due runtime duration; CI will run them on PR.

### DIFF
--- a/test/integration-tests/ui/ProjectPanelProvider.test.ts
+++ b/test/integration-tests/ui/ProjectPanelProvider.test.ts
@@ -19,6 +19,7 @@ import * as vscode from "vscode";
 import { FolderContext } from "@src/FolderContext";
 import { WorkspaceContext } from "@src/WorkspaceContext";
 import { Commands } from "@src/commands";
+import { PlaygroundProvider } from "@src/playgrounds/PlaygroundProvider";
 import { Playground } from "@src/sourcekit-lsp/extensions";
 import { createBuildAllTask } from "@src/tasks/SwiftTaskProvider";
 import {
@@ -44,33 +45,13 @@ tag("medium").suite("ProjectPanelProvider Test Suite", function () {
     let folderContext: FolderContext;
     let treeProvider: ProjectPanelProvider;
 
-    type PlaygroundChangeEvent = {
-        uri: string;
-        playgrounds: Playground[];
-    };
-
-    class MockPlaygroundProvider {
-        private didChangePlaygroundsEmitter: vscode.EventEmitter<PlaygroundChangeEvent> =
-            new vscode.EventEmitter();
-        private workspacePlaygrounds: Playground[] = [];
-
-        onDidChangePlaygrounds: vscode.Event<PlaygroundChangeEvent> =
-            this.didChangePlaygroundsEmitter.event;
-
-        async getWorkspacePlaygrounds(): Promise<Playground[]> {
-            return this.workspacePlaygrounds;
-        }
-
+    class MockPlaygroundProvider extends PlaygroundProvider {
         setPlaygrounds(playgrounds: Playground[]) {
-            this.workspacePlaygrounds = playgrounds;
+            this.setWorkspacePlaygrounds(playgrounds);
             this.didChangePlaygroundsEmitter.fire({
                 uri: playgrounds[0]?.location.uri ?? "",
                 playgrounds,
             });
-        }
-
-        dispose() {
-            this.didChangePlaygroundsEmitter.dispose();
         }
     }
 
@@ -166,9 +147,8 @@ tag("medium").suite("ProjectPanelProvider Test Suite", function () {
 
         beforeEach(() => {
             existingPlaygroundProvider = folderContext.playgroundProvider;
-            mockPlaygroundProvider = new MockPlaygroundProvider();
-            folderContext.playgroundProvider =
-                mockPlaygroundProvider as unknown as FolderContext["playgroundProvider"];
+            mockPlaygroundProvider = new MockPlaygroundProvider(folderContext);
+            folderContext.playgroundProvider = mockPlaygroundProvider;
             treeProvider.observeFolder(folderContext);
         });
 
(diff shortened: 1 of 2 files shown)


Flamki:
FYI on the current failing check (`Test / Windows (6.0 - windows-2022)`): from the job snippet it looks like the test suite itself completed successfully (`440 passing`, extension host exit code `0`) and then the CI wrapper exited with `FAILED (1)` in post-processing.

I revalidated this branch locally (`npm run format`, `npm run lint`) and there are no pending code changes needed on the PR branch.

This appears consistent with an infra/wrapper flake rather than a deterministic code regression in this PR. Once GitHub connectivity is stable on my side again, I can re-run just the failed job if needed.

test/integration-tests/ui/ProjectPanelProvider.test.ts
```diff
@@ -114,6 +160,77 @@ tag("medium").suite("ProjectPanelProvider Test Suite", function () {
...
     });
 
+    suite("Playgrounds", () => {
+        let mockPlaygroundProvider: MockPlaygroundProvider | undefined;
+        let existingPlaygroundProvider: FolderContext["playgroundProvider"];
+
+        beforeEach(() => {
+            existingPlaygroundProvider = folderContext.playgroundProvider;
+            mockPlaygroundProvider = new MockPlaygroundProvider();
+            folderContext.playgroundProvider =
```
award999: we should try to find a way to fake out the `playgroundProvider` so we can avoid the typecasting and manual teardown. I see a few options, message me if you want some suggestions :)

test/integration-tests/ui/ProjectPanelProvider.test.ts
```diff
@@ -40,13 +41,58 @@ import {
...
     let workspaceContext: WorkspaceContext;
+    let folderContext: FolderContext;
     let treeProvider: ProjectPanelProvider;
 
+    type PlaygroundChangeEvent = {
+        uri: string;
+        playgrounds: Playground[];
+    };
+
+    class MockPlaygroundProvider {
```
award999: You can extend PlaygroundProvider class and make didChangePlaygroundsEmitter protected. The super-class can reset the dictionary of playgrounds with the new values.

test/integration-tests/ui/ProjectPanelProvider.test.ts
```diff
@@ -114,6 +160,77 @@ tag("medium").suite("ProjectPanelProvider Test Suite", function () {
         );
     });
 
+    suite("Playgrounds", () => {
```
award999, line 144: these 2 test cases are a great set to include in this suite. Similar to tasks, can we also have a case to make sure the `command` of the `PlaygroundNode` is hooked up correctly?

test/integration-tests/ui/ProjectPanelProvider.test.ts
Flamki: Addressed in 9a75e13. I replaced the ad-hoc mock with `MockPlaygroundProvider extends PlaygroundProvider` and removed the unsafe cast/manual mock wiring. The test now uses a real provider subtype and direct assignment to `folderContext.playgroundProvider`.

test/integration-tests/ui/ProjectPanelProvider.test.ts
Flamki: Added in 9a75e13. New test `Hooks up playground command` verifies `PlaygroundNode.toTreeItem()` wiring: `title = Open Playground`, `command = vscode.openWith`, and expected URI/editor args including selection payload.

test/integration-tests/ui/ProjectPanelProvider.test.ts
Flamki: Implemented as suggested in 9a75e13. I changed `didChangePlaygroundsEmitter` to `protected` in `PlaygroundProvider` and added `setWorkspacePlaygrounds(...)` so subclasses/tests can reset grouped workspace playground state through the superclass path.

Flamki:
@award999 I addressed all requested changes in 9a75e13 (provider subclass approach, protected emitter + superclass reset hook, and explicit playground command wiring test). Could you please take another look when you have a moment?

test/integration-tests/ui/ProjectPanelProvider.test.ts
```diff
@@ -40,13 +42,38 @@ import {
 
 tag("medium").suite("ProjectPanelProvider Test Suite", function () {
     let workspaceContext: WorkspaceContext;
+    let folderContext: FolderContext;
     let treeProvider: ProjectPanelProvider;
 
+    class MockPlaygroundProvider extends PlaygroundProvider {
+        setPlaygrounds(playgrounds: Playground[]) {
+            this.setWorkspacePlaygrounds(playgrounds);
+            this.didChangePlaygroundsEmitter.fire({
```
award999, line 51: Since `setWorkspacePlaygrounds` handles the bulk of the logic, you can fire the event in the `setWorkspacePlaygrounds` then you don't need a subclass and this potentially error prone setup/teardown significantly simplifying the test setup

test/integration-tests/ui/ProjectPanelProvider.test.ts
```diff
@@ -114,6 +141,106 @@ tag("medium").suite("ProjectPanelProvider Test Suite", function () {
         );
     });
 
+    suite("Playgrounds", () => {
+        let mockPlaygroundProvider: MockPlaygroundProvider | undefined;
+        let existingPlaygroundProvider: FolderContext["playgroundProvider"];
```
award999, line 146: should use an explicit type

PR #2140, opened 2026-03-06
Add "Review Script" option to post-install dialog (#2139)

xlmtvv:
## Description

Adds a "Review Script" button to the Swiftly post-install confirmation dialog on Linux, so users can inspect the script before executing it.

When a post-install script is detected after toolchain installation:
1. The dialog now shows two new options: "Execute Script" and "Review Script" (Cancel is VS Code's built-in modal dismiss)
2. Clicking "Review Script" opens the script in a preview tab as a shellscript document
3. The confirmation dialog re-appears, letting the user execute or cancel

Changes in `src/toolchain/swiftly.ts`:
- `showPostInstallConfirmation()` returns the raw button choice (`string | undefined`) instead of a boolean
- `handlePostInstallFile()` handles the new review flow
- New `showPostInstallScript()` method opens the script as an untitled document

Issue: https://github.com/swiftlang/vscode-swift/issues/2139

## Tasks

### Testing

Added 2 unit tests in `test/unit-tests/toolchain/swiftly.test.ts`:
- Review then Execute: user clicks "Review Script", sees the script, then clicks "Execute Script"
- Review then Cancel: user clicks "Review Script", sees the script, then dismisses the dialog
- Multiple reviews: user clicks "Review Script" twice before executing (tests the while loop)

All existing post-install tests pass without changes.

Results:
- macOS: 459 passing, 0 failing
- Linux ARM64 (Dev Container, xvfb-run): 448 passing, 0 failing, 11 pending (macOS-only Xcode Watcher tests)
- `npm run lint` clean

image

image

### DIFF
--- a/src/toolchain/swiftly.ts
+++ b/src/toolchain/swiftly.ts
@@ -892,9 +892,14 @@ export class Swiftly {
             return;
         }
 
-        const shouldExecute = await this.showPostInstallConfirmation(version, validation, logger);
+        let choice = await this.showPostInstallConfirmation(version, validation, logger);
 
-        if (shouldExecute) {
+        while (choice === "Review Script") {
+            await this.showPostInstallScript(postInstallFilePath);
+            choice = await this.showPostInstallConfirmation(version, validation, logger);
+        }
+
+        if (choice === "Execute Script") {
             await this.executePostInstallScript(
                 postInstallFilePath,
                 version,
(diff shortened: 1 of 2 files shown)


matthewbastien (changes requested):
My main issue with this is that the dialog is modal which means that it blocks and dims part of the screen after opening the script.

The contents of the script are expected to be pretty short. Why not just simplify this by putting the full contents of the script into the modal dialog? That way we don't even need a review button at all.

matthewbastien:
Closing due to inactivity.

Issue #2029, opened 2026-01-07
Use `swiftly run +TOOLCHAIN_ID` when using swiftly


MahdiBM:
**Is your feature request related to a problem? Please describe.**
See this issue: https://github.com/swiftlang/swift-package-manager/issues/9562

**Describe the solution you'd like**
Do e.g. `swiftly run +6.3-snapshot-2026-01-03 swift test`
instead of `.../Library/Developer/Toolchains/swift-6.3-DEVELOPMENT-SNAPSHOT-2026-01-03-a.xctoolchain/usr/bin/swift test` when using swiftly toolchains.

This also fixes a bug, other than being better human readable.

**Describe alternatives you've considered**
-

**Additional context**
-

MahdiBM:
I guess this should be done behind a feature flag of some sort?

I'm worried this might fix 1 thing and break others. Hopefully not though.

matthewbastien:
The Swift extension really shouldn't be side-stepping swiftly like this, in my opinion. I think it's best if we were to just use the `swift` binary in `PATH` if we know that it comes from swiftly. I'm working out how big of a change this would be to implement. A lot of our logic uses the binary from the toolchain directly.

roulpriya:
@matthewbastien We can refactor toolchain command launching so that whenever the active toolchain is managed by swiftly, every toolchain executable is invoked through swiftly run swift ... instead of being executed directly by path.

I can pick this up.

matthewbastien:
Yeah that's what I wanted to do for this. If you want to take a stab at it then go for it!

matthewbastien:
Verified with cdf412eb

When swiftly is detected, the Swift extension will now use `swiftly run` to launch executables from the toolchain instead of executing them directtly.

PR #2177, opened 2026-03-29
Invoke toolchain through swiftly

roulpriya:
## Description
Swiftly-managed toolchains cannot always be launched correctly by directly executing binaries from the toolchain path.
But for toolchains managed by swiftly, those tools should be invoked through

`swiftly run <tool>`

Changes:

Updated the extension to use this new invocation flow across the places where it launches toolchain tools:

- Swift package commands and tasks now run through swiftly run swift ... when needed.
- SourceKit-LSP now launches through swiftly run sourcekit-lsp for swiftly toolchains.
- Debug/test configurations now use the correct swiftly run swift ... command shape.
- LLDB library lookup now invokes lldb via the toolchain invocation abstraction.

Issue: #2029 

## Tasks


assets/test/.vscode/settings.json
```diff
@@ -17,5 +17,6 @@
     },
     "swift.packageArguments": [
         "--cache-path=${workspaceFolder}${pathSeparator}.spm-cache"
-    ]
+    ],
+    "swift.path": "/Users/priyaroul/Library/Developer/Toolchains/swift-latest.xctoolchain/usr/bin"
```
matthewbastien: Looks like this was included accidentally.

src/toolchain/toolchain.ts
```diff
@@ -373,12 +386,31 @@ export class SwiftToolchain implements ExternalSwiftToolchain {
...
 
     /**
-     * Return fullpath for toolchain executable
+     * Return fullpath for toolchain executable.
+     * Use this only when a raw filesystem path is needed (diagnostics,
+     * `--toolchain` arguments, existence checks). For process spawning use
+     * {@link getToolchainInvocation} instead.
      */
-    public getToolchainExecutable(executable: string): string {
+    public getToolchainExecutablePath(executable: string): string {
```
matthewbastien, line 394: We should merge this into `getToolchainInvocation()`. I took a look at the remaining places where this method is still used in the code base and we should be using `getToolchainInvocation()` for those as well.

src/utilities/utilities.ts
```diff
@@ -256,20 +256,25 @@ export async function execSwift(
...
     } else if (toolchain === "default") {
-        swift = getSwiftExecutable();
+        command = getSwiftExecutable();
+        commandArgs = args;
     } else {
-        swift = toolchain.getToolchainExecutable("swift");
         args = toolchain.buildFlags.withAdditionalFlags(args);
+        const inv = toolchain.getToolchainInvocation("swift", args);
+        command = inv.command;
+        commandArgs = inv.args;
```
matthewbastien, line 271: Can't all of this just be replaced with `toolchain.getToolchainInvocation()`?

matthewbastien (changes requested):
Thank you for putting this together! I just ran into the issue that this resolves today and your PR fixed it.

There's just a few things to fix up before we can merge.

matthewbastien:
Oh and you'll need to add a changelog entry for this as well.

src/toolchain/toolchain.ts
roulpriya: Currently, the code in both places uses that path. How should I use `getToolchainInvocation()` there?

matthewbastien: The `LanguageClientManager` is just launching `sourcekit-lsp` and `TestingConfigurationFactory` is just launching `swift`. Both of those can be done with the new `getToolchainInvocation()`. Technically we could also extend this to the code that launches `lldb-dap`, but I don't think that's super necessary right now.

src/utilities/utilities.ts
roulpriya: I could replace it partially, but there is one case where we pass swift path explicitly

matthewbastien:
FYI: I pushed a change yesterday to the `CHANGELOG.md` because 2.16.3 was released. You'll have to rebase and then move your changelog line up to the new `### Fixed` heading.

roulpriya:
@matthewbastien I have updated the changelog and addressed most of the comments.

roulpriya:
Can we trigger the CI?

src/debugger/debugAdapterFactory.ts
```diff
@@ -64,12 +64,34 @@ export function registerDebugger(workspaceContext: WorkspaceContext): vscode.Dis
...
+ * Only invoked for swiftly-managed toolchains, where the session type is left as
+ * "swift" (instead of being rewritten to "lldb-dap"). Launches the adapter via
+ * `swiftly run lldb-dap` so the correct toolchain version is used.
+ */
+export class SwiftDebugAdapterDescriptorFactory implements vscode.DebugAdapterDescriptorFactory {
+    createDebugAdapterDescriptor(
+        _session: vscode.DebugSession,
+        _executable: vscode.DebugAdapterExecutable | undefined
+    ): vscode.ProviderResult<vscode.DebugAdapterDescriptor> {
+        return new vscode.DebugAdapterExecutable("swiftly", ["run", "lldb-dap"]);
```
matthewbastien: We should stick to using the `lldb-dap` launch type instead of rolling our own. The LLDB DAP extension lets us set the runtime executable and arguments via the `debugAdapterExecutable` and `debugAdapterArgs` configuration options which we already do in the `configureLLDBDap()` function in this file.

src/debugger/debugAdapterFactory.ts
```diff
@@ -284,6 +306,14 @@ export class LLDBDebugConfigurationProvider implements vscode.DebugConfiguration
...
         }
 
+        if (
+            toolchain.manager === "swiftly" &&
+            configuration.debugger.customDebugAdapterPath.length === 0
+        ) {
+            launchConfig.type = SWIFT_LAUNCH_CONFIG_TYPE;
+            return true;
+        }
+
```
matthewbastien:
```suggestion
```

matthewbastien: Delete this as per the comment on line 91.

matthewbastien (changes requested):

src/commands/captureDiagnostics.ts
```diff
@@ -367,10 +367,18 @@ async function sourcekitDiagnose(ctx: FolderContext, dir: string) {
...
+    let diagCommand: string;
+    let diagPrefixArgs: string[];
+    if (serverPathConfig.length > 0) {
+        diagCommand = serverPathConfig;
+        diagPrefixArgs = [];
+    } else {
+        const inv = ctx.toolchain.getToolchainInvocation("sourcekit-lsp", []);
+        diagCommand = inv.command;
+        diagPrefixArgs = inv.args;
+    }
```
matthewbastien:
```suggestion
    const inv: ToolchainInvocation = serverPathConfig
        ? { command: serverPathConfig, args: [] }
        : ctx.toolchain.getToolchainInvocation("sourcekit-lsp", []);
```
This can be made a little easier to read. You'll also have to change the code below to use `inv.command` and `inv.args`.

src/debugger/debugAdapterFactory.ts
```diff
@@ -276,6 +276,15 @@ export class LLDBDebugConfigurationProvider implements vscode.DebugConfiguration
...
 
+        if (
+            toolchain.manager === "swiftly" &&
+            configuration.debugger.customDebugAdapterPath.length === 0
+        ) {
+            launchConfig.debugAdapterExecutable = "swiftly";
+            launchConfig.debugAdapterArgs = ["run", "lldb-dap"];
+            return true;
+        }
+
```
matthewbastien: I'd rather leave the logic for this in `getToolchainInvocation()` rather than have multiple places know how to handle swiftly toolchains.

The code in `DebugAdapter.getLLDBDebugAdapterPath()` can just be moved into this method since it's not used anywhere else. We only need to check for `fileExists()` if the user has explicitly set their own path to `lldb-dap` in settings. So, this can all be simplified to two steps:
1. Check for a custom lldb-dap path in settings (and make sure it exists)
2. Use toolchain.getToolchainInvocation() if no custom path exists

src/debugger/lldb.ts
```diff
@@ -46,17 +46,17 @@ export function updateLaunchConfigForCI(
  * @returns Library path for LLDB
  */
 export async function getLLDBLibPath(toolchain: SwiftToolchain): Promise<Result<string>> {
-    let executable: string;
     try {
-        executable = await toolchain.getLLDB();
+        await toolchain.getLLDB();
     } catch (error) {
         return Result.makeFailure(error);
     }
```
matthewbastien: This code is no longer necessary because you're using `getToolchainInvocation()` below. We can just remove `getLLDB()` from `SwiftToolchain` along with the other custom executable finders.

src/debugger/debugAdapterFactory.ts
```diff
@@ -276,6 +276,15 @@ export class LLDBDebugConfigurationProvider implements vscode.DebugConfiguration
             launchConfig.env = this.convertEnvironmentVariables(launchConfig.env);
         }
 
+        if (
+            toolchain.manager === "swiftly" &&
+            configuration.debugger.customDebugAdapterPath.length === 0
```
matthewbastien:
```suggestion
            !configuration.debugger.customDebugAdapterPath
```
More often than not `customDebugAdapterPath` will be `undefined` and fail this check.

src/sourcekit-lsp/LanguageClientManager.ts
```diff
@@ -420,9 +420,20 @@ export class LanguageClientManager implements vscode.Disposable {
...
+        let serverCommand: string;
+        let serverPrefixArgs: string[];
+        if (serverPathConfig.length > 0) {
+            serverCommand = serverPathConfig;
+            serverPrefixArgs = [];
+        } else {
+            const inv = toolchain.getToolchainInvocation("sourcekit-lsp", []);
+            serverCommand = inv.command;
+            serverPrefixArgs = inv.args;
+        }
```
matthewbastien:
```suggestion
        const inv: ToolchainInvocation = serverPathConfig
            ? { command: serverPathConfig, args: [] }
            : toolchain.getToolchainInvocation("sourcekit-lsp", []);
```
This can be made more readable. You'll also have to use `inv.command` and `inv.args` below.

src/utilities/utilities.ts
matthewbastien: My bad, didn't notice that `toolchain` had multiple possible types in it. Good call to get rid of `"default"` now that it's no longer used 👍

matthewbastien (changes requested):

matthewbastien (approved):
LGTM. Thank you so much for fixing this!

PR #2191, opened 2026-04-10
Add per-target debug session support for swiftbuild build system

plemarquand:
## Description
Swift 6.4.0+ defaults to the 'swiftbuild' build system, which produces one test binary per test target instead of one per package. Modify debugSession to create per-target debug configurations when using the swiftbuild build system.

## Tasks

### DIFF
--- a/src/TestExplorer/TestRunner.ts
+++ b/src/TestExplorer/TestRunner.ts
@@ -18,6 +18,7 @@ import * as stream from "stream";
 import * as vscode from "vscode";
 
 import { FolderContext } from "../FolderContext";
+import { TargetType } from "../SwiftPackage";
 import { WorkspaceContext } from "../WorkspaceContext";
 import configuration from "../configuration";
 import {
@@ -25,6 +26,8 @@ import {
     SwiftTestingBuildAguments,
     SwiftTestingConfigurationSetup,
     TestingConfigurationFactory,
+    effectiveBuildSystem,
+    groupTestsByTarget,
 } from "../debugger/buildConfig";
 import { LoggingDebugAdapterTracker } from "../debugger/logTracker";
 import { createSwiftTask } from "../tasks/SwiftTaskProvider";
@@ -701,13 +704,19 @@ export class TestRunner {
         const fifoPipePath = this.generateFifoPipePath(testRunTime);
 
         await TemporaryFolder.withNamedTemporaryFiles([fifoPipePath], async () => {
+            const allBuildArgs = [
+                ...configuration.buildArguments,
+                ...configuration.folder(this.folderContext.workspaceFolder).additionalTestArguments,
+            ];
+            const buildSystem = effectiveBuildSystem(this.folderContext.swiftVersion, allBuildArgs);
+
             if (this.testArgs.hasSwiftTestingTests) {
                 // macOS/Linux require us to create the named pipe before we use it.
                 // Windows just lets us communicate by specifying a pipe path without any ceremony.
                 if (process.platform !== "win32") {
                     await execFile("mkfifo", [fifoPipePath], undefined, this.folderContext);
                 }
-                // Create the swift-testing configuration JSON file, peparing any
+                // Create the swift-testing configuration JSON file, preparing any
                 // directories the configuration may require.
                 const attachmentFolder = await SwiftTestingConfigurationSetup.setupAttachmentFolder(
                     this.folderContext,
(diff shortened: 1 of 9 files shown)


src/debugger/buildConfig.ts
```diff
@@ -29,6 +29,33 @@ import { Version } from "../utilities/version";
 import { SWIFT_LAUNCH_CONFIG_TYPE } from "./debugAdapter";
 import { updateLaunchConfigForCI } from "./lldb";
 
+type BuildSystem = "native" | "swiftbuild";
+
+export function effectiveBuildSystem(
+    swiftVersion: Version,
+    buildArgs: readonly string[]
+): BuildSystem {
+    const buildSystemIndex = buildArgs.lastIndexOf("--build-system");
```
award999: technically `--build-system=swiftbuild` would be valid too

src/debugger/buildConfig.ts
```diff
@@ -29,6 +29,33 @@ import { Version } from "../utilities/version";
 import { SWIFT_LAUNCH_CONFIG_TYPE } from "./debugAdapter";
 import { updateLaunchConfigForCI } from "./lldb";
 
+type BuildSystem = "native" | "swiftbuild";
+
+export function effectiveBuildSystem(
```
award999, line 45: could we extract this into a common function in `BuildFlags` that finds some expected flag? This situation will likely come up again

src/TestExplorer/TestRunner.ts
```diff
@@ -719,63 +727,20 @@ export class TestRunner {
...
-                        const configJSON = JSON.stringify(xcTestBuildConfig);
-                        this.workspaceContext.logger.debug(
-                            `XCTest Debug Config: ${configJSON}`,
-                            this.folderContext.name
-                        );
-                    }
-
-                    buildConfigs.push(xcTestBuildConfig);
-                }
+                const configs =
```
award999, line 739: just want to confirm, we already have logic in place that stopping the current debug session stops any subsequent ones from continuing?

src/TestExplorer/TestRunner.ts
```diff
@@ -800,10 +765,161 @@ export class TestRunner {
...
+
+        // If we're testing in both frameworks we're going to start more than one debugging
+        // session. If both build configurations have the same name LLDB will replace the
+        // output of the first one in the Debug Console with the output of the second one.
+        // If they each have a unique name the Debug Console gets a nice dropdown the user
+        // can switch between to see the output for both sessions.
+        swiftTestBuildConfig.name = `Swift Testing: ${swiftTestBuildConfig.name}`;
+
+        // output test build configuration
+        if (configuration.diagnostics) {
```
award999: We should deprecate `diagnostics`, `SwiftLogger` gives precedence to `outputChannelLogLevel` set to `debug`

src/TestExplorer/TestRunner.ts
plemarquand: Yep, the `startDebugSession` method will check `testRun.isCancellationRequested` first and exit early if the user stops a session. We still iterate all the build configs, but do nothing.

src/TestExplorer/TestRunner.ts
plemarquand: I've removed these guards. If `configuration.diagnostics` is true the logger is configured to print at the `debug` level already

src/debugger/buildConfig.ts
plemarquand: Done

src/debugger/buildConfig.ts
plemarquand: Good catch, done

matthewbastien (approved):

src/toolchain/BuildFlags.ts
```diff
@@ -392,4 +392,38 @@ export class BuildFlags {
...
+     * supporting both `--flag value` and `--flag=value` formats.
+     *
+     * @param args The argument list to search
+     * @param flag The flag name to look for (e.g. `"--build-system"`)
+     * @param validate Optional type guard to restrict which values are accepted.
+     *                 When provided, occurrences whose value fails validation are
+     *                 skipped and the search continues to the next match.
+     * @returns The value of the last valid occurrence, or `undefined` if none match
+     */
+    static lastFlagValue<T extends string = string>(
```
award999, line 407: would love a few tests for this

award999 (commented):
Would be great if we could add some tests for `lastFlagValue` although I know it is implicitly tested in buildConfig.test.ts, but I'd prefer to test this function in isolation. Otherwise this looks great, thanks @plemarquand

src/toolchain/BuildFlags.ts
plemarquand: done

award999 (approved):

PR #2196, opened 2026-04-16
Remove the reload extension dialog

matthewbastien:
## Description
Instead of forcing the user to reload the extension when toolchain changes are detected, the Swift extension will now automatically reload the `FolderContext` or `WorkspaceContext` in these cases. The extension can detect changes to:
- The global swiftly toolchain by polling `swiftly use -g`.
- Per-folder swiftly toolchains by watching for changes to `.swift-version` files.
- Xcode toolchains by polling `/var/select/developer_dir`.

Each of these will only force a reload if the active toolchain is managed by either `swiftly` or `xcrun`.

I've added an `onDidChangeWorkspaceContext` event to the `SwiftExtensionApi` that can be used to detect when the `WorkspaceContext` has been reloaded. For updates to the `FolderContext` toolchain I reused the `FolderOperation.swiftVersionUpdated` event that we added a while back, but didn't do anything with.

## Tasks


src/toolchain/SwiftlyToolchainWatcher.ts
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        switch (operation) {
+            case FolderOperation.add: {
+                const version = await this.readSwiftVersionFile(folder);
+                this.localSwiftVersions.set(folder.folder.fsPath, version);
+                const fsWatcher = vscode.workspace.createFileSystemWatcher(
+                    new vscode.RelativePattern(folder.folder, ".swift-version")
+                );
+                fsWatcher.onDidCreate(async () => await this.handleSwiftVersionFileChanged(folder));
```
award999, line 102: question: does the dispose() on line 109 handle the disposables returned by these events? If not, could store a compound disposable

src/ui/RestartEditor.ts
```diff
@@ -0,0 +1,38 @@
...
+    const selection = await vscode.window.showInformationMessage(
+        `Restart ${editorName}`,
+        {
+            modal: true,
+            detail: `You must restart ${editorName} in order for the ${type} installation to take effect.`,
+        },
+        `Quit ${editorName}`
+    );
+    if (selection === `Quit ${editorName}`) {
+        await vscode.commands.executeCommand(Workbench.ACTION_QUIT);
```
award999, line 36: question: would `workbench.action.restartExtensionHost` refresh the environment so that the user then doesn't have to restart vscode on their own?

src/toolchain/SelectedXcodeWatcher.ts
```diff
@@ -14,116 +14,91 @@
...
             return;
         }
-
-        if (developerDir && !matchesDeveloperDir(xcodePath)) {
+        // Check if this is a valid change in the Xcode path
+        this.xcodePath = newXcodePath;
+        if (!oldXcodePath || newXcodePath === oldXcodePath) {
+            return;
+        }
+        this.logger.info(`Selected Xcode changed from "${this.xcodePath}" to "${newXcodePath}"`);
```
plemarquand: this needs to be moved up because as at this point, this.xcodePath has already been assigned to newXcodePath

src/InternalSwiftExtensionApi.ts
```diff
@@ -301,6 +336,60 @@ export class InternalSwiftExtensionApi implements SwiftExtensionApi {
...
+    reloadWorkspaceContext(): void {
+        if (this.state.type === "uninitialized") {
+            throw new Error("The Swift extension has not been activated yet.");
+        }
+        if (this.state.type === "initializing") {
+            this.state.cancel();
+        }
+        let disposePromise = Promise.resolve();
+        if (this.state.type === "active") {
+            this.state.subscriptions.forEach(s => s.dispose());
```
plemarquand, line 365: should we immediately transition the state out of "active" once we start a dispose?

src/utilities/utilities.ts
```diff
@@ -474,3 +474,10 @@ export function destructuredPromise<T>(): {
     return { promise: p, resolve: resolve!, reject: reject! };
 }
 /* eslint-enable @typescript-eslint/no-explicit-any */
+
+export function captitalizeFirstLetter(str: string): string {
```
plemarquand:
```suggestion
export function capitalizeFirstLetter(str: string): string {
```

src/sourcekit-lsp/LanguageClientToolchainCoordinator.ts
```diff
@@ -76,16 +77,28 @@ export class LanguageClientToolchainCoordinator implements AsyncDisposable {
         }
         const singleServer = folder.swiftVersion.isGreaterThanOrEqual(new Version(5, 7, 0));
         switch (operation) {
+            case FolderOperation.swiftVersionUpdated: {
+                const existingClient = this.foldersToClient.get(folder.folder.toString());
```
plemarquand, line 81: do we need to update the `clients` map when we move the folder?

src/ui/RestartEditor.ts
matthewbastien: Unfortunately, it doesn't. VS Code only loads the terminal environment when the main process starts. Extension host reloads/restarts do not trigger a refresh.

src/utilities/utilities.ts
matthewbastien: Done.

src/InternalSwiftExtensionApi.ts
matthewbastien: Yeah this makes more sense to avoid race conditions. Done.

src/toolchain/SelectedXcodeWatcher.ts
matthewbastien: Good catch! I changed to using the existing `oldXcodePath` variable instead. Incidentally, that's what it was there for, but I guess I forgot to use it.

src/toolchain/SwiftlyToolchainWatcher.ts
matthewbastien: IIRC the only thing that the disposables returned by events do is remove the listener from the event emitter. We dispose of all active file system watchers in `dispose()` which should be enough.

src/sourcekit-lsp/LanguageClientToolchainCoordinator.ts
matthewbastien: Agreed, but this ended up being a whooooole can of worms. Turns out that the base `LanguageClient` implementation adds its own handling of `workspaceFolders` which ends up conflicting with how we're using the `didChangeWorkspaceFolders` request in `LanguageClientManager`. I'm working on another PR that will fix this.

plemarquand (approved):

PR #2197, opened 2026-04-22
Clean up CI workflows

matthewbastien:
## Description
Clean up various aspects of our CI workflows:
- Remove outdated swift versions
  - `Nightly` will now run tests against all active Swift versions
  - `Nightly - VS Code Insiders` will now run tests against only `nightly-6.3` and `nightly-main` Swift toolchains
- Rename the `tests` job to be more consistent between workflows
- Include the build output when DiagnosticsManager's tests fail
- Print the cause(s) of an error in the job summary if a test fails

The DiagnosticsManager test suite has been failing against nightly-main. This was tracked down to two issues:
1.  Debug builds are now bailing early when encountering errors. Run release builds for the test suite to force the build system to do a monolithic build that will report more errors.
2. There was a legitimate issue in the DiagnosticsManager where terminal output would be mangled because we weren't properly handling terminal control sequences, resulting in diagnostics being missed. We will now use a headless Xterm in order to get the true process output as seen by the end user in the terminal.

I've also disabled the tests involving code coverage on Windows with Swift nightly-main because they are currently failing due to swiftlang/swift-build#1353.

## Tasks

### DIFF
--- a/.github/workflows/nightly-insiders.yml
+++ b/.github/workflows/nightly-insiders.yml
@@ -51,7 +51,7 @@ jobs:
     with:
       needs_token: true
       # Linux
-      linux_exclude_swift_versions: '[{"swift_version": "5.8"}, {"swift_version": "5.9"}, {"swift_version": "5.10"}, {"swift_version": "6.0"}, {"swift_version": "6.1"}, {"swift_version": "nightly-6.1"}, {"swift_version": "nightly-6.2"}, {"swift_version": "nightly-6.3"}, {"swift_version": "nightly-main"}]'
+      linux_swift_versions: '["nightly-6.3", "nightly-main"]'
       linux_env_vars: |
         NODE_VERSION=${{ needs.package.outputs.node-version }}
         NODE_PATH=${{ needs.package.outputs.node-path }}
@@ -64,7 +64,7 @@ jobs:
       linux_pre_build_command: . .github/workflows/scripts/setup-linux.sh
       linux_build_command: ./scripts/test.sh
       # Windows
-      windows_exclude_swift_versions: '[{"swift_version": "5.9"}, {"swift_version": "6.0"}, {"swift_version": "6.1"}, {"swift_version": "nightly-6.1"}, {"swift_version": "nightly-6.2"}, {"swift_version": "nightly-6.3"}, {"swift_version": "nightly"}, {"swift_version": "nightly-main"}]'
+      windows_swift_versions: '["nightly-6.3", "nightly-main"]'
       windows_env_vars: |
         CI=1
         VSCODE_VERSION=insiders
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -63,7 +63,7 @@ jobs:
     with:
       needs_token: true
       # Linux
-      linux_exclude_swift_versions: '[{"swift_version": "nightly-6.1"}]'
+      # Run against all of the swift versions from https://github.com/swiftlang/github-workflows/blob/main/.github/workflows/swift_package_test.yml#L44
       linux_env_vars: |
         NODE_VERSION=${{ needs.package.outputs.node-version }}
         NODE_PATH=${{ needs.package.outputs.node-path }}
@@ -74,7 +74,7 @@ jobs:
       linux_pre_build_command: . .github/workflows/scripts/setup-linux.sh
       linux_build_command: ./scripts/test.sh
       # Windows
-      windows_exclude_swift_versions: '[{"swift_version": "nightly-6.1"}, {"swift_version": "nightly"}, {"swift_version": "6.2"}]'  # Missing https://github.com/swiftlang/swift/pull/80144
+      # Run against all of the swift versions from https://github.com/swiftlang/github-workflows/blob/main/.github/workflows/swift_package_test.yml#L85
       windows_env_vars: |
         CI=1
         VSCODE_SWIFT_VSIX_ID=${{needs.package.outputs.artifact-id}}
--- a/src/DiagnosticsManager.ts
+++ b/src/DiagnosticsManager.ts
@@ -18,11 +18,10 @@ import * as vscode from "vscode";
 import { WorkspaceContext } from "./WorkspaceContext";
 import configuration from "./configuration";
 import { SwiftExecution } from "./tasks/SwiftExecution";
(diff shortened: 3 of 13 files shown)


matthewbastien:
I also ran the two nightly CI workflows on my personal fork to make sure they run the correct Swift versions:
 - Nightly: https://github.com/matthewbastien/vscode-swift/actions/runs/24796367631
 - Nightly - VS Code Insiders: https://github.com/matthewbastien/vscode-swift/actions/runs/24797019977

rbenegal (approved):

.github/workflows/nightly-insiders.yml
```diff
@@ -44,14 +44,14 @@ jobs:
...
-    name: Test VS Code Insiders
+  tests:
+    name: Tests
     needs: package
     uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.11
     with:
       needs_token: true
       # Linux
-      linux_exclude_swift_versions: '[{"swift_version": "5.8"}, {"swift_version": "5.9"}, {"swift_version": "5.10"}, {"swift_version": "6.0"}, {"swift_version": "6.1"}, {"swift_version": "nightly-6.1"}, {"swift_version": "nightly-6.2"}, {"swift_version": "nightly-6.3"}, {"swift_version": "nightly-main"}]'
+      linux_swift_versions: '["nightly-6.3", "nightly-main"]'
```
award999, line 54: ohhh nice, glad they added this

.github/workflows/nightly-insiders.yml
```diff
@@ -44,14 +44,14 @@ jobs:
             *.vsix
             vscode-swift-sha.txt
 
-  tests_insiders:
-    name: Test VS Code Insiders
+  tests:
```
award999: we will want to keep the `test_` names or something similar, otherwise `act` doesn't know how to distinguish when running at desk

.github/workflows/nightly.yml
```diff
@@ -56,14 +56,13 @@ jobs:
             *.vsix
             vscode-swift-sha.txt
 
-  tests_release:
-    name: Test Release
+  tests:
```
award999: same

.github/workflows/nightly.yml
```diff
@@ -56,14 +56,13 @@ jobs:
...
+    name: Tests
     needs: package
     uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.11
     with:
       needs_token: true
       # Linux
-      linux_exclude_swift_versions: '[{"swift_version": "nightly-6.1"}]'
```
award999, line 66: I had to look up because my hunch was there was some default: https://github.com/swiftlang/github-workflows/blob/main/.github/workflows/swift_package_test.yml#L44

can you add a comment to point a newcomer where this is defined

.github/workflows/nightly.yml
```diff
@@ -74,7 +73,6 @@ jobs:
       linux_pre_build_command: . .github/workflows/scripts/setup-linux.sh
       linux_build_command: ./scripts/test.sh
       # Windows
-      windows_exclude_swift_versions: '[{"swift_version": "nightly-6.1"}, {"swift_version": "nightly"}, {"swift_version": "6.2"}]'  # Missing https://github.com/swiftlang/swift/pull/80144
```
award999, line 77: same

.github/workflows/nightly.yml
matthewbastien: Done.

.github/workflows/nightly.yml
matthewbastien: Done.

.github/workflows/nightly.yml
matthewbastien: Done.

.github/workflows/nightly-insiders.yml
matthewbastien: Done.

award999 (approved):

src/terminal/TerminalEmulator.ts
```diff
@@ -0,0 +1,91 @@
...
+            })
+        );
+    }
+
+    private sendLineData(buffer: IBuffer, lineIndex: number): void {
+        let line = buffer.getLine(lineIndex);
+        if (!line) {
+            return;
+        }
+        let lineData = line.translateToString(true);
```
award999, line 75: why true here, but false below? Cannot see API so just curious and could be worth a couple quick comments

src/terminal/TerminalEmulator.ts
```diff
@@ -0,0 +1,91 @@
...
+            }
+            lineData = line.translateToString(false) + lineData;
+        }
+        this.lineDataEmitter.fire(lineData);
+    }
+
+    public dispose(): void {
+        this.terminal.dispose();
+        this.subscriptions.dispose();
+        this.lineDataEmitter.dispose();
```
award999, line 90: what about closeEmitter?

src/DiagnosticsManager.ts
```diff
@@ -303,43 +302,37 @@ export class DiagnosticsManager implements Disposable {
                 disposables.forEach(d => d.dispose());
                 res(diagnostics);
             };
-            let remainingData: string | undefined;
             let lastDiagnostic: vscode.Diagnostic | undefined;
             let lastDiagnosticNeedsSaving = false;
+            const terminalEmulator = new TerminalEmulator(swiftExecution);
```
award999, line 307: this could also cause issues with build status. Can do this somewhere within the SwiftExecution or pseudo terminal?

matthewbastien: I'll look into that in another PR. I didn't want to touch something as low level as SwiftExecution or the pseudo terminal just yet. This would completely change the events coming out of those. Though it might be a good idea to do that instead of capturing the entire output which could be large for some build processes.

src/terminal/TerminalEmulator.ts
matthewbastien: That argument controls whether or not the returned string has its end trimmed. We want to trim the line terminator on the last line since it isn't needed. The following while loop walks up the lines in the buffer if the line is wrapped and we don't want to trim in that case because it could remove necessary whitespace.

src/terminal/TerminalEmulator.ts
matthewbastien: Good catch! Done.

Issue #1986, opened 2025-12-05
Test Failure: Test Explorer Suite | Debugging | lldb-dap | Debugs specified XCTest test failed on Windows nightly-main


matthewbastien:
**Failure logs**

```
AssertionError [ERR_ASSERTION]: 
  Build Output:
  > d:\a\vscode-swift\vscode-swift\assets\test\defaultPackage\.build\x86_64-unknown-windows-msvc\debug\defaultPackagePackageTests.xctest PackageTests.PassingXCTestSuite/testPassing

  + expected - actual

      {
        "enqueued": []
        "errored": []
        "failed": []
    -  "passed": []
    -  "skipped": [
    +  "passed": [
          "PackageTests.PassingXCTestSuite"
          "PackageTests.PassingXCTestSuite/testPassing"
        ]
    +  "skipped": []
        "unknown": 0
      }
  
  at assertTestResults (d:\a\vscode-swift\vscode-swift\test\integration-tests\testexplorer\utilities.ts:121:12)
  at Context.runXCTest (d:\a\vscode-swift\vscode-swift\test\integration-tests\testexplorer\TestExplorerIntegration.test.ts:111:30)
```

Link to failed run: https://github.com/swiftlang/vscode-swift/actions/runs/19972891704

**Additional context**
I was able to reproduce this myself on an x86_64 Windows machine with the latest nightly Swift release (6.3-dev). When you debug XCTest suites they will always show up as skipped in the UI after the debug session completes:

<img width="409" height="135" alt="Image" src="https://github.com/user-attachments/assets/d83b1353-36b8-4e2c-b023-d23b8f2e550b" />

matthewbastien:
Looks like `lldb-dap` in Swift 6.3 and up on Windows is not reporting process stdout under the `stdout` category anymore. We are explicitly filtering out non-stdout/stderr output when debugging XCTest to avoid having to parse through unrelated LLDB messages.

In Swift 6.2.4 (and 6.3 for macOS/Linux) the DAP reports debugee output under the `stdout` category:
```json
{
  "body": {
    "category": "stdout",
    "output": "Test Suite 'Selected tests' started at 2026-03-24 14:04:37.437\n"
  },
  "event": "output",
  "type": "event"
}
```

In Swift 6.3 on Windows the category is now `console`:
```json
{
  "body": {
    "category": "console",
    "output": "Test Suite 'Selected tests' started at 2026-03-24 14:04:37.437\n"
  },
[...]

PR #2232, opened 2026-05-14
Re-enable "Debugs specified XCTest test" test on Windows

matthewbastien:
## Description
A change was made in Swift 6.3 that correctly puts `lldb-dap`'s own process output under the `console` category. However, on Windows it cannot distinguish between its own output and the debuggee's. So, both of them will end up under `console` which the Swift extension ends up ignoring. This was fixed in Swift 6.4 by having `lldb-dap` use ConPTY to launch the debuggee.

This PR re-enables the failing tests and adds a workaround for Swift 6.3 that will include output from the `console` category as process output.

Issue: #1986 

## Tasks

### DIFF
--- a/src/debugger/logTracker.ts
+++ b/src/debugger/logTracker.ts
@@ -13,18 +13,41 @@
 //===----------------------------------------------------------------------===//
 import * as vscode from "vscode";
 
+import { WorkspaceContext } from "../WorkspaceContext";
 import { SwiftLogger } from "../logging/SwiftLogger";
 import { Disposable } from "../utilities/Disposable";
+import { Version } from "../utilities/version";
 import { LaunchConfigType } from "./debugAdapter";
 
 /**
  * Factory class for building LoggingDebugAdapterTracker
  */
 class LoggingDebugAdapterTrackerFactory implements vscode.DebugAdapterTrackerFactory {
+    constructor(private workspaceContext: WorkspaceContext) {}
+
     createDebugAdapterTracker(
         session: vscode.DebugSession
     ): vscode.ProviderResult<vscode.DebugAdapterTracker> {
-        return new LoggingDebugAdapterTracker(session.id);
+        const trackerOptions: LoggingDebugAdapterTrackerOptions = { ignoreConsoleOutput: true };
+        if (session.type === LaunchConfigType.LLDB_DAP && process.platform === "win32") {
+            // Normally, we ignore the console output category since it's reserved for messages from the
+            // debugger. However, in Swift 6.3 lldb-dap cannot properly distinguish between lldb's and the
+            // debuggee's output and will instead output everything under "console". Therefore, we need to
+            // include the "console" category with this configuration.
+            //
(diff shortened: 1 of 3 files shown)


rbenegal (approved):

matthewbastien:
Issue was fixed in `nightly-main`, but needs to be fixed in `6.3` and `nightly-6.3` before this PR can be merged.

matthewbastien:
On further discussion with the folks maintaining `lldb-dap`, it looks like this is just a known limitation in Swift 6.3 which has since been fixed in Swift 6.4. I've added the appropriate workaround and updated the PR description to indicate this.

src/debugger/logTracker.ts
```diff
@@ -21,10 +22,28 @@ import { LaunchConfigType } from "./debugAdapter";
...
+        const trackerOptions: LoggingDebugAdapterTrackerOptions = {};
+        if (session.type === LaunchConfigType.LLDB_DAP && process.platform === "win32") {
+            // Normally, we ignore the console output category since it's reserved for messages from the
+            // debugger. However, in Swift 6.3 lldb-dap cannot properly distinguish between lldb's and the
+            // debuggee's output and will instead output everything under "console". Therefore, we need to
+            // include the "console" category with this configuration.
+            //
+            // This issue was fixed in Swift 6.4, but the change cannot be easily cherry picked back to 6.3.
+            const swiftVersion = this.workspaceContext.folders.find(
+                f => f.workspaceFolder.uri === session.workspaceFolder?.uri
```
plemarquand: I remember being bit before by comparing Uris with strict equality because while their `fspath` may be the same other properties weren't. 

debugAdapterFactory does `f.workspaceFolder.uri.fsPath === folder?.uri.fsPath`, we should do the same here

src/debugger/logTracker.ts
```diff
@@ -21,10 +22,28 @@ import { LaunchConfigType } from "./debugAdapter";
...
+            )?.swiftVersion;
+            trackerOptions.ignoreConsoleOutput = Boolean(
+                swiftVersion?.isLessThan({ major: 6, minor: 3, patch: 0 }) ||
```
plemarquand: if no `swiftVersion` is supplied is `false` OK?

plemarquand: nit: elsewhere we just use `new Version(6, 3, 0)`

src/debugger/logTracker.ts
matthewbastien: Good point. Fixed.

src/debugger/logTracker.ts
matthewbastien: It probably is fine since that means that the debug session most likely wasn't launched by the Swift extension. Doesn't hurt to fix that to have the default be true, though. I've updated the code.

matthewbastien: Fixed.

plemarquand (approved):

Issue #2231, opened 2026-05-14
"Find All References" always includes the symbol declaration


YooGyeongMo:
## Description

Hi vscode-swift team! I'm Demian from south korea

I'd like to share a follow-up investigation that started in [swiftlang/sourcekit-lsp#2638](https://github.com/swiftlang/sourcekit-lsp/issues/2638), and ask for your thoughts before going further.

## Background

A few days ago, I picked up [swiftlang/sourcekit-lsp#2638](https://github.com/swiftlang/sourcekit-lsp/issues/2638), originally filed by @tothambrus11 , as my first open source contribution. The issue describes **Find All References** always including the symbol declaration in VS Code.

My first attempt was a naive fix on the SourceKit-LSP side. After looking more carefully at the LSP spec and discussing it with @ahoppen , it became clear that SourceKit-LSP is behaving correctly according to the `includeDeclaration` value it receives.

To verify what the client was sending, @ahoppen  suggested enabling extended logging and capturing a `sourcekit-lsp diagnose` bundle while reproducing the issue in VS Code.

Following that suggestion, I inspected the raw `textDocument/references` request and confirmed that VS Code sends
[related comment](https://github.com/swiftlang/sourcekit-lsp/issues/2638#issuecomment-4432518139)

  ```json                                                                          
  {                                                                                                                         
    "context": {                          
      "includeDeclaration": true                                                                                            
    }                                                                                                                       
  }                                           
  ```
All `textDocument/references` requests captured during the reproduction used `includeDeclaration: true`.
                                                                                                                            
I then added test coverage for the `includeDeclaration: false` case on the SourceKit-LSP side in [swiftlang/sourcekit-lsp#2644](https://github.com/swiftlang/sourcekit-lsp/pull/2644).

## Observation

Based on the diagnose result above, I started looking through the VS Code Swift extension.

In `src/sourcekit-lsp/LanguageClientConfiguration.ts`, the `middleware` object defines handlers for several language features, such as: 

- `provideDefinition`
- `provideCompletionItem`
- `provideCodeLenses`
- `provideDocumentSymbols`
- `provideInlayHints`
- `provideDiagnostics`

However, I could not find a `provideReferences` middleware.
                      
If I'm understanding this correctly, the extension may currently be relying on the default reference request behavior from `vscode-languageclient`, which appears to send `includeDeclaration: true`. That matches what was observed in the diagnose bundle.
  
## Possible approaches
                                          
One possible approach would be:

[...]

PR #2236, opened 2026-05-15
Add new swift.sourcekit-lsp.includeDeclarationInFindAllReferences setting

YooGyeongMo:
## Description

This PR adds a new `swift.sourcekit-lsp.includeDeclarationInFindAllReferences` setting, defaulting to `true`, that controls whether the symbol declaration is included in `Find All References` results.

When the setting is `false`, the extension sends `includeDeclaration: false` to SourceKit-LSP through a new `provideReferences` middleware, allowing users to opt out of including declarations in reference results.

The default value is `true` to preserve the existing behavior and stay consistent with other VS Code language extensions.

Issue: [https://github.com/swiftlang/vscode-swift/issues/2231](https://github.com/swiftlang/vscode-swift/issues/2231)

### Changes

- `package.json`: Add the new `swift.sourcekit-lsp.includeDeclarationInFindAllReferences` setting under `contributes.configuration`
- `src/configuration.ts`: Add the corresponding `LSPConfiguration` field and getter, validated via `validateBooleanSetting`
- `src/sourcekit-lsp/LanguageClientConfiguration.ts`: Add a `provideReferences` middleware that applies the setting to the outgoing references request
- `test/integration-tests/language/LanguageClientIntegration.test.ts`: Add an integration test case for the new behavior
- `CHANGELOG.md`: Add an entry under the unreleased `### Added` section

### Verification

I verified the behavior locally in the Extension Development Host against the `defaultPackage` test asset.

#### 1. New setting in VS Code Settings UI - `true`

The new `Swift › Sourcekit-lsp: Include Declaration In Find All References` setting appears in the Settings UI with the `markdownDescription` rendered correctly.

스크린샷 2026-05-16 오전 12 53 31

#### 2. Setting toggled to `false`

When the user disables the setting, the extension forwards `includeDeclaration: false` to SourceKit-LSP via the new `provideReferences` middleware.

스크린샷 2026-05-16 오전 12 57 46

#### 3. Default behavior — setting `true`

`Find All References` returns 3 results: the declaration in `PackageLib.swift` (`public let a = "B"`) and the usages in `main.swift` and `hello.swift`.

스크린샷 2026-05-16 오전 12 56 13

#### 4. Opt-out behavior — setting `false`

With the setting disabled, the declaration is excluded and only the 2 usages remain.

스크린샷 2026-05-16 오전 12 57 17

## Tasks

- [x] Added an entry to `CHANGELOG.md` if applicable

### DIFF
--- a/package.json
+++ b/package.json
@@ -971,6 +971,13 @@
             "order": 6,
             "scope": "machine-overridable"
           },
+          "swift.sourcekit-lsp.includeDeclarationInFindAllReferences": {
+            "type": "boolean",
+            "default": true,
+            "markdownDescription": "Whether the symbol declaration should be included in the results of `Find All References`. When `false`, only usage sites are returned.",
+            "order": 7,
+            "scope": "machine-overridable"
+          },
           "sourcekit-lsp.inlayHints.enabled": {
             "type": "boolean",
             "default": true,
--- a/package.json
+++ b/package.json
@@ -972,9 +972,19 @@
             "scope": "machine-overridable"
           },
           "swift.sourcekit-lsp.includeDeclarationInFindAllReferences": {
-            "type": "boolean",
-            "default": true,
-            "markdownDescription": "Whether the symbol declaration should be included in the results of `Find All References`. When `false`, only usage sites are returned.",
+            "type": "string",
+            "default": "default",
+            "enum": [
+              "default",
+              "always",
+              "never"
+            ],
+            "enumDescriptions": [
+              "Use the value sent by VS Code.",
+              "Always include the symbol declaration in `Find All References` results.",
+              "Never include the symbol declaration in `Find All References` results."
+            ],
+            "markdownDescription": "Controls whether the symbol declaration is included in the results of `Find All References`.",
             "order": 7,
             "scope": "machine-overridable"
           },
--- a/src/sourcekit-lsp/LanguageClientConfiguration.ts
+++ b/src/sourcekit-lsp/LanguageClientConfiguration.ts
@@ -287,6 +287,17 @@ export function lspClientOptions(
                 }
                 return result;
             },
+            provideReferences: async (document, position, options, token, next) => {
+                return next(
+                    document,
+                    position,
+                    {
+                        ...options,
+                        includeDeclaration: configuration.lsp.includeDeclarationInFindAllReferences,
+                    },
+                    token
+                );
+            },
             // temporarily remove text edit from Inlay hints while SourceKit-LSP
             // returns invalid replacement text
             provideInlayHints: async (document, position, token, next) => {
--- a/src/sourcekit-lsp/LanguageClientConfiguration.ts
+++ b/src/sourcekit-lsp/LanguageClientConfiguration.ts
@@ -288,15 +288,12 @@ export function lspClientOptions(
                 return result;
             },
             provideReferences: async (document, position, options, token, next) => {
-                return next(
-                    document,
-                    position,
-                    {
-                        ...options,
-                        includeDeclaration: configuration.lsp.includeDeclarationInFindAllReferences,
-                    },
-                    token
-                );
+                const setting = configuration.lsp.includeDeclarationInFindAllReferences;
+                const referenceOptions =
+                    setting === "default"
+                        ? options
+                        : { ...options, includeDeclaration: setting === "always" };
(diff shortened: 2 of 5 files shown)


YooGyeongMo:
For the tests, I added an integration test covering the main behavior: when the setting is disabled, `Find All References` excludes the declaration.

If you think more focused coverage would be useful on the `vscode-swift` side, I’d be happy to open a follow-up issue for it. For example, we could add a unit test for the `provideReferences` middleware, verify the outgoing `textDocument/references` request directly, or expand the integration test to cover more symbol kinds.

YooGyeongMo:
Hi @matthewbastien ,

Just following up on this PR.

When you get a chance, I’d really appreciate it if you could take a look.

I’m happy to make any changes if there’s a better direction for this.

Thank you :)

package.json
```diff
@@ -971,6 +971,13 @@
             "order": 6,
             "scope": "machine-overridable"
           },
+          "swift.sourcekit-lsp.includeDeclarationInFindAllReferences": {
+            "type": "boolean",
```
matthewbastien: Just in case VS Code ever changes the default, I would suggest we make this an enum with the following options:
- `default` which will not modify the request coming from VS Code
- `always` which will always set `includeDeclaration` to `true`
- `never` which will always set `includeDeclaration` to `false`

matthewbastien (changes requested):
Thank you for putting together this PR! I have one comment about the contents of the setting, but other than that it looks good.

src/sourcekit-lsp/LanguageClientConfiguration.ts
```diff
@@ -287,6 +287,14 @@ export function lspClientOptions(
                 }
                 return result;
             },
+            provideReferences: async (document, position, options, token, next) => {
+                const setting = configuration.lsp.includeDeclarationInFindAllReferences;
+                const referenceOptions =
+                    setting === "default"
+                        ? options
+                        : { ...options, includeDeclaration: setting === "always" };
+                return next(document, position, referenceOptions, token);
```
matthewbastien: nit: we can improve the readability of the code here.
```suggestion
                if (setting === "default") {
                    return next(document, position, options, token);
                }
                return next(
                    document,
                    position,
                    { ...options, includeDeclaration: setting === "always" },
                    token
                );
```

package.json
YooGyeongMo: Thanks for the suggestion!

Thanks to this, I was able to revisit the `"default"` behavior more deliberately and shape the code with extensibility in mind. I referenced the existing string literal union convention in this codebase (e.g., `traceServer`) rather than using TypeScript's `enum` keyword.

A couple of implementation notes:
- Used a string literal union (`"default" | "always" | "never"`) for the TypeScript type to follow the existing convention.
- The middleware branches via a ternary expression. Happy to refactor it to a `switch` statement if you find that clearer.

One quick question for my own learning

beyond forward-compatibility against VS Code default changes, were there other considerations like extensibility for future modes, or aligning with the existing `swift.diagnosticsStyle` pattern? PTAL!

matthewbastien (approved):

package.json
matthewbastien: No other reason, really. It was just to distinguish between the user explicitly setting `true` or `false` vs wanting the default behaviour. With just a boolean value we would always be setting `includeDeclaration` to `true` by default even if VS Code changed it later.

We use this pattern in a few other places for a similar reason. As you noticed, `swift.diagnosticsStyle` is one of them. The compiler could change the default diagnostic style on us later and we don't want to mess with that.

src/sourcekit-lsp/LanguageClientConfiguration.ts
YooGyeongMo: The early return reads much cleaner, thank you ! 

Reminds me of Swift's `guard` pattern in TypeScript form. Applied and pushed.

package.json
YooGyeongMo: @matthewbastien 

Thanks for the detailed explanation! Makes complete sense — distinguishing user intent from the default behaviour, and staying robust against upstream defaults changing. Good to know `swift.diagnosticsStyle` uses the same pattern for the same reason.

This has been a really enjoyable contribution, and I'd love to take on more issues like this when something fits. Thanks again for the patient review and clear feedback :)  genuinely helpful for learning.

PR #2238, opened 2026-05-16
Restore original LSP diagnostic message before forwarding to server

NSExceptional:
## Summary
- `DiagnosticsManager` capitalizes and strips `(fix available)` from incoming LSP diagnostic messages for display. That mutated message is stored in the `DiagnosticCollection` and is the one VS Code subsequently puts into `context.diagnostics` on `textDocument/codeAction` requests.
- clangd matches diagnostics in `context.diagnostics` against its fix-it cache by `(range, message)` ([`ClangdLSPServer.h` — `struct DiagKey`](https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clangd/ClangdLSPServer.h#L251-L260)). A mutated message means the lookup misses and no quickfix comes back.
- Stash the pre-mutation message on each diagnostic and add a `provideCodeActions` middleware that restores it before forwarding. Display-side behavior is unchanged.

This is the user-visible symptom in swiftlang/sourcekit-lsp#2186 (no "Add method implementation" quickfix for `-Wincomplete-implementation` warnings on Objective-C files). With this patch, the existing clangd fix-it round-trips through vscode-swift end-to-end.

## Repro before this PR
Project with an `@interface` declaring methods and an `@implementation` that doesn't define all of them:
- `Foo.h`: declares `-bravo:` and `+charlie`
- `Foo.m`: `@implementation Foo` (alpha only)

clangd publishes two `-Wincomplete-implementation` warnings. Cursor on `Foo` in `@implementation Foo`, hit Cmd+. → no code actions appear. With the fix, the menu shows:
- "Apply fix: insert `- (void)bravo:(NSString *)s {…`"
- "Apply fix: insert `+ (NSInteger)charlie {…`"

The clangd FIXME at `ClangdLSPServer.h:246-249` notes that long-term clangd should switch to using `Diagnostic.data`. That's being addressed separately; this PR is the immediate fix on the client side.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `eslint` clean
- [x] `npm run compile` clean
- [ ] CI runs the new `DiagnosticsManager.test.ts` (stashes original, no-op when nothing was mutated)
- [ ] Manual repro in VS Code with an iOS/Obj-C project that has missing method implementations

---

Yes, I authored this with my $200 Claude plan, I hope that's okay — it is the only way I can find the time to contribute features and bug fixes to the projects I love

### DIFF
--- a/src/DiagnosticsManager.ts
+++ b/src/DiagnosticsManager.ts
@@ -47,6 +47,27 @@ const isSourceKit: DiagnosticPredicate = diagnostic =>
     isSource(diagnostic, DiagnosticsManager.isSourcekit);
 
 /**
+ * Symbol-like key used to stash the original LSP message on a `vscode.Diagnostic`
+ * before `DiagnosticsManager` mutates it for display (capitalize, strip
+ * `(fix available)`, etc.). Read back in the `provideCodeActions` middleware
+ * so we forward the un-mutated message to the server.
+ *
+ * This matters because clangd looks up cached fix-its by `(range, message)`
+ * ([ClangdLSPServer.h] `struct DiagKey`); if the message differs from what it
+ * published, the lookup misses and the user sees no quickfix.
+ */
+const ORIGINAL_LSP_MESSAGE = "__vscodeSwiftOriginalLSPMessage";
+
+export function getOriginalLSPMessage(diagnostic: vscode.Diagnostic): string | undefined {
+    const raw = (diagnostic as unknown as Record<string, unknown>)[ORIGINAL_LSP_MESSAGE];
+    return typeof raw === "string" ? raw : undefined;
+}
+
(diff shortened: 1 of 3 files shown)


matthewbastien:
Thank you for putting this up! My first thought is that this is a fairly complicated solution that involves mutating VS Code's diagnostic objects. Especially considering that the extension probably shouldn't even be modifying the diagnostics coming back from the compiler/LSP in the first place.

Can you raise a corresponding issue with clear steps to reproduce (or ideally a sample project)? I'd like to take a closer look at this. The `DiagnostcsManager` code is fairly old at this point and probably needs a re-work in general.

NSExceptional:
@matthewbastien Done! Let me know if you need anything else!

PR #2254, opened 2026-05-27
Reveal the task terminal when clicking the Swift build status item

### LINKED ISSUE #2239: Clicking on resolving/building package status item shows notifications and not terminal
It used to be when you clicked on the resolving package or building status items it would show the terminal where these processes were running. In recent versions of the extension it just shows the notifications dialog saying the same thing as the status item, which is not particularly useful. Can we revert to showing the terminal where the process is running.

ayush-that:
## Description

When `swift.showBuildStatus` is set to the default `swiftStatus`, clicking the spinner in the status bar used to open VS Code's running tasks picker, and one more click took you to the task's terminal. After we switched to `vscode.window.withProgress({ location: ProgressLocation.Window })` in #2150, the spinner lost its click command, so clicking it now expands a popup with the same one-line message instead of jumping you to the terminal.

Restore the manual `StatusBarItem` path for `swiftStatus` mode so the click action points back at `workbench.action.tasks.showTasks`. The `progress` mode keeps `withProgress({ Window })` and `notification` keeps `withProgress({ Notification })`, preserving the documented split between the extension's own status item and VS Code's built-in progress indicator.

Resolves #2239.

## Tasks


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Fix swift.buildPath being ignored by swift package plugin tasks ([#2235](https://github.com/swiftlang/vscode-swift/pull/2235))
 - Fix nightly toolchains providing a value to `--swift-testing-event-stream-version` that swift-testing doesn't support ([#2247](https://github.com/swiftlang/vscode-swift/pull/2247))
 - Fix error during toolchain selection when swiftly has no active toolchain ([#2246](https://github.com/swiftlang/vscode-swift/pull/2246))
+- Reveal the task terminal when clicking the Swift build status item instead of showing a popup ([#2239](https://github.com/swiftlang/vscode-swift/issues/2239))
```
matthewbastien: The change log links to PRs, not issues:

```suggestion
- Reveal the task terminal when clicking the Swift build status item instead of showing a popup ([#2254](https://github.com/swiftlang/vscode-swift/pull/2254))
```

matthewbastien (changes requested):
Thank you for tackling this issue! I think we can improve this a little bit before merging.

The PR that switched to `withProgress()` was trying to avoid the constant flickering of the loading spinner when the build progress was being updated. Unfortunately, `withProgress()` does not have a good way to link back to the task that's being run. Reverting back to the old behaviour does solve the issue you linked, but regresses back to the flickering behaviour.

I'm fine with going back to the original status bar item, but I think we can resolve both issues at once If we limit the number of text changes that are made to the status bar item. This would involve stripping out the build progress indicator (e.g. `[225/3000]`) from the status bar item.

ayush-that:
okie. dropped the [x/y] counter so status bar text stays static (no flicker), click to reveal still works. and changelog updated.

src/ui/StatusItem.ts
```diff
@@ -13,34 +13,109 @@
...
+    private showTask(task: RunningTask, message?: string) {
+        const text = message ?? task.name;
+        const newText = `$(loading~spin) ${text}`;
+        // Skip redundant updates so the spinner doesn't flicker on repeated messages.
+        if (this.item.text !== newText) {
+            this.item.text = newText;
+            this.item.accessibilityInformation = { label: text };
+        }
+        // Only vscode.Task instances have a terminal to reveal via the running tasks picker.
+        this.item.command =
```
matthewbastien, line 113: I know this is how it worked before, but it is possible for multiple tasks to be running at the same time and this command can't distinguish between them. Is it possible to grab the terminal from `vscode.tasks.taskExecutions` and show it? We probably have to create our own command for this, but I think it would improve the UX.

test/unit-tests/ui/SwiftBuildStatus.test.ts
```diff
@@ -162,22 +173,21 @@ suite("SwiftBuildStatus Unit Test Suite", function () {
...
         // Ignore old stuff
-        expect(mockedProgress.report).to.not.have.been.calledWith({
-            message: "My Task: Fetching Dependencies",
-        });
-        expect(mockedProgress.report).to.not.have.been.calledWith({ message: "My Task: [6/7]" });
+        expect(statusItem.update).to.not.have.been.calledWith(
+            mockedTask,
+            "My Task: Fetching Dependencies"
+        );
+        expect(statusItem.update).to.not.have.been.calledWith(mockedTask, "My Task: [6/7]");
```
matthewbastien: Given that we only want the build progress to be updated a single time when it's in the status bar, we can replace all of these checks with a single `calledOnceWithExactly()` just to make the test a little easier to read:
```suggestion
        // swiftStatus shows just the task name, not the [completed/total] counter
        expect(statusItem.update).to.have.been.calledOnceWithExactly(mockedTask, "My Task");
```

test/unit-tests/ui/SwiftBuildStatus.test.ts
```diff
@@ -146,14 +156,15 @@ suite("SwiftBuildStatus Unit Test Suite", function () {
...
     });
 
     test("Update build progress", async () => {
-        configurationMock.setValue("progress");
-        buildStatus = new SwiftBuildStatus();
+        configurationMock.setValue("swiftStatus");
+        buildStatus = new SwiftBuildStatus(instance(statusItem));
         await didStartTaskMock.fire({ execution: mockedTaskExecution });
 
         testSwiftProcess.write(
```
matthewbastien: I would write multiple times to the process here with different progress amounts just to be sure that the status bar item is only updated once.

test/unit-tests/ui/SwiftBuildStatus.test.ts
```diff
@@ -146,14 +156,15 @@ suite("SwiftBuildStatus Unit Test Suite", function () {
...
-        expect(mockedProgress.report).to.have.been.calledWith({
-            message: "My Task: Fetching Dependencies",
-        });
+        expect(statusItem.update).to.have.been.calledWith(
+            mockedTask,
+            "My Task: Fetching Dependencies"
+        );
     });
 
     test("Update build progress", async () => {
```
matthewbastien:
```suggestion
    test("Only updates build progress once when shown in the status bar", async () => {
```

matthewbastien (changes requested):

ayush-that:
done. the status item click now reveals the specific task's terminal instead of the generic picker. since there's no api to map a task to its terminal, i match it by name (new swift.revealTaskTerminal command) and fall back to the picker if nothing resolves. tested with a few terminals open and it picks the right one. a first-class TaskExecution.terminal upstream would let us drop the heuristic.

src/utilities/tasks.ts
```diff
@@ -85,3 +85,11 @@ export function resolveScope(scope: vscode.WorkspaceFolder | vscode.TaskScope) {
     }
     return scope;
 }
+
+// VS Code has no API to map a task to its terminal, so match by name. The task surfaces with
+// a source prefix (e.g. "swift: Build All"), hence the suffix match alongside the exact one.
+export function findTaskTerminal(task: vscode.Task): vscode.Terminal | undefined {
+    return vscode.window.terminals.find(
+        terminal => terminal.name === task.name || terminal.name.endsWith(`: ${task.name}`)
```
matthewbastien: I think we should be explicitly matching on the `swift: ` prefix here since it's the only known prefix for our tasks.
```suggestion
        terminal => terminal.name === task.name || terminal.name === `swift: ${task.name}`
```

matthewbastien (changes requested):

src/utilities/tasks.ts
ayush-that: done.

matthewbastien (approved):
LGTM, thank you! This will be merged once CI confirms that the tests are passing.

PR #2260, opened 2026-05-28
Adding a clean build option

### LINKED ISSUE #1499: Consider adding a `Clean Build` option with SPM
**Is your feature request related to a problem? Please describe.**
When building with CMake, it is possible to do a clean build. This is not currently an option with the SPM based projects with VSCode.

**Describe the solution you'd like**
Having an additional build entry for `Clean Build` would allow you to do a full complete, clean build in a single invocation.

**Describe alternatives you've considered**
Running `swift package clean` before running the build task is a workaround.

**Additional context**

Pradyumn-cloud:
## Description
 I tried to fix it by introducing `swift.cleanRebuildAll` inside `commands.ts` resolving to a new function in build.ts. The logic links folderCleanBuild with `getBuildAllTask`. I modified `package.json` to inject "Clean Rebuild All" into the command palette.
To test my implementation i manually spun up the extension development, executed the test/assets module, and validated terminal logic correctly cascaded swift package clean before swift build hands free via keybindings. I also added test file for above change.

Issue: #1499

## Tasks


matthewbastien:
Thank you for taking a shot at this!

I looked over the issue with the team and we think that adding this as a new command is a bit niche and opens the door to add a bunch of other niche commands which could pollute the command palette/tasks view. Instead, we think it'd be better to:

- Have the extension provide a new task called `Clean Build Folder` that performs the `swift package clean` (pretty much the exact same thing as the `Swift: Clean Build Folder` command) and
- Update our documentation to show how tasks can be combined to accomplish a clean rebuild in the `tasks.json` by providing an example and linking to the VS Code tasks documentation.

This way custom tasks that involve cleaning before performing a build/run/test can be configured more easily by users. These tasks can also be added to debug configurations in the `launch.json` (we should probably link to the VS Code `preLaunchTask` docs for that as well).

If you're still up for continuing this PR then I'd be more than happy to assign the issue to you. Let me know if you run into any issues and I'd be more than happy to help out!

Pradyumn-cloud:
I agree with you and I will continue to work on this. I will close the current PR and start a new one implementing the proposed changes. Thanks for the guidance.

PR #2261, opened 2026-05-29
Show post-install script contents inline in confirmation dialog (#2139)

shivansh090:
## Description

  Embeds the contents of the Swiftly post-install script directly inside the confirmation modal so the user can review what
  will be executed before granting elevated permissions. Previously the modal said "the script will perform the following
  actions" but only summarised the action category — there was no way to see the actual commands without opening the file
  out-of-band.

  This is the approach @matthewbastien recommended in the review of the previous attempt (#2140), where a separate **Review
  Script** button + preview tab caused the modal to dim/block when the user switched tabs. Putting the script inline removes
  that friction entirely.

  The script content is rendered in the modal's `detail` area via the existing `MessageOptions.detail` field on
  `vscode.window.showWarningMessage` — no new UI primitives.

  **Edge cases handled:**
  - Scripts longer than 50 lines or 4000 characters are truncated with a `… (script truncated for display)` notice so the
  modal stays readable.
  - Empty or comments-only scripts show `(empty script)` instead of a blank detail block.
  - File-read failures are still surfaced as before, via the existing validation path.

  **Example — before vs. after (typical Swiftly script):**

  Before:
image

  After (same modal, with `detail` now populated with sample data):
image

  Issue: Fixes #2139. Supersedes #2140.

  ## Tasks

### DIFF
--- a/src/toolchain/swiftly.ts
+++ b/src/toolchain/swiftly.ts
@@ -993,7 +993,7 @@ export class Swiftly {
     }
 
     /**
-     * Shows confirmation dialog to user for executing post-install script
+     * Shows confirmation dialog to user for running the post-install commands.
      *
      * @param version The toolchain version being installed
      * @param validation The validation result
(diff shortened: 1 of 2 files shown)


matthewbastien (changes requested):
Thank you for putting up a PR for this! I like how this looks, but I think we can modify the wording a bit. Something like:

```
Swift ${version} installation requires additional system packages to be installed.

The Swift extension would like to run the following commands on your behalf:
${scriptContents}

Do you want to proceed with running these commands? You will be prompted for administrator privileges.
```

The user doesn't really need to know that it's being run in a script, just the commands that are being executed.

shivansh090:
Thanks for the review, @matthewbastien! That wording reads much more naturally
  from the user's perspective — "commands" is the right level of abstraction, the
  fact that they live inside a script is a detail the user doesn't care about.

  I'll update the message text along your wording, drop "Script contents:" from
  the detail block, and rename the action button from "Execute Script" to
  "Run Commands" so the whole dialog speaks the same language. Pushing the
  changes shortly.

shivansh090:
@matthewbastien Updated the dialog per your suggestion. PTAL whenever you have a moment.

image

test/unit-tests/toolchain/swiftly.test.ts
```diff
@@ -1111,6 +1111,17 @@ apt-get -y install libncurses5-dev`;
...
+            // so the user can review them without leaving the modal (issue #2139).
+            expect(mockVscodeWindow.showWarningMessage).to.have.been.calledWith(
+                match.string,
+                match({
+                    modal: true,
+                    detail: match("apt-get -y install build-essential").and(
+                        match("apt-get -y install libncurses5-dev")
+                    ),
+                })
+            );
```
matthewbastien: Can you separate this out into its own test?

src/toolchain/swiftly.ts
```diff
@@ -1002,25 +1005,62 @@ export class Swiftly {
...
-            `Swift ${version} installation requires additional system packages to be installed. ` +
-            `This will require administrator privileges.\n\n${firstTwoLines}\n\n` +
-            `Do you want to proceed with running the post-install script?`;
+            `Swift ${version} installation requires additional system packages to be installed.\n\n` +
+            `The Swift extension would like to run the following commands on your behalf:`;
+
+        const detail =
+            `${this.formatCommandsForReview(validation.scriptContent)}\n\n` +
+            `Do you want to proceed with running these commands? ` +
+            `You will be prompted for administrator privileges.`;
```
matthewbastien: Sorry in advance for making you change it again, but I thought about this a bit more (and tested on Ubuntu). I think simplifying the message to `Install System Packages`, removing some of the redundant wording from the detail, indenting the commands being run, and changing the button name to just `Install` makes this read a lot better.

Modal on Linux

src/toolchain/swiftly.ts
```diff
@@ -1002,25 +1005,62 @@ export class Swiftly {
...
+        if (body.length > maxChars) {
+            body = body.slice(0, maxChars);
+            truncated = true;
+        }
+
+        if (truncated) {
+            body += "\n… (truncated for display)";
+        }
+
+        return body;
```
matthewbastien: I find this hard to read and I don't think we need the character limit, really. It can be simplified to something like:
```suggestion
        const trimmedScriptContent = scriptContent.trim();
        if (trimmedScriptContent === "") {
            return "(no commands)";
        }
        
        let lines = trimmedScriptContent.split("\n");
        const maxLines = 10;
        if (lines.length > maxLines) {
            lines = lines.slice(0, maxLines - 1);
            lines.push("… (truncated for display)");
        }
        return lines.join("\n");
```

matthewbastien (changes requested):

test/unit-tests/toolchain/swiftly.test.ts
shivansh090: Done

src/toolchain/swiftly.ts
shivansh090: Addressed

src/toolchain/swiftly.ts
shivansh090: Resolved. PTAL

shivansh090:
All three points addressed in <1472e54>. PTAL.

matthewbastien (approved):
LGTM

I'll merge this once CI confirms that the tests are all passing. Thank you!

Issue #2263, opened 2026-06-02
Resolve repeatedly downloads packages and prints out warnings


adam-fowler:
### Description

Whenever I try to build or resolve all the dependencies are downloaded, regardless of whether I have already downloaded them already and I get warnings like the following
```
warning: 'hummingbird': skipping cache due to an error: Couldn’t fetch updates from remote repositories:
    fatal: cannot use bare repository '/Users/adamfowler/Library/Caches/org.swift.swiftpm/repositories/hummingbird-8c46860b' (safe.bareRepository is 'explicit')
```
This does not happen if run outside of VSCode. Even if I use the exact same command line using swiftly.

### Steps to reproduce

Run resolve, build, test on any project with dependencies

### Expected behavior

Not to download all my dependencies on every build.

### Extension version

2.16.5

### Swift & OS version

swift-driver version: 1.148.6 Apple Swift version 6.3.2 (swiftlang-6.3.2.1.108 clang-2100.1.1.101)
Target: arm64-apple-macosx26.0
Darwin Adams-MBP-M1-Max.local 25.3.0 Darwin Kernel Version 25.3.0: Wed Jan 28 20:53:15 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T6000 arm64

matthewbastien:
Can you create and upload a diagnostic bundle (`Swift: Capture Diagnostic Bundle`)?

adam-fowler:
[vscode-diagnostics-2026-06-02T16-56-21+01-00.zip](https://github.com/user-attachments/files/28517740/vscode-diagnostics-2026-06-02T16-56-21%2B01-00.zip)

matthewbastien:
I think this is caused by swiftlang/swift-package-manager#8068. I can only reproduce the issue when I set `safe.bareRepository` to `explicit` in my global git configuration:
```
git config set --global safe.bareRepository explicit
```
However, the same behaviour happens in both VS Code and when running `swift build` in my terminal. So, I'm not sure why it's only happening for you in VS Code.

I do know that newer versions of git have `safe.bareRepository` set to `explicit` by default now. Perhaps your VS Code environment is picking up a newer version of git somehow?

Can you try changing the setting to `all` to see if that resolves your issue?
```
git config set --global safe.bareRepository all
```

adam-fowler:
So I set safe.bareRepository to all and it is still happening, but only when I run a resolve via the VSCode task setup by the swift extension.

[...]

PR #2266, opened 2026-06-03
Override git safe.bareRepository for SwiftPM commands

plemarquand:
## Description
New VS Code versions do hardening and injects `safe.bareRepository=explicit` via GIT_CONFIG_* env vars. SwiftPM currently fails to pull dependencies from bare repositories (see https://github.com/swiftlang/swift-package-manager/issues/8068)

Add `safeBareRepositoryEnvironmentOverride` to reset matching entries to `all`, applied at every swift spawn.

Issue: #2263

## Tasks


adam-fowler:
This should probably only be temporary until MS get their act together. Or SwiftPM...

plemarquand:
@adam-fowler agreed; I'd like to eventually have SwiftPM support bare repos properly (but we'd still need this workaround to support older toolchains).

matthewbastien (approved):

rbenegal (approved):

PR #2276, opened 2026-06-10
Clean up our logging setup

matthewbastien:
## Description
Simplify how logging works in the extension by:
- Reducing the scope of `SwiftLogger` to be just formatting the logs appropriately. All things related to transports have been moved to their own classes that can be added or removed from the logger at runtime. This also allows the tests to use `SwiftLogger` directly rather than having to mock it.
- Creating a new `LoggedOutputChannel` class for output channels that we want logged to a file. VSCode's `OutputChannel` does not have a really good mapping for log levels and we ended up adding a bunch of unnecessary complexity to `SwiftLogger` by trying to support the output channel's `append()` function. This new class will just write the output channel's contents directly to a file as it's updated.

## Tasks


src/logging/FileWriteStream.ts
```diff
@@ -0,0 +1,112 @@
...
+    constructor(private readonly filePath: Promise<string>) {
+        void this.initializeFile();
+    }
+
+    private async initializeFile(): Promise<void> {
+        try {
+            const filePath = await this.filePath;
+            const fileHandle = createWriteStream(filePath, { flags: "a" });
+            fileHandle.on("ready", () => {
+                this.fileHandle = fileHandle;
```
plemarquand, line 37: If we've already been disposed at this point we shouldn't set fileHandle or it will be retained

src/logging/SwiftLoggerFactory.ts
```diff
@@ -11,44 +11,70 @@
...
-        return new SwiftLogger(name, join(folder.path, `${name}.log`));
+    createOutputChannel(name: string, logFileName: string): vscode.OutputChannel {
+        if (path.isAbsolute(logFileName)) {
+            throw Error(`Log file must be a relative path: "${logFileName}"`);
+        }
+        return new LoggedOutputChannel(name, this.logFilePath(logFileName));
+    }
+
+    private async ensureLogFolderExists(logFolderUri: vscode.Uri): Promise<string> {
+        if (!(await pathExists(logFolderUri.fsPath))) {
```
plemarquand: I like to use `await fs.mkdir(logFolderUri.fsPath, { recursive: true });` (essentially `mkdir -p`) instead since it avoids potential race between checking for existence and creation.

src/logging/FileWriteStream.ts
matthewbastien: Good catch! Done.

src/logging/SwiftLoggerFactory.ts
matthewbastien: Done.

plemarquand (approved):

PR #2283, opened 2026-06-15
Re-resolve unresolved breakpoints when the debugger stops

matthewbastien:
## Description
Works around an issue in older versions of `lldb-dap` where breakpoints would not be resolved correctly on startup. The extension will now force the resolution of unverified breakpoints when a breakpoint stop event is received from the debug adapter.

Issue: #1871 

## Tasks

### DIFF
--- a/test/unit-tests/debugger/breakpointVerificationTracker.test.ts
+++ b/test/unit-tests/debugger/breakpointVerificationTracker.test.ts
@@ -1,0 +1,204 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the VS Code Swift open source project
+//
+// Copyright (c) 2026 the VS Code Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of VS Code Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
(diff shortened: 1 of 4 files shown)


test/unit-tests/debugger/breakpointVerificationTracker.test.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    test("only refreshes once even with multiple stopped events", async () => {
+        const bp = new vscode.SourceBreakpoint(
+            new vscode.Location(vscode.Uri.file("/test.swift"), new vscode.Position(9, 0))
+        );
+        breakpointsValue.setValue([bp]);
+        mockSession.getDebugProtocolBreakpoint.resolves({ verified: false });
+
+        await tracker.onDidSendMessageAsync(stoppedEvent("breakpoint"));
+        await tracker.onDidSendMessageAsync(stoppedEvent("breakpoint"));
```
rbenegal, line 91: Is this supposed to be here twice?

rbenegal (approved):

test/unit-tests/debugger/breakpointVerificationTracker.test.ts
matthewbastien: Yeah, this test is trying to make sure that the breakpoint resolution only happens once. If it didn't work the first time then there's no point in trying again for this session.

award999 (approved):

PR #2288, opened 2026-06-15
Use swift folder path when resolving toolchain executables

matthewbastien:
## Description
The logic for finding executables in a direct toolchain installation was assuming that a `bin` directory existed within `toolchainPath`. However, this is not always the case (especially when setting `swift.path` to a custom toolchain build). Use the already known `swiftFolderPath` instead which is guaranteed to exist.

## Tasks


award999 (approved):

dschaefer2 (approved):
Thanks Matt. It's working great!

PR #2291, opened 2026-06-18
Expose backgroundPreparationMode setting

ktoso:
I've been hitting various mysterious issues in swift-java and finally tracked it down to being caused by the lazy indexing.

It seems to have problems when macros add conformances through extensions to existing types. Changing to the noLazy mode resolves the issues.

I have verified in swift-java that this can be fixed by configuring this setting in  `.sourcekit-lsp/config.json` file, but I think it's probably worth exposing it as a setting in the plugin.

## Tasks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Toolchain changes will now automatically reload the workspace ([#2196](https://github.com/swiftlang/vscode-swift/pull/2196))
+- Add `swift.sourcekit-lsp.backgroundPreparationMode` setting to control SourceKit-LSP's prepare-for-indexing mode. ([#2291](https://github.com/swiftlang/vscode-swift/pull/2291))
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


ahoppen:
Since this is just a workaround for an indexing issue and given that we already have a way to set it (through `.sourcekit-lsp/config.json`), I would prefer not to expose it as part of VS Code settings. Also, could you file an issue for the issue you’re seeing here to https://github.com/swiftlang/sourcekit-lsp?

matthewbastien (changes requested):
I agree with Alex that this should probably just stay as a configuration option rather than a VSCode setting. However, I wouldn't be opposed to adding a link to SourceKit-LSP's documentation in the `settings.md` and/or `language-features.md` userdocs to provide more visibility.

ktoso:
Sounds good, I’ll file an issue in sourcekit then

ktoso:
Here you go: https://github.com/swiftlang/sourcekit-lsp/issues/2696

PR #2293, opened 2026-06-23
Respect scratch path when listing Swift package plugins

plemarquand:
## Description
`swift package plugin --list` regenerates workspace-state.json as a side effect, and loadSwiftPlugins reads it back immediately afterwards. If the user has a `swift.buildArguments` setting that sets the scratch path these two paths could diverge, causing `loadSwiftPlugins` to read back a stale workspace-state.json instead of the one SwiftPM had just written.

Resolve the build directory once in loadSwiftPlugins and thread it into both loadPlugins (as an explicit --scratch-path on the --list call) and loadWorkspaceState, so the write and read always agree.

## Tasks
- [ ] ~Documentation has been updated~

### DIFF
--- a/src/SwiftPackage.ts
+++ b/src/SwiftPackage.ts
@@ -296,6 +296,7 @@ export class SwiftPackage implements ExternalSwiftPackage, Disposable {
         folder: vscode.Uri,
         toolchain: SwiftToolchain,
         logger: SwiftLogger,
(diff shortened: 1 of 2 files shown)


plemarquand:
Tracked by rdar://179873642

matthewbastien (approved):

PR #2304, opened 2026-07-13
Add setting to configure the build system backend used by SourceKit-LSP

owenv:
## Description
Add a setting to allow configuring the build system backend used by SourceKit-LSP to prepare a package for indexing and provide compiler arguments. This makes it easier to opt in to living on the experimental Swift Build BSP implementation as it's being developed.

## Tasks

### DIFF
--- a/package.json
+++ b/package.json
@@ -1017,6 +1017,23 @@
             "order": 7,
             "scope": "machine-overridable"
           },
+          "swift.sourcekit-lsp.swiftPM.buildSystemBackend": {
+            "type": "string",
+            "default": "auto",
+            "enum": [
+              "auto",
+              "native",
+              "swiftbuild"
+            ],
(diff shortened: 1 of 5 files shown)


award999:
@owenv thanks for the PR. Do you suspect that the build system is something that users will change all that often? The `buildSystem` is already a property in the [sourcekit-lsp config file](https://github.com/swiftlang/sourcekit-lsp/blob/main/Documentation/Configuration%20File.md#configuration-file) which I at least feel we should point users to that over introducing new settings. It's an interesting line like is SDK really more deserving of a unique setting over build system? Anyway, good conversation for us to have to establish that line.

owenv:
@award999 I expect this to be customized fairly rarely, feel free to close if you'd rather delegate it to the SourceKit-LSP config file

PR #2309, opened 2026-07-17
Handle uninstalled snapshot versions correctly

thePianoKid:
There is an edge case where [this regex](https://github.com/swiftlang/vscode-swift/pull/2309/changes#diff-7f73ec29addd84604470452045fa0e51d477f57654aa416aa7e36d54040a51acL152) makes the extension throw an error when it should just install the toolchain version in `.swift-version`. 

In this PR, I take out all the code that parses the error message to look for the toolchain version. Instead of that parsing logic, I just run `swiftly install`. It automatically installs the toolchain in `.swift-version`: no need to specify the version explicitly.


src/toolchain/swiftly.ts
```diff
@@ -633,38 +609,25 @@ export class Swiftly {
...
-                        // If the user dismisses the installation prompt then fall back
-                        // to using the global toolchain
-                        return await Swiftly.getActiveToolchain(extensionRoot, logger);
-                    }
-                }
+        } catch {
+            // The active toolchain may not be installed yet. Ask swiftly to install
+            // whichever version is currently selected (resolved from `.swift-version`
+            // or the global default) before locating it again.
+            logger.info("Active swift toolchain is not installed, installing via swiftly");
```
bkhouri, line 630: Suggestion: should we prompt the user before installing, which can be omitted in a setting?

thePianoKid: This breaks from what swiftly does. If you install Swiftly from the CLI, it chooses a toolchain and automatically installs it.

I think we should conform to swiftly's way of doing things. Happy to change my mind though.

thePianoKid:
In this commit, I changed the signature of `installToolchain` to make version optional. If the version is not given, it just runs `swiftly install`. 

This means that when the user opens a package that specifies a swift version they don't have installed, the extension will start the installation automatically. 

https://github.com/user-attachments/assets/67ceb7dd-aa97-457f-b5ca-72b02798e700

thePianoKid:
https://github.com/swiftlang/vscode-swift/pull/2309/commits/ddd9050a9508618c0b0bad41791706d040050844 added a setting to prevent the extension from automatically installing the missing toolchain specified in `.swift-version`. 

https://github.com/swiftlang/vscode-swift/pull/2309/commits/34f6c847b35e21b02666d6f572a739fad0ef5123 changed the "Missing toolchain" error so that the user knows they are getting this error because they disabled automatic toolchain installation:

image

src/toolchain/swiftly.ts
```diff
@@ -717,87 +692,56 @@ export class Swiftly {
...
-                    cancellationHandler?.dispose();
-                    resolve();
-                });
-
-                rl.on("error", err => {
-                    cancellationHandler?.dispose();
-                    reject(err);
-                });
-            });
+            progressPromise = this.readProgressPipe(
```
matthewbastien, line 732: Thank you for cleaning this up!

src/ui/ToolchainSelection.ts
```diff
@@ -54,27 +54,37 @@ async function selectToolchainFolder() {
...
  * @returns true if the user made a selection (and potentially updated toolchain settings), false if they dismissed the dialog
  */
-export async function showToolchainError(folder?: vscode.Uri): Promise<boolean> {
+export async function showToolchainError(
+    logger: SwiftLogger,
+    folder?: vscode.Uri
+): Promise<boolean> {
     let selected: "Remove From Settings" | "Select Toolchain" | "Open Documentation" | undefined;
-    const folderName = folder ? `${FolderContext.uriName(folder)}: ` : "";
+    const swiftVersionPath = folder ? `${FolderContext.uriName(folder)}/.swift-version: ` : "";
```
matthewbastien, line 67: The `.swift-version` file is not guaranteed to exist within the folder. For example, a project I'm working on has a top level `.swift-version` file with several subdirectories that contain `Package.swift` files. In this case the file is one directory above the folder root. Not sure the best way to surface that here without walking up the directory tree to find a file like swiftly does...

src/toolchain/swiftly.ts
matthewbastien: We should always prompt when the user opens up a folder that contains an uninstalled Swift version. There are times where a user might not want the toolchain to be installed and will instead modify the `.swift-version` file. I don't think this will have much of an impact on the installation error case that you're trying to fix here.

I do agree that we should be auto-installing the toolchain(s) when installing swiftly for the first time, though.

matthewbastien (changes requested):

PR #2332, opened 2026-08-11
Make symbol links clickable in documentation preview

Padmashree06:
## Description
- Makes symbol links in the Documentation Preview clickable.
- Extracts the symbol link from the `href` attribute of the clicked link.
- Resolves the symbol link's definition via `DocCSymbolLinkDefinitionRequest` and navigates to
  the definition location if it's found, otherwise shows an error message if the link can't be resolved.
  
### Dependency
This PR is dependent on https://github.com/swiftlang/sourcekit-lsp/pull/2748

## Tasks

### DIFF
--- a/src/ContextKeyManager.ts
+++ b/src/ContextKeyManager.ts
@@ -16,7 +16,11 @@ import * as vscode from "vscode";
 
 import { FolderContext } from "./FolderContext";
 import { SourceKitLanguageClient } from "./sourcekit-lsp/client/SourceKitLanguageClient";
-import { DocCDocumentationRequest, ReIndexProjectRequest } from "./sourcekit-lsp/extensions";
+import {
+    DocCDocumentationRequest,
+    DocCSymbolLinkDefinitionRequest,
+    ReIndexProjectRequest,
+} from "./sourcekit-lsp/extensions";
 import { Version } from "./utilities/version";
 
 /**
@@ -89,6 +93,11 @@ export interface ContextKeys {
     supportsDocumentationLivePreview: boolean;
 
     /**
+     * Whether the SourceKit-LSP server supports resolving DocC symbol links to their definition.
+     */
+    supportsSymbolLinkDefinition: boolean;
+
+    /**
      * Whether the installed version of Swiftly can be used to install toolchains from within VS Code.
      */
     supportsSwiftlyInstall: boolean;
@@ -139,6 +148,7 @@ export class ContextKeyManager implements ContextKeys {
     private _createNewProjectAvailable = false;
     private _supportsReindexing = false;
     private _supportsDocumentationLivePreview = false;
+    private _supportsSymbolLinkDefinition = false;
     private _supportsSwiftlyInstall = false;
     private _switchPlatformAvailable = false;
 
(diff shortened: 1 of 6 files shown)


src/documentation/webview/WebviewMessage.ts
```diff
@@ -33,6 +37,15 @@ export interface RenderedMessage {
     type: "rendered";
 }
 
+/**
+ * Sent from the webview to the extension when the user clicks a DocC symbol link,
+ * so the extension can resolve it via SourceKit-LSP and navigate to its definition.
+ */
+export interface SymbolLinkClickedMessage {
+    type: "symbolLinkClicked";
+    data: string;
```
matthewbastien: **nit:** I think this makes more sense to be called  `link` or`href` since that's what's actually getting passed here -`data` is a bit too generic.
```suggestion
    link: string;
```

src/documentation/DocumentationPreviewEditor.ts
```diff
@@ -163,6 +164,68 @@ export class DocumentationPreviewEditor implements Disposable {
...
+        }
+
+        const symbolLink = href.startsWith("/") ? href.slice(1) : href;
+        this.context.logger.debug(
+            `resolveSymbolLink: resolving "${symbolLink}" from ${document.uri.toString()}`
+        );
+
+        try {
+            const client = this.context.languageClientManager.getClient(folderContext);
+            this.context.logger.debug(`resolveSymbolLink: got client, sending request`);
```
matthewbastien: This debug message (along with some of the others) can be removed. I think it makes sense to keep logging the error paths to make diagnosing issues easier in the future. However, messages like this might be a bit too much.

src/documentation/DocumentationPreviewEditor.ts
```diff
@@ -163,6 +164,68 @@ export class DocumentationPreviewEditor implements Disposable {
...
+            const location = await client.useLanguageClient(async client => {
+                const result = await client.sendRequest(DocCSymbolLinkDefinitionRequest.type, {
```
matthewbastien, line 201: We should be checking whether or not SourceKit-LSP actually supports this request before we send it.

src/documentation/DocumentationPreviewEditor.ts
```diff
@@ -163,6 +164,68 @@ export class DocumentationPreviewEditor implements Disposable {
...
+            } else {
+                this.context.logger.debug(
+                    `resolveSymbolLink: server returned null/no location for "${symbolLink}"`
+                );
+                void vscode.window.showErrorMessage(
+                    `Symbol link could not be resolved: ${symbolLink}`
+                );
+            }
+        } catch (error) {
+            this.context.logger.error(`Failed to resolve symbol link "${symbolLink}": ${error}`);
```
matthewbastien: Use the built in error cause from NodeJS for this.
```suggestion
            this.context.logger.error(Error(`Failed to resolve symbol link "${symbolLink}"`, { cause: error }));
```

src/ContextKeyManager.ts
```diff
@@ -246,6 +256,18 @@ export class ContextKeyManager implements ContextKeys {
         );
     }
 
+    get supportsSymbolLinkDefinition(): boolean {
+        return this._supportsSymbolLinkDefinition;
+    }
+    set supportsSymbolLinkDefinition(value: boolean) {
```
matthewbastien: This doesn't need to be a context key right now since it's not actually being used anywhere. Normally, context keys are used in `"when"` clauses in the `package.json` to show/hide UI elements or commands.

matthewbastien (changes requested):

Padmashree06:
Thank you for the review! 
I have addressed the comments!

matthewbastien (approved):
Thank you!

PR #2333, opened 2026-08-11
Fix duplicate `LanguageStatusItem` caused by `reloadWorkspaceContext`

Padmashree06:
### Problem
If a configuration change (e.g. `swift.SDK`) fires while the extension is still activating, `reloadWorkspaceContext()` would cancel the in-flight `initializeWorkspace` call and immediately start a new one without waiting for the first to actually stop. Both ended up running in parallel and registering the same `LanguageStatusItem`, causing:

`LanguageStatusItem with id 'swiftlang version' already exists` error and a broken LSP.

### Fix 
Now when its `initializing`, instead of starting the new `InitializeWorkspace` immediately, we wait for the previous `initializeWorkspace` to fully finish (and clean up) before starting the next one.

### Testing 
Reproduced the issue with a custom LSP config that triggers a config-change reload during activation. After the fix, the duplicate `LanguageStatusItem` error no longer shows up and the LSP starts correctly.

## Tasks

### DIFF
--- a/src/InternalSwiftExtensionApi.ts
+++ b/src/InternalSwiftExtensionApi.ts
@@ -364,16 +364,21 @@ export class InternalSwiftExtensionApi implements SwiftExtensionApi {
         if (this.state.type === "uninitialized") {
             throw new Error("The Swift extension has not been activated yet.");
         }
-        if (this.state.type === "initializing") {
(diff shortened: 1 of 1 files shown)


matthewbastien (approved):
Thank you!

PR #2342, opened 2026-08-30
Move SourceKit-LSP configuration settings under swift.sourcekit-lsp

mihhhir08:
The docs for the Generate SourceKit-LSP Configuration command point people at a `swift.sourcekit-lsp.configurationBranch` setting, but that setting doesn't exist. It isn't in the configuration contributed by package.json and it doesn't show up anywhere in src. The setting that actually pins the schema branch is `swift.lspConfigurationBranch`, which is what generateSourcekitConfiguration.ts reads. I also fixed a small typo in the description for `swift.checkLspConfigurationSchema` where sourcekit was spelled sourckit. That string shows up in the Settings UI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -917,13 +917,17 @@
           },
           "swift.lspConfigurationBranch": {
             "type": "string",
-            "markdownDescription": "Set the branch to use when setting the `$schema` property of the SourceKit-LSP configuration. For example: \"release/6.1\" or \"main\". When this setting is unset, the extension will determine the branch based on the version of the toolchain that is in use."
+            "markdownDescription": "Set the branch to use when setting the `$schema` property of the SourceKit-LSP configuration. For example: \"release/6.1\" or \"main\". When this setting is unset, the extension will determine the branch based on the version of the toolchain that is in use.",
+            "deprecationMessage": "Deprecated: Please use \"swift.sourcekit-lsp.configurationBranch\" instead.",
+            "markdownDeprecationMessage": "**Deprecated**: Please use `#swift.sourcekit-lsp.configurationBranch#` instead."
           },
           "swift.checkLspConfigurationSchema": {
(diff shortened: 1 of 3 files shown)


matthewbastien (changes requested):
Thank you for the PR!

I think it makes more sense to match the rest of our SourceKit-LSP related settings by changing `swift.lspConfigurationBranch` to `swift.sourcekit-lsp.configurationBranch` and `swift.checkLspConfigurationSchema` to `swift.sourcekit-lsp.checkConfigurationSchema` (and any others like that if there are any).

We can leave the old settings to give an easier transition for those already using them, but can mark them as deprecated in the `package.json`. Have a look at the`swift.sourcekit-lsp.support-c-cpp` setting which was changed to `swift.sourcekit-lsp.supported-languages` for an example.

mihhhir08:
Thanks, that makes sense. I renamed both settings to `swift.sourcekit-lsp.configurationBranch` and `swift.sourcekit-lsp.checkConfigurationSchema` and moved them into the SourceKit-LSP section. Those two were the only ones left under the plain `swift` root, everything else was already namespaced. I kept the old settings and marked them deprecated pointing at the new names, same as `support-c-cpp`, and the getters still read the old key as a fallback when the new one isn't set so existing configs keep working. Funnily enough the docs already said `swift.sourcekit-lsp.configurationBranch`, so my original doc change is gone and the code now matches what the docs always claimed. I updated the tests to the new names and added one that checks the deprecated setting is still picked up. One thing I wasn't sure about, I left the accessor names in configuration.ts as `lspConfigurationBranch` and `checkLspConfigurationSchema` instead of moving them into the `sourcekitLSP` group, just to keep the diff small. Happy to move them if you'd rather.

matthewbastien (approved):
Thank you!

